PR #4, opened 2017-03-24
Publish firrtl locally first, to make Chisel happy

ben-k:
@zhemao Hopefully this is what you had in mind.


colinschmidt:
I'm not convinced this is the right way to do this. I believe people will complain about firrtl publishing over their potentially different and previous locally-published firrtl jar.

I think there is a different solution to this problem using SBT, so unless we explicitly don't want that solution I think we should use it here as we have used it in rocket-chip.

ben-k:
OK, I've demoted this to a README update until the more sophisticated version is implemented then.

PR #5, opened 2017-04-05
Typo in Readme

hossein1387:
The Readme has obvious typos in it.


zhemao:
This section has already been modified in the README (we only have two kinds of traits now), so this edit is no longer necessary.

Issue #6, opened 2017-04-06
Unresolved dependency when making verisim


ViccoYang:
When I run make in the "verisim" directory, something goes wrong.

vicco@ubuntu:~/project-template/verisim$ make
cd /home/vicco/project-template/rocket-chip && java -Xmx2G -Xss8M -XX:MaxPermSize=256M -jar /home/vicco/project-template/rocket-chip/sbt-launch.jar pack
Java HotSpot(TM) 64-Bit Server VM warning: ignoring option MaxPermSize=256M; support was removed in 8.0
[info] Loading project definition from /home/vicco/project-template/rocket-chip/project
Using addons: 
[info] Set current project to rocketchip (in build file:/home/vicco/project-template/rocket-chip/)
[info] Updating {file:/home/vicco/project-template/rocket-chip/}coreMacros...
[info] Resolving edu.berkeley.cs#firrtl_2.11;1.1-SNAPSHOT ...
[warn] 	module not found: edu.berkeley.cs#firrtl_2.11;1.1-SNAPSHOT
[warn] ==== local: tried
[warn]   /home/vicco/.ivy2/local/edu.berkeley.cs/firrtl_2.11/1.1-SNAPSHOT/ivys/ivy.xml
[warn] ==== public: tried
[warn]   https://repo1.maven.org/maven2/edu/berkeley/cs/firrtl_2.11/1.1-SNAPSHOT/firrtl_2.11-1.1-SNAPSHOT.pom
[info] Resolving jline#jline;2.12.1 ...
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn] 	::          UNRESOLVED DEPENDENCIES         ::
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn] 	:: edu.berkeley.cs#firrtl_2.11;1.1-SNAPSHOT: not found
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn] 
[warn] 	Note: Unresolved dependencies path:
[warn] 		edu.berkeley.cs:firrtl_2.11:1.1-SNAPSHOT
[warn] 		  +- edu.berkeley.cs:coremacros_2.11:3.1-SNAPSHOT
sbt.ResolveException: unresolved dependency: edu.berkeley.cs#firrtl_2.11;1.1-SNAPSHOT: not found
	at sbt.IvyActions$.sbt$IvyActions$$resolve(IvyActions.scala:313)
[...]

PR #12, opened 2017-07-14
Build firrtl first so we dont get that weird error

colinschmidt:
This addresses #4 and #6.

It ended up being much closer to #4 than I thought back then. Sorry @ben-k


ben-k:
It's OK!  You might want to adjust the README though if the publish-local step is no longer necessary.

colinschmidt:
Fixed the readme. @zhemao Any input?

colinschmidt:
enough time has passed merging this myself.

Issue #20, opened 2017-09-29
Problem when running PWMConfig


Ddnirvana:
Hi, guys,

I am new about chisel and rocket project so I am trying learn things about them by following this project-template. 
I can successfully run this project using DefaultExampleConfig and SimBlockDeviceConfig, but when I tried the PWMConfig, I failed in the `make CONFIG=PWMConfig` command.

The error is as the following: 

Could you please tell me how to fix this problem and what's cause?
Any advice will be helpful! Thanks a lot!

<pre>
info] Running example.Generator /home/dd/devlop/riscv/chisel/project-template/project-template/verisim/generated-src example TestHarness example PWMConfig
[info] [0.012] Elaborating design...
Interrupt map (2 harts 0 interrupts):

[error] (run-main-0) java.lang.reflect.InvocationTargetException
java.lang.reflect.InvocationTargetException
        at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
        at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
        at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
        at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
        at testchipip.GeneratorApp$$anonfun$circuit$1.apply(Generator.scala:22)
        at testchipip.GeneratorApp$$anonfun$circuit$1.apply(Generator.scala:19)
        at chisel3.core.Module$.do_apply(Module.scala:42)
        at chisel3.Driver$$anonfun$elaborate$1.apply(Driver.scala:92)
        at chisel3.Driver$$anonfun$elaborate$1.apply(Driver.scala:92)
        at chisel3.internal.Builder$$anonfun$build$1.apply(Builder.scala:240)
        at chisel3.internal.Builder$$anonfun$build$1.apply(Builder.scala:238)
        at scala.util.DynamicVariable.withValue(DynamicVariable.scala:58)
        at chisel3.internal.Builder$.build(Builder.scala:238)
        at chisel3.Driver$.elaborate(Driver.scala:92)
        at testchipip.GeneratorApp$class.circuit(Generator.scala:19)
        at example.Generator$.circuit$lzycompute(TestHarness.scala:20)
        at example.Generator$.circuit(TestHarness.scala:20)
        at testchipip.GeneratorApp$class.generateFirrtl(Generator.scala:30)
[...]

PR #22, opened 2017-12-13
Fix PWM beatByte problem: issure #20

hauhsu:


zhemao:
Issue has been fixed in a different way.

Issue #26, opened 2018-02-13
Failed to build default project


olix86:
Hi,

I did a clean install as described in th README and I'm trying to build the default example by running "make" in the verisim directory. 

The problem seems to be that VTestHarness.mk is used as a makefile but it's empty... (which is logic since it seems to be created by touch just before the "make" that crashes). Here's the output :

touch riscv/project-template/verisim/generated-src/example.TestHarness.DefaultExampleConfig/VTestHarness.mk
make VM_PARALLEL_BUILDS=1 -C riscv/project-template/verisim/generated-src/example.TestHarness.DefaultExampleConfig -f VTestHarness.mk
make[1]: Entering directory 'riscv/project-template/verisim/generated-src/example.TestHarness.DefaultExampleConfig'
make[1]: *** No targets.  Stop.
make[1]: Leaving directory '/riscv/project-template/verisim/generated-src/example.TestHarness.DefaultExampleConfig'
Makefile:55: recipe for target '/riscv/project-template/verisim/simulator-example-DefaultExampleConfig' failed
make: *** [/riscv/project-template/verisim/simulator-example-DefaultExampleConfig] Error 2

Do you have any idea what would be the problem?

Thanks!

colinschmidt:
Were there any other failures before this?

olix86:
No that was the first error. Thanks!

colinschmidt:
Can you post your full log in a https://gist.github.com/

Thanks

olix86:
There you go

https://gist.github.com/olix86/066df240711bdf9252872fd40152fbe6

zhemao:
The verilator command is supposed to create the VTestHarness.mk file, the "touch" command just updates the timestamp (some weird GNU make issue). Can you try deleting the generated-src/example.TestHarness.DefaultExampleConfig directory and running the verilator command manually?

Also, do you see the proper verilog in generated-src/example.TestHarness.DefaultExampleConfig.v file?

olix86:
I deleted generated-src/example.TestHarness.DefaultExampleConfig and ran the verilator command manually (the last one executed before the error message, which is what I assume you meant?) here it is just to be sure :

[...]

PR #28, opened 2018-03-09
Update Makefrag-verilator

olix86:
Changed verilator version from 3.904 to 3.920, which fixes a bug that prevented the default example to compile correctly


olix86:
Fixes issue #26

zhemao:
Tested and merged into master.

PR #29, opened 2018-04-13
bootrom: Fix race condition with interrupt_loop

a0u:
To detect the presence of other harts in the system, hart 0 sets each other hart's `MSIP` and then checks if the bit was successfully written.  However, if that hart somehow clears its own `MSIP` before the check occurs, hart 0 will prematurely exit `interrupt_loop` and fail to bring up the remaining harts.

This ensures that the other harts do not clear `MSIP` until hart 0 finishes executing `interrupt_loop`, which is signaled by hart 0 clearing its own MSIP.

`run-regression-tests` passes with a dual-core configuration - I inspected the logs to confirm that hart 1 behaves correctly.


zhemao:
The bootrom code has been moved to testchipip. I created a pull request there. Could you take a look?

https://github.com/ucb-bar/testchipip/pull/18

zhemao:
Also, there is an issue with the DualCoreConfig. It actually only produces one core. I'll fix that and rerun the regression tests.

a0u:
I hotfixed `DualCoreConfig` before running it.

PR #33, opened 2018-04-19
Update rocket-chip again

zhemao:


jwright6323:
My understanding of freechips/rocket-chip@8f3009feb1cc5f0242d17822f4578e1a5f8c3e57 is that ROCKET_CLASSES sets the BlackBox path and obviates the need to set rocketchip_vsrc_dir

zhemao:
No, ROCKET_CLASSES just sets the classpath so that FIRRTL can find the transformations to match the annotations. I've not yet figured out how the new BlackBox resources thing works.

jwright6323:
@grebe ?

grebe:
So, the current state that things are in is a little bit weird. I had a big PR that updated several different things all at once:
* build.sbt
* black box resources
* makefiles to set firrtl classpath correctly
* makefiles to use the newly generated black box resources

This was a lot to digest, so they got broken up and are still in the process of getting reviewed, merged, etc.

The current state of rocketchip is that:
1. vsrc/csrc got moved to `src/main/resources`, which includes them in rocketchip's jar.
2. `HasBlackBoxResource` was added to any black boxes, which means when firrtl runs it will spit out the appropriate black box into the build dir.

Rocket's build system still hasn't been updated to actually use these generated files, though. I'm using rocket as a library for things like queues and tilelink masters, etc., so rocket's build system doesn't really impact me (but it is *way* easier for me now that the blackboxes get emitted), but you guys have some decisions to make. If you want to be as much like rocket currently is right now, you should have references to vsrc/csrc and ignore the verilog/cpp that gets emitted by firrtl (other than your top level verilog file, of course).

If you want to do things the "right way", you can look at how I did it with makefiles in [my big PR](https://github.com/freechipsproject/rocket-chip/pull/1230). It was kind of fragile, so I wouldn't recommend it unless you like getting fancy with makefiles.

Does that help?

zhemao:
Yeah, let's stick with the way rocket-chip does it for now.

PR #43, opened 2019-01-28
fix README section on adding new submodules

zhemao:


README.md
```diff
@@ -396,7 +396,16 @@ to the project template.
 
     git submodule add https://git-repository.com/yourproject.git
 
-Then add `yourproject` to the `EXTRA_PACKAGES` variable in the Makefrag.
```
abejgonzalez: It is still important to keep this right just so that make still tracks the dependencies. I know that SBT handles the build once it is called but I think this is good to keep right?

abejgonzalez:
Other than my comment, I can also confirm that this works! LGTM

README.md
abejgonzalez: For example, the case that I am trying to avoid is this... you make a change in your new submodule "yourproject" and try to compile without added to the MakeFrag, then make will see that there is no changed sources and the build will continue without taking into account the updated changes. This is if the firrtl file is already generated.

abejgonzalez:
Just to confirm, is this good to merge @zhemao @alonamid ?

PR #47, opened 2019-02-19
Bump rocket, make possible to use published deps

grebe:
* Overriding `SBT` to `sbt -DROCKET_USE_MAVEN ++2.12.4` causes build to ignore testchipip and rocket-chip, instead using published dependencies. If you don't override `SBT`, it uses the submodules as before
* The build uses firrtl_black_box_resources.f to include blackboxes instead of hardcoded paths. This cleans up the build a bit, only including things you actually use
  * One caveat is that some blackboxes haven't been included correctly. [This PR](https://github.com/freechipsproject/rocket-chip/pull/1834) fixes the main one, and i want to clean up verisim/csrc a bit.
* bootrom is moved into project-template- maybe it shouldn't be in testchipip? I think it makes more sense here
* barstools is still a submodule. I would like to update it to use the new stages/phase stuff so it works nicely as a maven dep (also, we wouldn't need separate make targets to invoke barstools multiple times)
* Oh, and I also update the example a bit to work with current rocket


Makefrag
```diff
@@ -1,8 +1,10 @@
 ROCKETCHIP_DIR=$(base_dir)/rocket-chip
+TESTCHIP_DIR = $(base_dir)/testchipip
 
 SCALA_VERSION=2.12.4
 SCALA_VERSION_MAJOR=$(basename $(SCALA_VERSION))
 
+# SBT ?= sbt ++2.12.4
```
jwright6323: ?

src/main/scala/example/TestHarness.scala
```diff
@@ -15,13 +15,27 @@ class TestHarness(implicit val p: Parameters) extends Module {
...
+  dut.l2_frontend_bus_axi4.foreach(axi => {
+    axi.tieoff()
+    experimental.DataMirror.directionOf(axi.ar.ready) match {
+      case core.ActualDirection.Input =>
+        axi.r.bits := 0.U.asTypeOf(axi.r.bits)
+        axi.b.bits := 0.U.asTypeOf(axi.b.bits)
+      case core.ActualDirection.Output =>
+        axi.aw.bits := 0.U.asTypeOf(axi.aw.bits)
+        axi.ar.bits := 0.U.asTypeOf(axi.ar.bits)
+        axi.w.bits := 0.U.asTypeOf(axi.w.bits)
```
jwright6323: Should this be `axi.w.bits := DontCare` ?
...etc

verisim/Makefile
```diff
@@ -21,19 +21,22 @@ LDFLAGS := $(LDFLAGS) -L$(RISCV)/lib -Wl,-rpath,$(RISCV)/lib -L$(sim_dir) -lfesv
 include $(base_dir)/Makefrag
 include $(sim_dir)/Makefrag-verilator
 
+sim_blackboxes = \
+  $(build_dir)/firrtl_black_box_resource_files.f
 rocketchip_vsrc_dir = $(ROCKETCHIP_DIR)/src/main/resources/vsrc
```
jwright6323: Can we add a blank line between `sim_blackboxes` and `rocketchip_vsrc_dir` (or put `sim_blackboxes` on one line) to match style?

Makefrag
grebe: Left in by accident

grebe: Left in by accident

src/main/scala/example/TestHarness.scala
grebe: Good point, I think that's better

verisim/Makefile
grebe: The problem with using csrc in testchipip is that I'd like this to be able to work without having the submodule there. Maybe that's silly, but I'm looking at the notebooks as a use case.

There's a way to make this work with the blackbox stuff, but I'm not sure how to make it work because they don't really belong to a blackbox. I'm thinking of the best way to make it work

grebe: The problem with using csrc in testchipip is that I'd like this to be able to work without having the submodule there. Maybe that's silly, but I'm looking at the notebooks as a use case.

There's a way to make this work with the blackbox stuff, but I'm not sure how to make it work because they don't really belong to a blackbox. I'm thinking of the best way to make it work

src/main/scala/example/TestHarness.scala
jwright6323: Any particular reason that AW, AR, W are handled differently than R & B?

grebe: Because I'm dumb lol

grebe:
My latest commit removes `verisim/vsrc` and all but a couple headers in `verisim/csrc`.

I made a transform based on the blackboxresource transform to attach some resources to the TestHarness. This adds the csrc. Ideally, I would conditionally include `emulator.cc` only if I know I'm building with verilator.

Headers are tricky because they should be included with the `-include` flag, which the blackboxresource annotation doesn't have a way of doing. I think it will need to stay more hardcoded in the Makefile, unfortunately. Still, I should be able to have TestHarness pull the headers from resources instead of having them copied into the repo.

build.sbt
```diff
@@ -23,18 +23,26 @@ lazy val rocketchip = RootProject(file("rocket-chip"))
 lazy val testchipip = project.settings(commonSettings)
   .dependsOn(rocketchip)
 
-lazy val example = (project in file("."))
+def dependsOnRocket(prj: Project): Project = {
```
davidbiancolin: Probably just a nit, but it might be good to either change the name of this function, or add a comment that makes it explicit that this will ignore the submodule-based projects defined above when if using published versions.

verisim/Makefile
jwright6323: Does the requisite tooling live in project-template for it to work without testchipip? Or is this a packaging issue? Can you publish them in src/main/resources and then unzip them from jar in project-template?

build.sbt
grebe: Good point, hopefully what I've added is better. Also, it technically depends on testchipip so saying `dependsOnRocket` is a bit confusing.

jwright6323:
@grebe what about something like this: https://github.com/ucb-bar/project-template/commit/f99ac28b4564723ecaccfdc510d1564d5a33cd65

grebe:
Yeah, that's sort of what I'm thinking

zhemao:
We had bootrom in project-template initially, but eventually moved it to testchipip. I'd rather we keep it there.

grebe:
@zhemao do you think it's OK to put bootrom in testchipip's `src/main/resources`? Symlinks could preserve the current layout for build scripts. If we're going to keep it in testchipip, it needs to end up in the jar somehow for use-cases that pull from maven.

zhemao:
Yeah, we could put it in src/main/resources.

grebe:
Now `bootrom/` is generated (along with the other necessary c++/verilog files) as a build rule before running the rocket generator. Bootrom gets put in the top level directory, and everything else winds up in $(build_dir). Now, all you need to do is source two `.f` files, one of which contains black boxes and the other of which contains files (and command line flags) that verilator needs.

build.sbt
```diff
@@ -23,18 +23,29 @@ lazy val rocketchip = RootProject(file("rocket-chip"))
...
   .dependsOn(rocketchip)
 
-lazy val example = (project in file("."))
+// Checks for -DROCKET_USE_MAVEN.
+// If it's there, use a maven dependency.
+// Else, depend on subprojects in git submodules.
+def conditionalDependsOn(prj: Project): Project = {
+  if (sys.props.contains("ROCKET_USE_MAVEN")) {
+    prj.settings(Seq(
+      // libraryDependencies += "edu.berkeley.cs" %% "rocket-dsptools" % "1.2-020719-SNAPSHOT",
```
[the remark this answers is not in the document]
grebe: I don't think it would cause any problems to enable, but it probably shouldn't be in here. I'll just remove it.

PR #52, opened 2019-03-07
Fix AXI4 example.

grebe:
I accidentally stumbled into a working AXI4 configuration by multiplying pbus.beatBytes by 8, but it was fragile. This is the "right way" to add an AXI4 peripheral.


colinschmidt:
Actually do you want `:*=` ?

colinschmidt:
It seems like this widget doesn't need multiple ports

grebe:
Fair point, mindless copying is bad.

PR #50, opened 2019-02-26
Help people who want to run tests

colinschmidt:
Now you can use run-asm-tests etc...

PR #53, opened 2019-03-08
[WIP] Add BOOM / Cleanup Makefiles / Other misc additions

abejgonzalez:
- [x] Add `boom`
- [ ] Bump `boom` to use most recent `rocket-chip`
- [x] Separate out `Makefile/frag/variables`
- [x] Comment `Makefile/frag/variables`
- [x] Bump Verilator to 4.008
- [x] Add `scripts` directory
- [x] Add `riscv-torture`
- [x] Cleanup `.gitignore`
- [ ] Fix BOOM build - Currently doesn't find the BOOM generator
- [ ] Figure out proper way to run project regressions easily


abejgonzalez:
Couple questions and feedback:

- BOOM's generator creates a list of tests that we use to run regressions (aka all the RVC asm tests, benchmark tests, etc). This is done in the make system. I can try to figure out how to make this work with the current repo but do people want to have the generator in `src/main/scala` do something similar?

- I reorganized the Makefiles (added a new file for shared variables). Is this fine with people?

- Anybody have any issues with Verilator 4.008? It works for BOOM and Rocket-Chip seemed like it worked with it.

- Are we adding in `riscv-tools` into the repo?

- What do people prefer to run each subprojects regressions/tests? A `build-boom.sh` or adding to the Makefiles a new rule? I think this was mentioned at the meeting but I think I forgot what we decided.

colinschmidt:
Hwacha template also use Makefrag-variables which is really nice IMO.

alonamid:
re: riscv-tool, it's on my docket, I'm just want to verify that I add a hash that's consistent across the projects (because I don't believe we're synced yet)

jwright6323:
> Hwacha template also use Makefrag-variables which is really nice IMO.

Can we add `.mk` to the name (possibly rendering `Makefrag-` redundant)?

abejgonzalez:
Sure, I think that naming is fine. `common.mk` and `variables.mk` I think are good names. Any opposition?

jwright6323:
- [ ] We should chat at the rebar meeting about what dotfiles, config files, etc. should live in rebar (e.g. `.ctags`). Do we want to instead have a list of "useful things" for BAR project development? (firrtl vim syntax, etc.)

abejgonzalez:
Closing. Split this PR into 2 individual PR's #55 and #54

colinschmidt:
Can we either merge #50 first and rebase this on top or merge #50 into this branch.

PR #54, opened 2019-03-09
Makefile + Ctags

abejgonzalez:
Split that Colin asked for in #53. Splits makefiles, renames them, adds ctags script.


jwright6323:
Replied to the wrong PR earlier:
- [ ] We should chat at the rebar meeting about what dotfiles, config files, etc. should live in rebar (e.g. `.ctags`). Do we want to instead have a list of "useful things" for BAR project development? (firrtl vim syntax, etc.)

abejgonzalez:
Is the `.gitignore` settings done fine with people? I just auto default to ignoring everything in simulation directories.

alonamid:
@davidbiancolin to give a perspective on the firesim side of the makefiles, because firesim still needs to have both a standalone and a shared option

.gitignore
```diff
@@ -3,10 +3,8 @@ bootrom
 target
 *.jar
 *.stamp
-/vsim
-/verisim/generated-src*
-/verisim/simulator-*
-/verisim/verilator
-simv*
```
colinschmidt: I don't understand removing this based on your comment. Are you saying your personal gitignore ignores `vsim` and `verisim` or are you saying you don't look at all the spew this will cause in `git status`

colinschmidt:
@abejgonzalez I commented inline about the `.gitignore` stuff

.gitignore
abejgonzalez: Sorry this is unclear from my comment. Here I removed the files from the `.gitignore` because in the `sims/verisim/.gitignore` and `sims/vsim/.gitignore`, I use `*` to by default ignore all files in the `sims/<verisim/vsim>/`. Then in those files I selectively "unignore" all relevant files". I think this is better so that the `.gitignore`s in the simulation directories tell git what to ignore rather than the top-level `.gitignore`

colinschmidt: Ah makes sense on my last skim I didn't see the changes to the sim directory gitignores.

colinschmidt:
Can we either merge #50 first or merge #50 into this branch?

abejgonzalez:
I think merge #50 then rebase this. Avoid doing too much extra work :)

colinschmidt:
@abejgonzalez you are go for rebase

common.mk
```diff
@@ -1,50 +1,38 @@
...
 ROCKET_CLASSES ?= "$(ROCKETCHIP_DIR)/target/scala-$(SCALA_VERSION_MAJOR)/classes:$(ROCKETCHIP_DIR)/chisel3/target/scala-$(SCALA_VERSION_MAJOR)/*"
 TESTCHIPIP_CLASSES ?= "$(TESTCHIP_DIR)/target/scala-$(SCALA_VERSION_MAJOR)/classes"
+
+#########################################################################################
+# jar creation variables and rules
+#########################################################################################
 FIRRTL_JAR ?= $(ROCKETCHIP_DIR)/lib/firrtl.jar
 
+# this should match whatever the commonSettings version is in build.sbt
+BARSTOOLS_VER=1.0
```
davidbiancolin: not a big deal, but is there a specific reason to keep the version around if it's just whatever the build.sbt assembles? Could we just remove the suffix as a final stage of assembly?

davidbiancolin:
From the FIreSim, we'll probably just alias the canonical rebar variables to the ones we're using. We can talk about deprecating our tuple in the future.

colinschmidt:
@davidbiancolin Is there some variable we could define (and include in the simulator run arguments) such that you would be able to use whats in common to run your simulations?
Or could you point to the current command line you would use?

davidbiancolin:
> @davidbiancolin Is there some variable we could define (and include in the simulator run arguments) such that you would be able to use whats in common to run your simulations?
> Or could you point to the current command line you would use?

Yeah i think that shouldn't be too difficult.

common.mk
abejgonzalez: @jwright6323 Is there a reason that you did this?

common.mk
```diff
@@ -1,50 +1,38 @@
...
-
+#########################################################################################
+# set default shell for make
+#########################################################################################
+SHELL=/bin/bash
+
+#########################################################################################
+# variables to get all *.scala files
+#########################################################################################
 lookup_scala_srcs = $(shell find $(1)/ -iname "*.scala" 2> /dev/null)
```
abejgonzalez: Adding @davidbiancolin comment here that he added to the other PR: "Maybe consider making this chase symbolic links."

sims/verisim/verilator.mk
```diff
@@ -1,7 +1,17 @@
-# Build and install our own Verilator, to work around versionining issues.
-VERILATOR_VERSION=3.920
+#########################################################################################
+# verilator installation makefrag
+#########################################################################################
+
+#########################################################################################
+# verilator version, binary, and path
+#########################################################################################
+VERILATOR_VERSION=4.008
```
abejgonzalez: Is this verilator versiion good for people? It matches RC's version. (Note: Multithreading support is not working with RC and Boom currently but this change is keeping single-threaded support)

common.mk
abejgonzalez: Added a `-L` to follow symlinks. Seems to work on `firechip` in the `firesim/sim` directory.

common.mk
```diff
@@ -0,0 +1,149 @@
...
+#########################################################################################
+# include build/project specific makefrags made from the generator
+#########################################################################################
+ifneq ($(filter run% %.run %.out %.vpd %.vcd,$(MAKECMDGOALS)),)
+-include $(build_dir)/$(long_name).d
+endif
+
+#########################################################################################
+# default regression tests variables and rules
+# TODO: Remove in favor of each project having its own regression tests?
```
jerryz123: The regression tests are meant to be subset of the full asm-tests that a project supports, so I don't think a general-purpose "regression suite" makes sense here.

Also, does anyone even use the regression tests?

abejgonzalez: On `boom`'s end, we do not use the regression tests (we use the full asm and benchmark suite).

sims/verisim/verilator.mk
colinschmidt: Yes matching RC is good.

common.mk
colinschmidt: I use the regression tests when I want to make sure base rocket-chip works correctly. They are much faster than the whole suite. I think we should leave them in for CI at the very least.

abejgonzalez:
If no-one else has any issues with this, Ill go ahead and merge this into `rebar` by later today.

PR #58, opened 2019-03-17
Bump barstools and re-add infer-rw for better SRAM mapping

colinschmidt:
Depends on https://github.com/ucb-bar/barstools/pull/45


colinschmidt:
You can now get rw memories in your *.conf and can use an SRAM compiler with MacroCompiler.

jwright6323:
*obviously once the barstools submodule pointer is updated after ucb-bar/barstools#45

PR #59, opened 2019-03-18
Bump barstools and re-add infer-rw for better SRAM mapping

colinschmidt:
replacing #58


colinschmidt:
All submodules are merged so whenever someone wants to push this button go for it.

PR #63, opened 2019-03-28
Bumps barstools and fixes build system after

colinschmidt:
Barstools now handles annotations correctly.
This means that the blackboxresources for the harness
and top are different and need to be merged in the build system.
We also add all Sim*.cc files to default resources as our new emulator
demands. We then remove them from the harness .f file to avoid having
to detect which ones to include selectively.

### DIFF
--- a/Makefrag
+++ b/Makefrag
@@ -34,6 +34,8 @@ HARNESS_FILE ?=$(build_dir)/$(long_name).harness.v
 SMEMS_FILE   ?=$(build_dir)/$(long_name).mems.v
 SMEMS_CONF   ?=$(build_dir)/$(long_name).mems.conf
 sim_dotf ?= $(build_dir)/sim_files.f
+sim_harness_blackboxes ?= $(build_dir)/firrtl_black_box_resource_files.harness.f
+sim_top_blackboxes ?= $(build_dir)/firrtl_black_box_resource_files.top.f
 
 REPL_SEQ_MEM = --infer-rw --repl-seq-mem -c:$(MODEL):-o:$(SMEMS_CONF)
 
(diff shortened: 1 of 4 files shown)


Makefrag
```diff
@@ -63,9 +65,11 @@ $(FIRRTL_FILE) $(ANNO_FILE): $(SCALA_SOURCES) $(sim_dotf)
 
 $(VERILOG_FILE) $(SMEMS_CONF): $(FIRRTL_FILE) $(ANNO_FILE) $(TAPEOUT_JAR)
 	$(TAPEOUT) barstools.tapeout.transforms.GenerateTop -o $(VERILOG_FILE) -i $(FIRRTL_FILE) --syn-top $(TOP) --harness-top $(MODEL) -faf $(ANNO_FILE) $(REPL_SEQ_MEM) -td $(build_dir)
+	cp $(build_dir)/firrtl_black_box_resource_files.f $(sim_top_blackboxes)
 
-$(HARNESS_FILE): $(FIRRTL_FILE) $(ANNO_FILE) $(TAPEOUT_JAR)
+$(HARNESS_FILE): $(FIRRTL_FILE) $(ANNO_FILE) $(TAPEOUT_JAR) $(sim_top_blackboxes)
 	$(TAPEOUT) barstools.tapeout.transforms.GenerateHarness -o $(HARNESS_FILE) -i $(FIRRTL_FILE) --syn-top $(TOP) --harness-top $(MODEL) -faf $(ANNO_FILE) -td $(build_dir)
+	grep -v "SimSerial.cc\|SimDTM.cc\|SimJTAG.cc" $(build_dir)/firrtl_black_box_resource_files.f > $(sim_harness_blackboxes)
```
jwright6323: Can this list of `cc`s be a variable?

Makefrag
```diff
@@ -34,6 +34,8 @@ HARNESS_FILE ?=$(build_dir)/$(long_name).harness.v
 SMEMS_FILE   ?=$(build_dir)/$(long_name).mems.v
 SMEMS_CONF   ?=$(build_dir)/$(long_name).mems.conf
 sim_dotf ?= $(build_dir)/sim_files.f
+sim_harness_blackboxes ?= $(build_dir)/firrtl_black_box_resource_files.harness.f
+sim_top_blackboxes ?= $(build_dir)/firrtl_black_box_resource_files.top.f
```
jwright6323: Is this always put here, or do we pass this path in as an option? (I don't see the latter anywhere)

colinschmidt: I mirrored the other .f file above which is always put in the same place.
Unfortunately firrtl is dumb here https://github.com/freechipsproject/firrtl/blob/97e132e24226e69afc3a52d88c4a591b0678f816/src/main/scala/firrtl/transforms/BlackBoxSourceHelper.scala#L154
And always puts the .f file in the same name so we are doing the management ourselves. You could change this variable if you'd like them to live somewhere else.

Makefrag
colinschmidt: They are already hardcoded in here: https://github.com/ucb-bar/project-template/pull/63/files#diff-c9cadd0995d51c9ba231cc5f38627139L84
Obviously a better thing would be to have them both use the same environment variable. Not sure how much work that would be, or how often the list of off-chip interfaces will change.

jwright6323: ok, makes sense.

Makefrag
jwright6323: ok, makes sense.

PR #67, opened 2019-04-17
align rebar with tip of project-template master | fixes build issues

abejgonzalez:
Now allows for `make` to run. Also aligns with the commits that John/Colin did in the main repo.


abejgonzalez:
Also, I think we should setup blocked PR merges until someone approves of them.

abejgonzalez:
@jwright6323 My impression is that `rebar-dev` would not be merged into master until the retreat.

alonamid:
Given that rebar-dev is likely not going to be stable until the retreat, I don't think you would want tapeout stuff to depend on an unstable upstream

PR #69, opened 2019-04-20
[rebar] add sifive blocks | add rebar configs for boom

abejgonzalez:
Added `sifive-blocks` without actually using it. Added BOOM rebar configurations.

Question to the pro Chisel devs in the group. Is there any way for the `BuildTop` parameter to be general so that both `BoomTop` and `RocketTop` work? Similar to a upper bound type for a class parameter? I was unable to figure this out, but I figured this was sufficient for now since it matches `firechip`


alonamid:
Why does `boomexample` need to be a separate package than `example`?
Do you have an example of a config that has both a boom core and a rocket core?
Might also be worth changing the names of current `ExampleTop` to `RocketExampleTop` (to keep in symmetrical with `BoomExampleTop`).

build.sbt
```diff
@@ -42,6 +42,7 @@ def conditionalDependsOn(prj: Project): Project = {
 }
 
 lazy val example = conditionalDependsOn(project in file("."))
+  .dependsOn(boom)
```
colinschmidt: You should add sifive-blocks to the build.sbt and make the top-level example project depend on all generator packages IMO.

colinschmidt:
I also agree with Alon about creating extra packages.

abejgonzalez:
@alonamid @colinschmidt 

1. Separate packages: The separate packages were just following what `firechip` did. I will go ahead and combine it into a single package (although names will get a bit longer so there are no naming conflicts)

2. Boom + Rocket: No, I don't have a config for that yet. I was going to not put a heterogeneous multicore config in this PR.

3. ExampleTop naming: Sounds like a plan

4. SiFive Peripheral: I think that would be a separate PR. Which one do we want?

abejgonzalez:
Actually, are you all in favor of getting rid of the `example` in all the class names and just having a comment header saying "THESE ARE EXAMPLE CONFIGS". Otherwise, adding example everywhere is kinda redundant.

colinschmidt:
Can you make sure you can run an assembly test on both boom and rocket before merging this?
Otherwise things make sense.

abejgonzalez:
Okay, I think this PR is ready to be merged in.

I addressed @colinschmidt concerns about building rocket. See the `SUB_PROJECT`=`rocketchip` to see the proper configs to set.

Additionally, made sure that asm tests ran correctly on rebar boom, rebar rocket, rocketchip, and boom.

build.sbt
```diff
@@ -24,10 +24,14 @@ lazy val rebarFirrtl = (project in file("tools/firrtl"))
 
 lazy val rocketchip = RootProject(file("generators/rocket-chip"))
 
-lazy val testchipip = (project in file("generators/testchipip"))
+lazy val rebarrocketchip = project
```
davidbiancolin: What does this achieve?

src/main/scala/example/ConfigMixins.scala
```diff
@@ -0,0 +1,170 @@
...
+})
+
+// ----------------------------------------
+// Rocket Top Level System Parameter Mixins
+// ----------------------------------------
+
+/**
+ * Class to specify a "plain" top level rocket-chip system
+ */
+class WithNormalRocketTop extends Config((site, here, up) => {
```
davidbiancolin: Why can't the generator inject these into the parameter space? Do you guys just specify a config, not a top-level module in your generator arguments?

build.sbt
abejgonzalez: This is to allow people to build directly from the `rocket-chip` repo. 

For example, you want to use their config, harness, generator, etc. So you can set all the make params to address this (use `freechips.rocketchip.system` etc). However, you also need to point to the SBT project that the generator is in. In this case, I was unable to use the `rocketchip` SBT project (assuming this is something with `RootProject` not allowing you to access SBT projects inside of it). So I did the Hwacha approach of having a simple project that only depends on `rocketchip` so that SBT doesn't have to rebuild the other projects in `example`. The cleaner solution is to just to use `example` but if `testchipip` changes you have to have SBT rebuild it when you don't want it in this case.

davidbiancolin: Ah, curious.

src/main/scala/example/ConfigMixins.scala
abejgonzalez: Thats an option that would remove all of these mixins, but we have to do the connections somehow (aka the `dut.connectIO`). Im assuming that would still be done in the generator?

abejgonzalez: Also, I think that can be punted until another PR?

davidbiancolin: I warming to the idea of having methods like that in our targets (see the comment in my PR). Eventually, i think we'd want to do the same thing with firesim endpoints.

build.sbt
abejgonzalez: Any thoughts? Does `firechip` care about this extra SBT project?

src/main/scala/example/ConfigMixins.scala
abejgonzalez: So, I think this can be punted down the road. And an issue can be made of it for followup.

abejgonzalez: Offline: Merging in. Dealing with this later.

variables.mk
```diff
@@ -17,19 +18,39 @@
...
+	MODEL=TestHarness
+	CFG_PROJECT=boom.system
+	SBT_PROJECT=boom
 	TOP=ExampleBoomSystem
 endif
+ifeq ($(SUB_PROJECT),rocketchip)
+	# for Rocket-chip developers
+	PROJECT=freechips.rocketchip.system
+	MODEL=TestHarness
+	CONFIG=DefaultConfig
```
colinschmidt: I know I'm too late but is it not worth letting the user set `CONFIG` for all subprojects but with sane defaults using `?=`?

abejgonzalez: True. I think the best thing to do is set the other defaults and leave `CONFIG` unset

abejgonzalez: Did #71 address your concerns?

colinschmidt: Yes.

PR #71, opened 2019-04-23
[rebar] adding Hwacha

jerryz123:
~Building Hwacha configs located in the Hwacha package does not work due to file naming issues.~
~Building Hwacha configs located in the Hwacha package seems to cause multiple elaborations.~

### DIFF
--- a/variables.mk
+++ b/variables.mk
@@ -5,26 +5,28 @@
 #########################################################################################
 # default variables to invoke the generator for a example Rocket system
 # descriptions:
-#   PROJECT = the scala package to find the MODEL/Generator in
+#   SBT_PROJECT = the SBT project that you should find the classes/packages in
 #   MODEL = the top level module of the project in Chisel (normally the harness)
 #   VLOG_MODEL = the top level module of the project in Firrtl/Verilog (normally the harness)
+#   MODEL_PACKAGE = the scala package to find the MODEL in
 #   CONFIG = the configuration class to give the parameters for the project
-#   CFG_PROJECT = the scala package to find the CONFIG class
-#   SBT_PROJECT = the SBT project that you should find the Generator class in
+#   CONFIG_PACKAGE = the scala package to find the CONFIG class
+#   GENERATOR_PACKAGE = the scala package to find the Generator class in
 #   TB = wrapper over the TestHarness needed to simulate in VCS
 #   TOP = top level module of the project (normally the module instantiated by the harness)
 #
 # project specific:
 # 	SUB_PROJECT = use the specific subproject default variables
 #########################################################################################
-PROJECT     ?= example
-MODEL       ?= RocketTestHarness
-VLOG_MODEL  ?= TestHarness
-CONFIG      ?= DefaultRocketConfig
-CFG_PROJECT ?= $(PROJECT)
-SBT_PROJECT ?= $(PROJECT)
-TB          ?= TestDriver
-TOP         ?= RocketTop
+SBT_PROJECT       ?= example
+MODEL             ?= RocketTestHarness
+VLOG_MODEL        ?= TestHarness
+MODEL_PACKAGE     ?= $(SBT_PROJECT)
+CONFIG            ?= DefaultRocketConfig
+CONFIG_PACKAGE    ?= $(SBT_PROJECT)
+GENERATOR_PACKAGE ?= $(SBT_PROJECT)
+TB                ?= TestDriver
+TOP               ?= RocketTop
 
 #########################################################################################
 # subproject overrides
@@ -42,28 +44,30 @@ ifeq ($(SUB_PROJECT),boomexample)
 endif
 # for BOOM developers
 ifeq ($(SUB_PROJECT),boom)
-	PROJECT=boom.system
-	MODEL=TestHarness
-	CFG_PROJECT=boom.system
 	SBT_PROJECT=boom
+	MODEL_PACKAGE=boom.system
+	MODEL=TestHarness
+	CONFIG_PACKAGE=boom.system
+	GENERATOR_PACKAGE=boom.system
 	TOP=ExampleBoomSystem
 endif
 # for Rocket-chip developers
 ifeq ($(SUB_PROJECT),rocketchip)
-	PROJECT=freechips.rocketchip.system
-	MODEL=TestHarness
-	CFG_PROJECT=freechips.rocketchip.system
 	SBT_PROJECT=rebarrocketchip
+	MODEL_PACKAGE=freechips.rocketchip.system
+	MODEL=TestHarness
+	CONFIG_PACKAGE=freechips.rocketchip.system
+	GENERATOR_PACKAGE=freechips.rocketchip.system
 	TOP=ExampleRocketSystem
 endif
 # for Hwacha developers
 ifeq ($(SUB_PROJECT),hwacha)
-	PROJECT=freechips.rocketchip.system
-	MODEL=TestHarness
-	CFG_PROJECT=hwacha
 	SBT_PROJECT=hwacha
+	MODEL_PACKAGE=freechips.rocketchip.system
+	MODEL=TestHarness
+	CONFIG_PACKAGE=hwacha
+	GENERATOR_PACKAGE=hwacha
 	TOP=ExampleRocketSystem
-	TB=TestDriver
 endif
 
 #########################################################################################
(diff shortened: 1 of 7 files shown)


abejgonzalez:
@colinschmidt This should now address your concerns from #69 about specifying configs.

abejgonzalez:
@jerryz123 The final commit now addresses both issues and fixes them.

abejgonzalez:
@alonamid and probably hwacha devs (@colinschmidt...)

1. You are referring to `SUB_PROJECT=hwacha` only being for `rocket-chip` based systems. So this is basically done this way to conform with what Hwacha does now. Right now, the submodule `hwatcha` only holds configs for `rocket-chip` based systems. Thus it sense for this `SUB_PROJECT` to only take into account `rocket-chip` systems (aka matching what `hwatcha-template` does). Now, this can be changed later to support `boom` in the `SUB_PROJECT` once configs are setup to support `boom` in the `hwacha` submodule (and change relevant `build.sbt` stuff). Otherwise, the short term solution for combining the two projects is to use `rebar` since `rebar` houses the configs that combine the two.

2. I think 1. addresses that :)

3. Agreed.

variables.mk
```diff
@@ -65,8 +79,8 @@ REBAR_FIRRTL_DIR = $(base_dir)/tools/firrtl
 long_name = $(PROJECT).$(MODEL).$(CONFIG)
 
 # if building from rocketchip, override the long_name to match what they expect
-ifeq ($(SBT_PROJECT),rebarrocketchip)
-	long_name=$(PROJECT).$(CONFIG)
+ifeq ($(PROJECT),freechips.rocketchip.system)
```
colinschmidt: Hwacha also uses this version of `long_name`

variables.mk
```diff
@@ -26,31 +26,45 @@ SBT_PROJECT ?= $(PROJECT)
...
 	PROJECT=freechips.rocketchip.system
 	MODEL=TestHarness
-	CONFIG=DefaultConfig
 	CFG_PROJECT=freechips.rocketchip.system
 	SBT_PROJECT=rebarrocketchip
 	TOP=ExampleRocketSystem
 endif
+# for Hwacha developers
+ifeq ($(SUB_PROJECT),hwacha)
+	PROJECT=freechips.rocketchip.system
```
colinschmidt: Hwacha will want to use its own generator which adds its assembly tests and benchmarks.

colinschmidt:
Also please just drop the `t`s... :(

variables.mk
abejgonzalez: Ok so it sounds like there needs to be another `ifeq` then.

variables.mk
abejgonzalez: Ok. We will switch to using Hwachas generator

abejgonzalez:
@colinschmidt I think it will be up to the Hwacha devs to figure out how to deal with the toolchain and PR it themselves. I think this is just to get the RTL building together.

Also, if I remove the `t` then I can't say... Hwatch (a)ut.

alonamid:
esp-tools is currently submoduled under the toolchains directory in rebar. The main issue (which I alluded to earlier) is when to build it - whether it should be manually built (using an init script), or associated with the Hwacha makefile variables.

jerryz123:
I think a manual build makes sense, since in the past variants of riscv-tools have been built with custom build scripts.

colinschmidt:
Well then did you at least build the tools manually yourself and run the tests?

jerryz123:
@colinschmidt @a0u Can you take a look at the `variables.mk` subproject for hwacha? `make SUB_PROJECT=hwacha CONFIG=ISCA2016Config` fails hwacha asm tests with
```
Assertion failed: 'A' channel re-used a source ID (connected at Debug.scala:487:16)
    at Monitor.scala:460 assert(!inflight(bundle.a.bits.source), "'A' channel re-used a source ID" + extra)
```

colinschmidt:
All assembly tests or just the ones that use hwacha? How does the top-level system differ from hwacha-template?

jerryz123:
Seems like the RTL simulator behaves correctly when I cleaned, rebuilt, and ran tests manually.

In any case there are bugs with the `make run-asm-tests` and the `run-asm-tests-debug` commands.  I think we'll iron out those bugs when we add CI.

colinschmidt:
What do you mean there are bugs with `make run-asm-tests`? a clean build didn't fix those?

jerryz123:
@abejgonzalez Why did you remove all the default configs from the sub projects?

jerryz123:
Both the RocketChip and Hwacha subprojects fail asm tests with.
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  could not open +verbose (did you misspell it? If VCS, did you forget +permissive/+permissive-off?)
```

colinschmidt:
Did you include `+permissive`? I don't see it in the vsim/makefile stuff but I'm not sure where it is supposed to live in the refactor.

colinschmidt:
See https://github.com/ucb-bar/hwacha-template/blob/master/vsim/Makefile#L52

abejgonzalez:
Spoke with @jerryz123 he confirmed that it works.

abejgonzalez:
So I am the one who made the naming change. The main goal of them was to change them for clarity purposes. Mainly, when you say what the `PROJECT` is, what does that mean... the SBT project, the package, the "meta" project associated with it, etc. I think with these names things are very clear so that if additional things are added (aka more projects) then there can be an easier understanding of what each variable means.


With that being said, I don't think it is a style preference. Since the variables are now longer, it hurts more to type out the entire `make` command. But I am in favor of doing this for clarity. Thoughts?

jwright6323:
My unsolicited 2 cents:
I would agree with @alonamid in categorizing it as a style preference. However, the new names are less ambiguous, so the changes seem OK to me. Just be careful not to try to boil the ocean with style nits.

abejgonzalez:
@jwright6323 @alonamid I can revert the changes if people want. They are not necessary like I said. They are there only to make things easier to understand.

jwright6323:
I'm not advocating reverting. I think it's fine to keep them in.

abejgonzalez:
@colinschmidt @davidbiancolin How does this look now for you all?

variables.mk
```diff
@@ -3,53 +3,88 @@
...
+#   SBT_PROJECT = the SBT project that you should find the classes/packages in
 #   MODEL = the top level module of the project in Chisel (normally the harness)
 #   VLOG_MODEL = the top level module of the project in Firrtl/Verilog (normally the harness)
+#   MODEL_PACKAGE = the scala package to find the MODEL in
 #   CONFIG = the configuration class to give the parameters for the project
-#   CFG_PROJECT = the scala package to find the CONFIG class
-#   SBT_PROJECT = the SBT project that you should find the Generator class in
+#   CONFIG_PACKAGE = the scala package to find the CONFIG class
+#   GENERATOR_PACKAGE = the scala package to find the Generator class in
 #   TB = wrapper over the TestHarness needed to simulate in VCS
```
colinschmidt: I don't think this is VCS specific, so I would just drop "in VCS" or change it to "to simulate using a verilog simulator"

variables.mk
```diff
@@ -62,11 +97,14 @@ REBAR_FIRRTL_DIR = $(base_dir)/tools/firrtl
...
 
-# if building from rocketchip, override the long_name to match what they expect
-ifeq ($(SBT_PROJECT),rebarrocketchip)
-	long_name=$(PROJECT).$(CONFIG)
+# match the long_name to what the specific generator will output
+ifeq ($(GENERATOR_PACKAGE),freechips.rocketchip.system)
+	long_name=$(CONFIG_PACKAGE).$(CONFIG)
+endif
+ifeq ($(GENERATOR_PACKAGE),hwacha)
+	long_name=$(MODEL_PACKAGE).$(CONFIG)
```
colinschmidt: Wait does this work? I thought hwacha's generator mirrored what rocket-chip does...

jerryz123: I think this does the equivalent thing as rocket-chip, just the naming is different. @abejgonzalez has the reasoning for why this appears different.

abejgonzalez: This works. However, the `hwacha` generator uses the following: `names.topModuleProject + "." + names.configs` whereas `rocket-chip` does the following: `names.configProject + "." + names.configs`. So they are slightly different

abejgonzalez: This change reflects that

PR #73, opened 2019-04-25
add basic regression script

abejgonzalez:
Super basic tests that should finish and pass.


colinschmidt:
Nice now we just gotta make sure make reports errors correctly.

abejgonzalez:
@colinschmidt that should address your concerns

PR #76, opened 2019-05-02
bump rocket-chip to enable large memory spaces

colinschmidt:
Just a place holder for those interested. Points to a branch of rocket-chip now, so will need to clean that up before merging.
Also going to test this in my downstream templates to make sure "real" memories still work.

### DIFF
--- a/Makefrag
+++ b/Makefrag
@@ -35,6 +35,9 @@ TOP_ANNO     ?=$(build_dir)/$(long_name).top.anno.json
 HARNESS_FILE ?=$(build_dir)/$(long_name).harness.v
 HARNESS_FIR  ?=$(build_dir)/$(long_name).harness.fir
 HARNESS_ANNO ?=$(build_dir)/$(long_name).harness.anno.json
+HARNESS_SMEMS_FILE ?=$(build_dir)/$(long_name).harness.mems.v
+HARNESS_SMEMS_CONF ?=$(build_dir)/$(long_name).harness.mems.conf
+HARNESS_SMEMS_FIR ?=$(build_dir)/$(long_name).harness.mems.fir
 SMEMS_FILE   ?=$(build_dir)/$(long_name).mems.v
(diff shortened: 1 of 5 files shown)


jerryz123:
Is it necessary to use MacroCompiler for the harness memories instead of vlsi_mem_gen? #77 explains why I want to use vlsi_mem_gen for those.

Makefrag
```diff
@@ -25,7 +25,11 @@ build_dir=$(sim_dir)/generated-src
 
 CHISEL_ARGS ?=
 
+ifneq ($(PROJECT),example)
+long_name=$(PROJECT).$(CONFIG)
+else
 long_name=$(PROJECT).$(MODEL).$(CONFIG)
+endif
```
jwright6323: Hmm why do we not use `$(PROJECT).$(MODEL).$(CONFIG)` for non-example projects? Also `ifneq then else` hurts my brain.

src/main/scala/example/TestHarness.scala
```diff
@@ -36,6 +40,82 @@ class TestHarness(implicit val p: Parameters) extends Module {
 }
 
 object Generator extends GeneratorApp {
+  //Copied from rocketchip
```
jwright6323: Do we not inherit this from rocketchip by using the `GeneratorApp` trait?

colinschmidt: I tried doing this some fancy way but since this branch is not long for the world I'm not gonna worry about it.

Makefrag
colinschmidt: hwacha and rocket-chip have different long_names but this works for now.

jwright6323: ok

src/main/scala/example/TestHarness.scala
jwright6323: fair enough

colinschmidt:
Ignoring the CI system because its not implemented on this branch. I've run all the tests in both vsim and verisim. (for the example design)

PR #77, opened 2019-05-08
Improve VCS waves performance by using vlsi-mem-gen for SimAXIMem

jerryz123:
**Issue**:
Dumping the SimAXIMem to waves degrades VCS performance, incurs a ~30 second startup penalty to VCS simulations, and generates >800MB .vpds. Ideally, we would be able to specify a flag in VCS to not dump memories above a certain size, or annotate the memory in Chisel/Firrtl as DoNotCollectWaves. The first solution is impossible, and the second requires too much re-engineering.

However, the SimAXIMem is larger than the fixed VCS memory dump size limit when it is generated as a unified memory, rather than a banked memory.

For `name mem_ext depth 33554432 width 64 ports mwrite,read mask_gran 8`, we should emit

```
reg [63:0] ram [33554431:0];
```
vs
```
reg [7:0] ram [0:33554431];
reg [7:0] ram_1 [0:33554431];
reg [7:0] ram_2 [0:33554431];
reg [7:0] ram_3 [0:33554431];
reg [7:0] ram_4 [0:33554431];
reg [7:0] ram_5 [0:33554431];
reg [7:0] ram_6 [0:33554431];
reg [7:0] ram_7 [0:33554431];
```

When simulating the combined ram, VCS will decide that it is too large to dump, ignoring it. The simulation will then startup instantly and generate small waveforms (~30M).

**Solution**
There are two workarounds for this.

The first is to generate a larger SimAXIMem with the `WithExtMemSize` mixin. This is undesirable since we shouldn't have to mess with the config to get good simulation performance.

The second is to use `vlsi_mem_gen` in rocketchip, which correctly generates the combined memory.

This PR attempts the second approach. I understand that `vlsi_mem_gen` is deprecated in favor of macrocompiler, but it should be fine to use that script only for simulated TestHarness memories?

- [ ] PR and bump barstools, since it needs to support repl-seq-mem in the generateHarness phase.
- [ ] Verify verisim still works

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -52,14 +52,20 @@ $(VERILOG_FILE) $(SMEMS_CONF) $(TOP_ANNO) $(TOP_FIR) $(sim_top_blackboxes): $(FI
 	cd $(base_dir) && $(SBT) "project tapeout" "runMain barstools.tapeout.transforms.GenerateTop -o $(VERILOG_FILE) -i $(FIRRTL_FILE) --syn-top $(TOP) --harness-top $(MODEL) -faf $(ANNO_FILE) -tsaof $(TOP_ANNO) -tsf $(TOP_FIR) $(REPL_SEQ_MEM) -td $(build_dir)"
 	cp $(build_dir)/firrtl_black_box_resource_files.f $(sim_top_blackboxes)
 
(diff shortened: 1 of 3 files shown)


jerryz123:
Closed since Colin's BigInt fixes will coincidentally resolve this

PR #78, opened 2019-05-08
Minor Makefile fixes

jerryz123:


sims/vsim/Makefile
```diff
@@ -94,4 +94,4 @@ $(output_dir)/%.vpd: $(output_dir)/% $(sim_debug)
 #########################################################################################
 .PHONY: clean
 clean:
-	rm -rf $(build_dir) csrc $(sim_prefix)-* ucli.key vc_hdrs.h
+	rm -rf $(sim_dir)/generated-src/* csrc $(sim_prefix)-* ucli.key vc_hdrs.h
```
abejgonzalez: I would define this in `variables.mk` so you can reuse it across the verilator and vcs sim directories. Which reminds me, can this be added in the verilator makefile also.

sims/vsim/Makefile
```diff
@@ -61,7 +61,7 @@ VCS_NONCC_OPTS = \
 	-sverilog \
 	+incdir+$(build_dir) \
 	+define+CLOCK_PERIOD=1.0 \
-    $(sim_vsrcs) \
+        $(sim_vsrcs) \
```
abejgonzalez: This also seems to still be misindented?

PR #79, opened 2019-05-10
Decouple SUB_PROJECT builds from example

jerryz123:
This is to prepare for the rocket-chip bump Colin proposed in #76.

The issue was that all subprojects depended on the example project to generate the simfiles. However, the example project depends on boom and hwacha for the example configs. This couples the build dependencies of projects, and makes rocket-chip bumps very painful.

For example, `make SUB_PROJECT=boom` would fail if Hwacha were broken, even though `boom` does not pull in Hwacha at all. 

The solution is to break the dependency between a sub-project and the top-level example-project by moving GenerateSimFiles to a separate project.
We now have two top-level projects.
- `example` will be where example configs live. This can (and probably should) depend on all subprojects, since regressions will build configs listed here
- `utilities` will house things that do not have dependencies. Currently only GenerateSimFiles lives here


common.mk
```diff
@@ -9,7 +9,7 @@ SHELL=/bin/bash
 lookup_scala_srcs = $(shell find -L $(1)/ -iname "*.scala" 2> /dev/null)
 
 PACKAGES=rocket-chip testchipip boom hwacha sifive-blocks
-SCALA_SOURCES=$(foreach pkg,$(PACKAGES),$(call lookup_scala_srcs,$(base_dir)/generators/$(pkg)/src/main/scala)) $(call lookup_scala_srcs,$(base_dir)/src/main/scala)
+SCALA_SOURCES=$(foreach pkg,$(PACKAGES),$(call lookup_scala_srcs,$(base_dir)/generators/$(pkg)/src/main/scala)) $(call lookup_scala_srcs,$(base_dir)/example/src/main/scala) $(call lookup_scala_srcs,$(base_dir)/utilities/src/main/scala)
```
colinschmidt: Maybe we should move them to `generators` so that you can just put `example` and `utilities` in the `PACKAGES` list?

abejgonzalez: I think we want to have a distinction between what are the base projects and what is stitching them together. I personally, this is fine despite it being a bit long.

colinschmidt: Why do we want this distinction? Isn't it evident from the git submodules? Would you be ok with `PACKAGES=$(BASE_PACKAGES) $(SUBMODULE_PACKAGES)`?

jerryz123: The `generators` directory is really a "ip" directory, so it feels wrong to move `example` and `utilities` there.

colinschmidt: I think co-locating these two in some other sub-directory fits with the other organization decisions.

colinschmidt: You could have moved to some other directory but I personally like this :)

jerryz123:
Tested to work.
Merging so CI work can continue.

PR #87, opened 2019-05-17
[rebar] speed up simulation after first

abejgonzalez:
Small improvement to reduce the amount of SBT calls. Should really only be called if the utilities files change or if the directory switches from `vsim` to/from `verisim`


abejgonzalez:
Wrong branch

PR #88, opened 2019-05-17
reduce amount of sbt calls for sim

abejgonzalez:
Small improvement to reduce the amount of SBT calls. Should really only be called if the utilities files change or if the directory switches from `vsim` to/from `verisim`

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -8,7 +8,7 @@ SHELL=/bin/bash
 #########################################################################################
 lookup_scala_srcs = $(shell find -L $(1)/ -iname "*.scala" 2> /dev/null)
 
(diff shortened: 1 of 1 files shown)


abejgonzalez:
I don't know who else should review this... I think one thing we need to determine is who should review rebar PR's (aka the short list of people)

colinschmidt:
Ideally who reviews it can be limited because of CI making sure the other things we don't think about get checked.
Does this save time in CI as well?

abejgonzalez:
This does not save time in CI since this is mainly an improvement for after the first time things are rebuilt

PR #91, opened 2019-05-20
rename output | helper rules to run binaries

abejgonzalez:
Rename simulator collateral into the `output/(long_name)` directory. Add simulator rules to run a binary (just wraps the parameters and helps in generating a nice vcd)


jerryz123:
So the command to run with a binary is like `make run-binary-debug CONFIG=myconfig BINARY=mybinary.riscv`?

sims/vsim/Makefile
```diff
@@ -86,6 +86,12 @@ $(sim_debug) : $(sim_vsrcs) $(sim_dotf)
 	rm -rf csrc && $(VCS) $(VCS_OPTS) -o $@ \
 	+define+DEBUG -debug_pp
 
+#########################################################################################
+# helper rules to run simulator with debug
+#########################################################################################
+run-binary-debug: $(sim_debug)
+	$(sim_debug) $(PERMISSIVEON) $(SIM_FLAGS) +vcdplusfile=$(long_name).vpd $(PERMISSIVEOFF) $(BINARY)
```
jerryz123: I would prefer if `$(long_name).vpd` was `$(BINARY).vpd`.

Also stderr should be piped ` 2> $(BINARY).out`. The output from `+verbose` goes to stderr.

abejgonzalez:
Yes. The run command would be the same that you use to build with the `BINARY` and `run-binary` tacked on.

So...

`make SUB_PROJECT=example BINARY=path/to/riscv-bin.riscv run-binary[-debug]`

sims/vsim/Makefile
abejgonzalez: How about something like `BINARY.LONG_NAME.out/vpd`. This matters because if you have multiple sims in the same folder and are running different configs, then the same file would conflict.

jerryz123: Sure. I didn't think of that.

common.mk
```diff
@@ -67,6 +67,12 @@ $(SMEMS_FILE) $(SMEMS_FIR): $(SMEMS_CONF)
 .PHONY: verilog
 verilog: $(sim_vsrcs)
 
+#########################################################################################
+# helper rules to run simulator
+#########################################################################################
+run-binary: $(sim)
+	$(sim) $(PERMISSIVE_ON) $(SIM_FLAGS) $(PERMISSIVE_OFF) $(BINARY) 3>&1 1>&2 2>&3 | spike-dasm > $(sim_out_name).out
```
jerryz123: You should pipe to `$(sim_out_name).out`, and then pipe that to `spike-dasm`. I like having the raw output around, it makes it easier to track down instruction bits.

sims/vsim/Makefile
```diff
@@ -86,11 +86,17 @@ $(sim_debug) : $(sim_vsrcs) $(sim_dotf)
 	rm -rf csrc && $(VCS) $(VCS_OPTS) -o $@ \
 	+define+DEBUG -debug_pp
 
+#########################################################################################
+# helper rules to run simulator with debug
+#########################################################################################
+run-binary-debug: $(sim_debug)
+	$(sim_debug) $(PERMISSIVE_ON) $(SIM_FLAGS) +vcdplusfile=$(sim_out_name).vpd $(PERMISSIVE_OFF) $(BINARY) 3>&1 1>&2 2>&3 | spike-dasm > $(sim_out_name).out
```
jerryz123: Ditto

common.mk
abejgonzalez: For now, I'll leave this so I can move to other stuff. But you should PR it!

PR #92, opened 2019-05-21
Add support for heterogenous SoCs (BOOM + Hwacha)

abejgonzalez:
Adds support for heterogeneous configs with both BOOM and Rocket.
- New subsystems, example system, and test harnesses
- Mixin to renumber harts based on the cores in the system

Also fixes the BOOM configs that were currently pushed in the repo to work.
- Forgot to specify `Build` variable so test harness could generate a system

Depends on the following PR's being merged in first:
#91 
https://github.com/riscv-boom/riscv-boom/pull/270

Preferably wait for (nicer prints for boom):
https://github.com/riscv-boom/riscv-boom/pull/269

Todo:
- [x]  Bump BOOM when PR https://github.com/riscv-boom/riscv-boom/pull/270 is merged 
- [x]  Rebase on #91 once done
- [x]  Add CI for the configs (if memory allows)
- [x] Remove debug prints in the tile


abejgonzalez:
Note: Some parts of this have to change with the BOOM + Rocket Chip version bump. So... keep that in mind.

abejgonzalez:
Preferably merged after https://github.com/riscv-boom/riscv-boom/pull/271 is merged.

abejgonzalez:
This needs to be bumped with https://github.com/riscv-boom/riscv-boom/pull/271 so that the BOOM CI can be added in. Otherwise, this is fine

abejgonzalez:
I will work on documentation this weekend so Ill put that in a different PR (since I want the "meaty" PRs to be done for David to continue on his FireSim work without major outstanding PRs)

alonamid:
`BoomAndRocket` doesn't really roll off my fingertips... but I guess we're already down too many naming debates right now to contain another one.
Should we have a naming convention for these heterogeneous configs that might make them easier to manage?

jerryz123:
So now we have `boomexample`, `example`, and `boomandrocketexample`.
Is there any way to roll all these together?

abejgonzalez:
@alonamid Agreed. However, this is what I went with for now because it makes thing extremely explicit at the cost of long strings. One option... though in the future... is to name the configs based on the name of the repo, i.e. RebarConfig, MarbleConfig, etc...

@jerryz123 The main difference between the two is that the `boomexample` and `boomrocketexample` share the `BoomAndRocket`... infrastructure while the `example` subproject uses a Rocket only top and harness. One option to reduce the amount of duplication could be to just use the heterogeneous infrastructure. However, we then don't have an example of using just a Rocket based system (we will only have the tile and harness that support both)

jerryz123:
Is there a good reason to keep around a Rocket-only system if the BoomAndRocket system has strictly more functionality?

abejgonzalez:
IMO its wierd to have `example` which is just used for Rocket based systems only, to use `BoomAndRocket` based components (esp. when `example` never uses `boom`)

jerryz123:
I guess the point is we could change `example` to support all possible combinations of Boom/Rocket.

abejgonzalez:
So the more I think about it... the more I think it is fine if `example` uses the heter-base subsystem in which case we can combine it. The only thing is that you now have to have the end user specify which `CONFIG` to use and portray that correctly in the documentation. Thoughts anyone?

alonamid:
I think if we can have a "catch-all" heter-base subsystem in which you have to change only the CONFIG, and it doesn't have an RTL overhead, then that's better

generators/example/src/main/scala/Configs.scala
```diff
@@ -14,7 +16,7 @@ class BaseRocketConfig extends Config(
   new freechips.rocketchip.system.DefaultConfig)
 
 class DefaultRocketConfig extends Config(
```
colinschmidt: This  name seems highly confusing. Also why are we getting rid of Rocket only configs?

variables.mk
```diff
@@ -174,4 +162,3 @@ sim_vsrcs = \
 # assembly/benchmark variables
 #########################################################################################
 timeout_cycles = 10000000
-bmark_timeout_cycles = 100000000
```
colinschmidt: This is helpful for people who have benchmarks that take longer than normal tests, but still want normal tests to timeout quickly.

generators/example/src/main/scala/ConfigMixins.scala
```diff
@@ -37,129 +41,63 @@ class WithGPIO extends Config((site, here, up) => {
...
-    }
-    top
-  }
-})
-
-// --------------------------------------
-// BOOM Top Level System Parameter Mixins
-// --------------------------------------
+// -----------------------------------------------
+// BOOM + Rocket Top Level System Parameter Mixins
```
colinschmidt: The use of `+` and `and` is confusing if my earlier comment about rocket only configs is not wrong. Should this be Boom And/Or Rocket? Do these configs work fro BoomOnly, RocketOnly, and BoomPlusRocket, or some subset of those?

variables.mk
abejgonzalez: I didn't see this get used... maybe I missed something but Ill add it back in.

generators/example/src/main/scala/ConfigMixins.scala
abejgonzalez: Ok. I can make this clearer. This for BOOM and/or Rocket.

generators/example/src/main/scala/Configs.scala
abejgonzalez: Are you refering to having both the `Base` and the `Default` configs? Also, we are not getting rid of Rocket configs... what are you refering to?

generators/example/src/main/scala/ConfigMixins.scala
abejgonzalez: The functionality provided by the `BoomAndRocket` named things allows you to do a Rocket only, BOOM only, and Rocket+BOOM system.

variables.mk
colinschmidt: ...another feature lost. Will make a PR this summer to fix this.

abejgonzalez:
Yea. About to do that.

variables.mk
abejgonzalez: I will not remove this then. Ill add it back

PR #93, opened 2019-05-24
Run Binary Default Flag Change

abejgonzalez:
Have the simulation rules share the `SIM_FLAGS` variable so that people can override them if they want.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -86,10 +86,10 @@ $(output_dir)/%: $(RISCV)/riscv64-unknown-elf/share/riscv-tests/isa/%
 	ln -sf $< $@
 
 $(output_dir)/%.run: $(output_dir)/% $(sim)
-	$(sim) $(PERMISSIVE_ON) $(SIM_FLAGS) $(PERMISSIVE_OFF) $< && touch $@
+	$(sim) $(PERMISSIVE_ON) +max-cycles=$(timeout_cycles) $(PERMISSIVE_OFF) $< && touch $@
 
 $(output_dir)/%.out: $(output_dir)/% $(sim)
-	$(sim) $(PERMISSIVE_ON) $(SIM_FLAGS) $(PERMISSIVE_OFF) $< 3>&1 1>&2 2>&3 | spike-dasm > $@
+	$(sim) $(PERMISSIVE_ON) +verbose +max-cycles=$(timeout_cycles) $(PERMISSIVE_OFF) $< 3>&1 1>&2 2>&3 | spike-dasm > $@
 
 #########################################################################################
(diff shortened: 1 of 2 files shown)


jerryz123:
We don't always want to run `+verbose`, it significantly slows simulation

abejgonzalez:
I personally, like to have the `+verbose` flag on by default. But then again... personal preference. What are other peoples thoughts?

jerryz123:
If you looked at how boom-template worked, `+verbose` was only used for generating .vpd and .out.
https://github.com/riscv-boom/boom-template/blob/4c0a00bc042f0dae27ee8cefffd5d44973a98a89/Makefrag#L55-L66

davidbiancolin:
+1 for defaulting for a faster option if it's materially faster (say >10%).  During debug, the user can add extra flags if they want more information.

abejgonzalez:
So Ill guess, Ill default to just having timeout as the main simulator flags.

It'll be up to the dev to add `+verbose`

colinschmidt:
Yes not verbose by default its way more than 10%.

common.mk
```diff
@@ -86,10 +86,10 @@ $(output_dir)/%: $(RISCV)/riscv64-unknown-elf/share/riscv-tests/isa/%
 	ln -sf $< $@
 
 $(output_dir)/%.run: $(output_dir)/% $(sim)
-	$(sim) $(PERMISSIVE_ON) +max-cycles=$(timeout_cycles) $(PERMISSIVE_OFF) $< && touch $@
+	$(sim) $(PERMISSIVE_ON) $(SIM_FLAGS) $(PERMISSIVE_OFF) $< && touch $@
 
 $(output_dir)/%.out: $(output_dir)/% $(sim)
-	$(sim) $(PERMISSIVE_ON) +verbose +max-cycles=$(timeout_cycles) $(PERMISSIVE_OFF) $< 3>&1 1>&2 2>&3 | spike-dasm > $@
```
colinschmidt: Doesn't this just delete verbose? Since it's not in simflags?

abejgonzalez: Yes. This does... My impression is that we would want to control this ourselves using `SIM_FLAGS` instead of it being hardcoded.

colinschmidt: There is no point in `.out` rules if they don't have verbose on.

abejgonzalez:
Ok. Just changed the `SIM_FLAGS` to not include `+verbose`. @colinschmidt will make a PR later to deal with the extra rules for `.out/.run` files.

abejgonzalez:
As I mentioned @colinschmidt wants to deal with these flags/rules in a separate PR so Ill keep this for now.

PR #95, opened 2019-05-25
MultiRoCC Support

abejgonzalez:
Based on #92 so wait until that is merged.

Allow for different RoCC accelerators to be attached to different harts. In this case, this is only for Hwacha but this can be used for other RoCC accelerators (Systolic Array).

Credit for initial work: @a0u @alonamid

### DIFF
--- a/generators/example/src/main/scala/ConfigMixins.scala
+++ b/generators/example/src/main/scala/ConfigMixins.scala
@@ -106,3 +106,43 @@ class WithGPIOBoomRocketTop extends Config((site, here, up) => {
     top
   }
 })
+
+// ------------------
+// Multi-RoCC Support
+// ------------------
+
+/**
+ * Map from a hartId to a particular RoCC accelerator
+ */
+case object MultiRoCCKey extends Field[Map[Int, Seq[Parameters => LazyRoCC]]](Map.empty[Int, Seq[Parameters => LazyRoCC]])
+
+/**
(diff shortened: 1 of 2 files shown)


generators/example/src/main/scala/ConfigMixins.scala
```diff
@@ -168,3 +176,41 @@ class WithGPIOBoomTop extends Config((site, here, up) => {
...
+ *
+ * For ex:
+ *   Core 0, 1, 2, 3 have been defined earlier
+ *     with hartIds of 0, 1, 2, 3 respectively
+ *   And you call WithMultiRoCCHwacha(Seq(0,1))
+ *   Then Core 0 and 1 will get a Hwacha
+ *
+ * @param harts Seq of harts to specify which will get a Hwacha
+ */
+class WithMultiRoCCHwacha(harts: Seq[Int]) extends Config((site, here, up) => {
```
a0u: Depending on the intended usage, a varargs parameter might be cleaner syntax:
```scala
// Example: WithMultiRoCCHwacha(0, 1)
class WithMultiRoCCHwacha(harts: Int*)
```

abejgonzalez: Good idea.

abejgonzalez:
Got confirmation from @a0u that this was fine.

PR #97, opened 2019-05-28
Move REBAR to master

abejgonzalez:
Move REBAR to master for the ADEPT retreat. This will now the main way to interact with ADEPT lab projects.


jerryz123:
I feel like this should be a force push

abejgonzalez:
@jerryz123 I made merge attempt that I will probably push soon. This was just a test.

PR #99, opened 2019-05-29
FireSim Integration

davidbiancolin:
This is the first step towards making FireSim a better library. To use FireSim out of REBAR you'll still need to use targets in the firechip SUB_PROJECT for now. 

Overview: 
- Adds FireChip as a separate SUB_PROJECT; target-level RTL simulation is currently broken since i have not yet provided a test harness.
- Modifies the existing init scripts, and adds a firesim-setup.sh that wraps build-setup.sh in firesim. 
- Modifies build.sbt to include FireSim-provided deps, and adds the RC hack to allow us to use midas annotations in RC source 

From FireSim:
- Moves most of the code under src/main/scala/firesim into REBAR.
  - includes Targets.scala, TargetConfigs.scala, Generator.scala, and the scalatest suite

From FireChip:
- Moves some target-land tests for the NIC and blkdev into tests/


build.sbt
```diff
@@ -9,28 +11,24 @@ lazy val commonSettings = Seq(
     case _ => MergeStrategy.first}},
   scalacOptions ++= Seq("-deprecation","-unchecked","-Xsource:2.11"),
   libraryDependencies += "org.scalatest" %% "scalatest" % "3.0.5" % "test",
-  libraryDependencies += "org.json4s" %% "json4s-native" % "3.6.1",
+  libraryDependencies += "org.json4s" %% "json4s-jackson" % "3.6.1",
   libraryDependencies += "org.scala-lang" % "scala-reflect" % scalaVersion.value,
-  libraryDependencies += "edu.berkeley.cs" %% "firrtl-interpreter" % "1.2-SNAPSHOT",
+  //libraryDependencies += "edu.berkeley.cs" %% "firrtl-interpreter" % "1.2-SNAPSHOT",
```
abejgonzalez: Just remove no?

davidbiancolin: Yeah. Do we use it anywhere?

abejgonzalez: I don't and nor does the BOOM team? Others maybe? @colinschmidt @zhemao @azidar

colinschmidt: Uh at some point barstools used this

davidbiancolin: Alright, should be easy to reintroduce.

abejgonzalez: Will it be used again? @colinschmidt @jwright6323

generators/firechip/src/main/scala/Generator.scala
```diff
@@ -0,0 +1,133 @@
+//See LICENSE for license details.
```
abejgonzalez: This file seems to be executable. `-rwxrwxr-x`. Can you confirm if this is the case on your end. And if so, can we switch it to something like `-rw-rw-r--`

generators/firechip/src/main/scala/Targets.scala
```diff
@@ -0,0 +1,176 @@
+package firesim.firesim
```
abejgonzalez: This file also seems to be executable. `-rwxrwxr-x`. Can you confirm if this is the case on your end? And if so, can we switch it to something like `-rw-rw-r--`

scripts/init-submodules-no-riscv-tools.sh
```diff
@@ -6,14 +6,27 @@ set -o pipefail
...
 unamestr=$(uname)
 RDIR=$(pwd)
+scripts_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 
 # ignore riscv-tools for submodule init recursive
 # you must do this globally (otherwise riscv-tools deep
 # in the submodule tree will get pulled anyway
 git config --global submodule.riscv-tools.update none
 git config --global submodule.esp-tools.update none
 git config --global submodule.experimental-blocks.update none
```
abejgonzalez: So just to make sure I understand correctly, the user must init the rebar submodules first, then they run the firesim init script. How do we want to document this (put this in the rebar docs? and firesim docs) or does FireSim handle submodule init differently?

abejgonzalez:
LGTM. Runs SmallBoomConfig nicely! Testing on Default for now.

abejgonzalez:
Just waiting on comments being addressed then I think this can be merged in.

build.sbt
colinschmidt: Its currently used to run tests for macro compiler: https://github.com/ucb-bar/barstools/blob/master/build.sbt#L34

abejgonzalez: Do the tests run without this right now if you pull this?

scripts/firesim-setup.sh
```diff
@@ -0,0 +1,18 @@
...
+RDIR=$(pwd)
+scripts_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+sims_dir=$scripts_dir/../sims/
+
+# Reenable the FireSim submodule
+git config --unset submodule.sims/firesim.update || true
+cd $sims_dir
+git submodule update --init firesim
+cd firesim
+./build-setup.sh $@ --library
```
jerryz123: This seems to rebuild riscv-tools. Is this necessary, given that REBAR already packages all the tools?

davidbiancolin: I figure that most people will use `./firesim-setup.sh --fast`, to get a prebuilt toolchain on AWS. I have yet to add a no-toolchain feature to build-setup.sh, but should do that in the future.

scripts/init-submodules-no-riscv-tools.sh
davidbiancolin: Yeah, that's correct. I figure that in many cases people won't want to use firesim (they might be doing local dev, for instance). 

REBAR docs; build-setup.sh has a different flow when initializing firesim-as-top.

generators/firechip/src/main/scala/Targets.scala
davidbiancolin: I will fix this when i get back to my machine.

generators/firechip/src/main/scala/Generator.scala
davidbiancolin: As below.

build.sbt
davidbiancolin: I can just fix this when i get back to my machine.

variables.mk
```diff
@@ -74,6 +74,18 @@ ifeq ($(SUB_PROJECT),hwacha)
 	TB                ?= TestDriver
 	TOP               ?= ExampleRocketSystem
 endif
+# Stand-in firechip variables
```
abejgonzalez: @zhemao What should these be to simulate `firechip` in Chipyard?

davidbiancolin: We've resolved to comment this out temporarily.

PR #108, opened 2019-06-01
Doc Update on Heter. Systems

abejgonzalez:
Explains how to use MultiRoCC and create multi-core systems with both BOOM and RoCC. Also creates a 'Advanced-Usage' section to hold this.

### DIFF
--- a/docs/Advanced-Usage/Heterogeneous-SoCs.rst
+++ b/docs/Advanced-Usage/Heterogeneous-SoCs.rst
@@ -1,0 +1,127 @@
+Heterogeneous SoCs
+===============================
+
+The REBAR framework involves multiple cores and accelerators that can be composed in arbitrary ways.
+This discussion will focus on how you combine Rocket, BOOM and Hwacha in particular ways to create a unique SoC.
+
+Creating a Rocket and BOOM System
+-------------------------------------------
+
+Instantiating an SoC with Rocket and BOOM cores is all done with the configuration system and two specific mixins.
+Both BOOM and Rocket have mixins labelled ``WithNBoomCores(X)`` and ``WithNBigCores(X)`` that automatically create ``X`` copies of the core.
+When used together you can create a heterogeneous system.
+The following example shows a dual core BOOM with a single Rocket.
+
+.. code-block:: scala
+
+    class DualCoreBoomAndRocketConfig extends Config(
+      new WithNormalBoomRocketTop ++
+      new WithBootROM ++
+      new boom.system.WithRenumberHarts ++
+      new boom.common.WithRVC ++
+      new boom.common.DefaultBoomConfig ++
+      new boom.system.WithNBoomCores(2) ++
+      new freechips.rocketchip.subsystem.WithoutTLMonitors ++
+      new freechips.rocketchip.subsystem.WithNBigCores(1) ++
+      new freechips.rocketchip.system.BaseConfig)
+
+In this example, the ``WithN...Cores(X)`` mixins set up particular parameters for the duplicated core.
+For example, the Rocket tiles will be "BigCores" while BOOM has some generic parameters set in its mixin.
+In BOOMs case, we override these base core parameters to create multiple "DefaultBoom"s with calling ``DefaultBoomConfig`` after the multi-core instantiation.
+This mixin applies to all the BOOM cores in the system and sets up the parameters for each.
+
+Great! Now you have a heterogeneous setup with BOOMs and Rockets.
+The final thing you need to make this system work is to renumber the ``hartId``'s of the cores so that each core has a unique ``hartId``.
+This is done with ``WithRenumberHarts`` (which can label the Rocket cores first or the BOOM cores first).
+The reason this is needed is because by default the ``WithN...Cores(X)`` mixin assumes that there are only that type of core in the system.
+Thus, without the renumber mixin, each set of cores is labeled starting from zero.
+
+Another option to create a multi heterogeneous core system is to override the parameters yourself so you can specify the core parameters per core.
+The mixin to add to your system would look something like the following.
+
+.. code-block:: scala
+
+    // create 6 cores (4 boom and 2 rocket)
+    class WithHeterCoresSetup extends Config((site, here, up) => {
+      case BoomTilesKey => {
+        val boomTile0 = BoomTileParams(...) // params for boom core 0
+        val boomTile1 = BoomTileParams(...) // params for boom core 1
+        val boomTile2 = BoomTileParams(...) // params for boom core 2
+        val boomTile3 = BoomTileParams(...) // params for boom core 3
+        boomTile0 ++ boomTile1 ++ boomTile2 ++ boomTile3
+      }
+
+      case RocketTilesKey => {
+        val rocketTile0 = RocketTileParams(...) // params for rocket core 0
+        val rocketTile1 = RocketTileParams(...) // params for rocket core 1
+        rocketTile0 ++ rocketTile1
+      }
+    })
+
+Then you could use this new mixin like the following.
+
+.. code-block:: scala
+
+    class SixCoreConfig extends Config(
+      new WithNormalBoomRocketTop ++
+      new WithBootROM ++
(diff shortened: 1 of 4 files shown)


abejgonzalez:
Note: This is rebased on #125 so it must be merged after that PR.

docs/Advanced-Usage/Heterogeneous-SoCs.rst
```diff
@@ -0,0 +1,127 @@
...
+      new WithBootROM ++
+      new boom.system.WithRenumberHarts ++
+      new boom.common.WithRVC ++
+      new boom.common.DefaultBoomConfig ++
+      new boom.system.WithNBoomCores(2) ++
+      new freechips.rocketchip.subsystem.WithoutTLMonitors ++
+      new freechips.rocketchip.subsystem.WithNBigCores(1) ++
+      new freechips.rocketchip.system.BaseConfig)
+
+In this example, the ``WithN...Cores(X)`` mixins set up particular parameters for the duplicated core.
```
alonamid: This sentence might be a bit confusing. Does this imply default parameters? If so, why do you need the additional (DefaultBOOMConfig mixin)?

docs/Advanced-Usage/Heterogeneous-SoCs.rst
```diff
@@ -0,0 +1,127 @@
...
+For example, the Rocket tiles will be "BigCores" while BOOM has some generic parameters set in its mixin.
+In BOOMs case, we override these base core parameters to create multiple "DefaultBoom"s with calling ``DefaultBoomConfig`` after the multi-core instantiation.
+This mixin applies to all the BOOM cores in the system and sets up the parameters for each.
+
+Great! Now you have a heterogeneous setup with BOOMs and Rockets.
+The final thing you need to make this system work is to renumber the ``hartId``'s of the cores so that each core has a unique ``hartId``.
```
alonamid: Are we assuming we explained ``hartId``s somewhere else in the documentation?

docs/Advanced-Usage/Heterogeneous-SoCs.rst
```diff
@@ -0,0 +1,127 @@
...
+
+.. code-block:: scala
+
+    class SixCoreConfig extends Config(
+      new WithNormalBoomRocketTop ++
+      new WithBootROM ++
+      new WithHeterCoresSetup ++
+      new freechips.rocketchip.system.BaseConfig)
+
+Note, in this setup you most likely don't need the ``WithRenumberHarts`` mixin since you have to assign the ``hartId`` of each tile in the tile parameters yourself.
```
alonamid: "most likely"?

docs/Advanced-Usage/Heterogeneous-SoCs.rst
```diff
@@ -0,0 +1,127 @@
...
+      new freechips.rocketchip.system.BaseConfig)
+
+In this example, the ``WithRenumberHarts`` relabels the ``hartId``'s of all the BOOM/Rocket cores.
+Then after that is applied to the parameters, the ``WithMultiRoCCHwacha(0)`` is used to assign to ``hartId`` zero a Hwacha (in this case ``hartId`` zero is Rocket).
+Finally, the ``WithMultiRoCC`` mixin is called.
+This mixin sets the ``BuildRoCC`` key to use the ``MultiRoCCKey`` instead of the default.
+This must be used after all the RoCC parameters are set because it needs to override the ``BuildRoCC`` parameter.
+If this is used earlier in the configuration sequence, then MultiRoCC does not work.
+
+This mixin can be changed to put more accelerators on more cores by changing the arguments to cover more ``hartId``'s (i.e. ``WithMultiRoCCHwacha(0,1,3,6,...)``).
```
alonamid: Should we mention a the generator approach here if we don't want to hard-code a hartId? (use length or number of tiles to apply the accelerator only to the last tile, or something of that sort)

docs/Advanced-Usage/Heterogeneous-SoCs.rst
abejgonzalez: Hmm. So this is just changed in BOOM land. Now, you have to have both mixins (aka. the With... and the XTypeConfig). I can reword this to something like: "In this example, the with "WithN...Cores(Z)" sets up the default parameters needed to have Z amount of cores. For something like BOOM, we add an extra mixin called "XYZ" that is used to change the parameters to a different size core."

docs/Advanced-Usage/Heterogeneous-SoCs.rst
abejgonzalez: I'm assuming that `hartId` is known by the person using the core. It is in the spec, however, I can add a sentence if you think it is necessary.

docs/Advanced-Usage/Heterogeneous-SoCs.rst
abejgonzalez: So.... if they take the default it will be broken because all cores with get `hartId == 0`. So they will have to set the `hartId`. I will make this clearer.

docs/Advanced-Usage/Heterogeneous-SoCs.rst
abejgonzalez: If I understand you correctly, to do your proposal you need a bunch of extra mixins (WithLastAccel, WithBeginningAccel, etc). IMHO, this is a cleaner approach because it makes sure that the user knows what accel. is attached to what system.

docs/Advanced-Usage/Heterogeneous-SoCs.rst
```diff
@@ -0,0 +1,126 @@
...
+      new freechips.rocketchip.subsystem.WithoutTLMonitors ++
+      new freechips.rocketchip.subsystem.WithNBigCores(1) ++
+      new freechips.rocketchip.system.BaseConfig)
+
+In this example, the ``WithNBoomCores`` and ``WithNBigCores`` mixins set up the default parameters for the multiple BOOM and Rocket cores, respectively.
+However, for BOOM, an extra mixin called ``DefaultBoomConfig`` is added to override the default parameters with a different set of more common default parameters.
+This mixin applies to all BOOM cores in the system and changes the parameters for each.
+
+Great! Now you have a heterogeneous setup with BOOMs and Rockets.
+The final thing you need to make this system work is to renumber the ``hartId``'s of the cores so that each core has a unique ``hartId`` (a ``hartId`` is the hardware thread id of the core).
```
alonamid: works

docs/Advanced-Usage/Heterogeneous-SoCs.rst
```diff
@@ -0,0 +1,126 @@
...
+This is done with ``WithRenumberHarts`` (which can label the Rocket cores first or the BOOM cores first).
+The reason this is needed is because by default the ``WithN...Cores(X)`` mixin assumes that there are only BOOM or only Rocket cores in the system.
+Thus, without the ``WithRenumberHarts`` mixin, each set of cores is labeled starting from zero causing multiple cores to be assigned the same ``hartId``.
+
+Another option to create a multi heterogeneous core system is to override the parameters yourself so you can specify the core parameters per core.
```
alonamid: Maybe we should emphasize that this is an alternative, and not an addition to the previous method? (I think it's clear, but it probably can't hurt to add additional separation?)

docs/Advanced-Usage/Heterogeneous-SoCs.rst
```diff
@@ -0,0 +1,126 @@
...
+      new freechips.rocketchip.subsystem.WithoutTLMonitors ++
+      new freechips.rocketchip.subsystem.WithNBigCores(1) ++
+      new freechips.rocketchip.system.BaseConfig)
+
+In this example, Hwachas are added to both BOOM tiles and to the Rocket tile.
+All with the same Hwacha parameters.
+
+Assigning Accelerators to Specific Tiles with MultiRoCC
+-------------------------------------------------------
+
```
alonamid: Maybe mention that in this example Hwacha is an example of a RoCC accelerator?

Issue #104, opened 2019-05-29
Add CI to ensure submodules point to master branches


jerryz123:
We should have a CI rule to check that the submodule commits this repository references exist on the master branches of their respective repositories.

PR #110, opened 2019-06-01
Bump boom

jerryz123:
The previous hash used was incorrect. We really need #104


abejgonzalez:
Can you also update `rebar-dev` to match what master has?

In the future we should only push to `rebar-dev` first... then to `master`

jerryz123:
Yeah, I forgot the dev branch still existed. I'll close this and re-open

PR #111, opened 2019-06-01
more comments on configs/classes

abejgonzalez:
More documentation!

### DIFF
--- a/generators/example/src/main/scala/ConfigMixins.scala
+++ b/generators/example/src/main/scala/ConfigMixins.scala
@@ -45,9 +45,9 @@ class WithGPIO extends Config((site, here, up) => {
     GPIOParams(address = 0x10012000, width = 4, includeIOF = false))
 })
 
-// -----------------------------------------------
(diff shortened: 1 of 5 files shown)


jerryz123:
I feel like most of these comments just reiterate what the config is doing.
For example,
```
/**
 * Config to setup Rocket parameters and add PWM device/system
 */
```
provides no new information from 
```
class PWMRocketConfig extends Config(
  new WithPWMBoomRocketTop ++
  new BaseRocketConfig)
```

Also the verbosity makes it difficult to see all the configs at a glance.
Perhaps an explanation of the Config system would be better? Teach a man to fish, essentially.

abejgonzalez:
Agreed. There needs to be documentation about the Config/Mixin system (coming this summer). However, I still think that these basic configs should be commented somehow within the code. How that should look so that it is useful is another story. This is me just getting something started to generate conversation.

abejgonzalez:
Closing since we need to find a more elegant solution to documenting the configs.

PR #114, opened 2019-06-03
Check if submodule commits are on master

abejgonzalez:
Addresses #104. Would be better to find a solution that doesn't require you to `init` all the submodules. This is just something that will temporarily work.


.circleci/check-commit.sh
```diff
@@ -0,0 +1,40 @@
...
+git submodule update --init
+status=$(git submodule status)
+
+search () {
+    for submodule in "${submodules[@]}"
+    do
+        echo "Running check on submodule $submodule in $dir"
+        hash=$(echo "$status" | grep $submodule | awk '{print$1}' | grep -o "[[:alnum:]]*")
+        echo "Searching for $hash in origin/master of $submodule"
+        git -C $dir/$submodule log origin/master | grep "$hash" # needs init'ed submodules
```
jerryz123: I think `git branch --contains <commit>` is a neater way to do way.

Issue #38, opened 2018-11-22
Usage of $(CHISEL_ARGS) in Makefile


jimmysitu:
https://github.com/ucb-bar/project-template/blob/d48587b6714bbc128566a5d0fac68496d732aa7c/Makefrag#L33

The generator only takes 5 args, and $(CHISEL_ARGS) seems a redundant args?
Any idea?

abejgonzalez:
@zhemao Do you know if this is used anymore? It looks like this was initially added by you at the creation of the repo?

abejgonzalez:
Confirmed that `rocket-chip` does not use this. So this will be deprecated.

PR #115, opened 2019-06-04
Remove CHISEL_ARGS

abejgonzalez:
Addresses #38


abejgonzalez:
If someone adds chisel and then it is a bug!

PR #116, opened 2019-06-04
Add initial VLSI flow scripts

jerryz123:
This is not usable, but this is all the non-tech specific stuff needed to run the VLSI flow. Tech-specific repos should branch off from here.

### DIFF
--- a/vlsi/Makefile
+++ b/vlsi/Makefile
@@ -1,0 +1,122 @@
+#########################################################################################
+# vlsi makefile
+#########################################################################################
+
+#########################################################################################
+# general path variables
+#########################################################################################
+base_dir=$(abspath ..)
+vlsi_dir=$(abspath .)
+sim_dir=$(abspath .)
+
+#########################################################################################
+# include shared variables
+#########################################################################################
+include $(base_dir)/variables.mk
+
+#########################################################################################
+# import other necessary rules and variables
+#########################################################################################
+include $(base_dir)/common.mk
+
+#########################################################################################
+# vlsi types and rules
+#########################################################################################
+
+#sim_name is unused, but GenerateSimFiles expects it
+sim_name     ?= vcs
+tech_name    ?= 
+tech_dir     ?= $(vlsi_dir)/hammer-$(tech_nam)-plugin/$(tech_name)
+SMEMS_COMP   ?= $(tech_dir)/sram-compiler.json
+SMEMS_HAMMER ?= $(build_dir)/$(long_name).mems.hammer.json
+MACROCOMPILER_MODE ?= -l $(SMEMS_COMP) --use-compiler -hir $(SMEMS_HAMMER)
+OBJ_DIR      ?= $(vlsi_dir)/build
+ENV_YML      ?= $(vlsi_dir)/hammer-cad-plugins/bwrc-env.yml
+INPUT_CONFS  ?= example.yml $(dir $(tech_dir))/bwrc.yml
+HAMMER_EXEC  ?= ./example-vlsi
+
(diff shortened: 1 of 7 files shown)


vlsi/example-vlsi
```diff
@@ -0,0 +1,36 @@
...
+
+def example_place_tap_cells(x: hammer_vlsi.HammerTool) -> bool:
+    x.append('''
+# TODO
+''')
+    return True
+
+class ExampleDriver(CLIDriver):
+
+    def action_map(self) -> Dict[str, Callable[[hammer_vlsi.HammerDriver, Callable[[str], None]], Optional[dict]]]:
```
colinschmidt: This isn't the right way to simply add a step.

vlsi/Makefile
```diff
@@ -0,0 +1,122 @@
...
+# import other necessary rules and variables
+#########################################################################################
+include $(base_dir)/common.mk
+
+#########################################################################################
+# vlsi types and rules
+#########################################################################################
+
+#sim_name is unused, but GenerateSimFiles expects it
+sim_name     ?= vcs
```
colinschmidt: What is this set for?

vlsi/Makefile
```diff
@@ -0,0 +1,122 @@
...
+tech_name    ?= 
+tech_dir     ?= $(vlsi_dir)/hammer-$(tech_nam)-plugin/$(tech_name)
```
colinschmidt: typo, `tech_nam`

vlsi/example-vlsi
jerryz123: This is all taken from @jwright6323 's intech22 branch of the old project-template

jwright6323: 🔥

vlsi/Makefile
jerryz123: https://github.com/ucb-bar/project-template/blob/9622a9757067e9c1dab9a4baf551ce932dc87b54/common.mk#L34-L35

vlsi/example-vlsi
colinschmidt: You just overload `get_extra_par_hooks`
```
def get_extra_par_hooks(self) -> List[HammerToolHookAction]:
  return [hammer_vlsi.HammerTool.make_replacement_hook("place_tap_cells", example_place_tap_cells)]
```

colinschmidt:
Please add a `setup-vlsi.sh` that sources the hammer sourceme and such.

jerryz123:
> Please add a `setup-vlsi.sh` that sources the hammer sourceme and such.

Wait is this actually the right thing to do? The setup scripts are meant to be run only once to setup the environment (`build-toolchains.sh`, `firesim-setup.sh`). Its up to the user to source the correct bash script when they try to do something (`env.sh`, or `sourceme-f1-manager.sh`).

colinschmidt:
Where do we should `vlsi-env.sh` then?

jerryz123:
Do you mean like the `sourceme.sh` in eagle-chip? This seems like a project/user-specific env script that should be committed into the downstream tapeout-specific repos.

colinschmidt:
Ok we can start without one but I think users will get annoyed.

jwright6323:
Future annoyed user checking in

jerryz123:
Testchipip test failing due to know branch issue, has since been resolved on dev.

`vlsi/sourceme.sh` will live in downstream tech-template.
 
On merge I will update the downstream tech-template, and work on merging the downstream tech-template with Beagle.

colinschmidt:
I think we should just merge this sooner rather than later and then make more PRs to fix issues.
The ones I know are outstanding are in the makefile but I'd rather just make them new PRs.

PR #117, opened 2019-06-06
[BEAGLE] Clocking Scheme

abejgonzalez:
New BEAGLE clocking scheme. This PR is to get direct code feedback on clock/reset paths. Naming, overall code organization is subject to change.


abejgonzalez:
The tile clocks are derived this way:
offchip clocks -> mux -> divider -> tile

The uncore clock is determined this way:
offchip clocks -> mux -> out_clk1
out_clk1 -> divider -> out_clk2
(out_clk1, out_clk2) -> mux -> uncore_clk

generators/beagle/src/main/scala/BeaglePeriphery.scala
```diff
@@ -22,85 +22,157 @@ case object PeripheryBeagleKey extends Field[BeagleParams]
...
+  io.rs_clk_out_divisor  := r_rs_clk_out_divisor.io.q
+  io.lbwif_clk_divisor   := r_lbwif_clk_divisor.io.q
+  io.uncore_clk_pass_sel := r_uncore_clk_pass_sel.io.q
+  io.hbwif_rsts          := VecInit(r_hbwif_rsts.map(_.io.q))
+  io.switcher_sel        := r_switcher_sel.io.q
+
+  // connect to mmio
+
+  // needed to space out the mmio correctly
+  require(c.uncoreClkDivBits == 8)
```
colinschmidt: This doesn't seem necessary. You can just have the reg map space them out right?

generators/beagle/src/main/scala/BeaglePeriphery.scala
```diff
@@ -22,85 +22,157 @@ case object PeripheryBeagleKey extends Field[BeagleParams]
...
+
+  // hbwif reset signals
+  val r_hbwif_rsts = Seq.fill(p(HbwifNumLanes)) {
     withReset(io.rst_async) {
       Module(new AsyncResetRegVec(w = 1, init = 1))
     }
   }
-  val switcher_sel = withReset(io.rst_async) {
+
+  // connect hbwif or lbwif to memory system
```
colinschmidt: Should consider whether you want to have each hbwif lane individually useable (in case only 1 of them works)

generators/beagle/src/main/scala/BeaglePeriphery.scala
```diff
@@ -192,42 +265,141 @@ trait HasPeripheryBeagleBundle {
...
+  val uncore_clk_divider = withClockAndReset(uncore_clk_mux_out, uncore_clk_mux_rst) {
+    Module(new testchipip.ClockDivider(outer.scrParams.uncoreClkDivBits))
+  }
+  uncore_clk_divider.io.divisor := scr_mod.io.uncore_clk_divisor
+
+  // mux between the divided uncore and the undivided
+  val uncore_clks_pre_muxed = Seq(uncore_clk_mux_out, uncore_clk_divider.io.clockOut)
+  require(scr_mod.io.uncore_clk_pass_sel.getWidth >= log2Ceil(uncore_clks_pre_muxed.length), "[sys-top] must be able to select the uncore clocks")
+  val uncore_clk_pass_mux = testchipip.ClockMutexMux(uncore_clks_pre_muxed)
+  uncore_clk_pass_mux.io.sel := scr_mod.io.uncore_clk_pass_sel
```
colinschmidt: Just make sure all these select signals come up as passthrough.

generators/beagle/src/main/scala/BeaglePeriphery.scala
```diff
@@ -192,42 +265,141 @@ trait HasPeripheryBeagleBundle {
...
+  require(rs_clk_sel.getWidth >= (log2Ceil(single_clks.length) + log2Ceil(diff_clks.length)), "[sys-top] must be able to select all input clocks")
+  require(uncore_clk_sel.getWidth >= (log2Ceil(single_clks.length) + log2Ceil(diff_clks.length)), "[sys-top] must be able to select all input clocks")
+
+  // setup uncore muxes and divider
+
+  val uncore_clk_mux = testchipip.ClockMutexMux(single_clks ++ diff_clks)
+  uncore_clk_mux.io.sel := uncore_clk_sel
+  uncore_clk_mux.io.resetAsync := rst_async
+  val uncore_clk_mux_out = uncore_clk_mux.io.clockOut
+  val uncore_clk_mux_rst = ResetCatchAndSync(uncore_clk_mux.io.clockOut, reset.asBool)
```
colinschmidt: How is `reset.asBool` different from `rst_async`?

generators/beagle/src/main/scala/BeaglePeriphery.scala
```diff
@@ -192,42 +265,141 @@ trait HasPeripheryBeagleBundle {
...
-
-  // other
-  val clk_sel = outer.scr.module.io.clk_sel
-
-  outer.scr.module.io.boot := boot
-  outer.scr.module.io.rst_async := rst_async
+  val hbwif_mod = outer.hbwif.module
+  hbwif_mod.hbwifResets := scr_mod.io.hbwif_rsts
+  hbwif_mod.resetAsync := rst_async
+  hbwif_mod.hbwifRefClocks := hbwif_clks
```
colinschmidt: Did these get setup here? I didn't see it.

generators/beagle/src/main/scala/TestHarness.scala
```diff
@@ -75,15 +75,18 @@ class BeagleTestHarnessInner(implicit p: Parameters) extends LazyModule
 
     dut.reset := reset
     dut.boot := true.B
-    dut.alt_clks.foreach { _ := clock }
-    dut.alt_clk_sel := 0.U
+    dut.single_clks.foreach { _ := clock }
+    dut.diff_clks.foreach { _ := DontCare }
```
colinschmidt: I think with the behavioral model you could set `diff_clks.n := clock` and `diff_clks.p := !clock`

generators/beagle/src/main/scala/TestHarness.scala
```diff
@@ -75,15 +75,18 @@ class BeagleTestHarnessInner(implicit p: Parameters) extends LazyModule
 
     dut.reset := reset
     dut.boot := true.B
-    dut.alt_clks.foreach { _ := clock }
-    dut.alt_clk_sel := 0.U
+    dut.single_clks.foreach { _ := clock }
```
colinschmidt: Also worth maybe making one of them faster and one of them slower. Maybe the single_clk is 2x slower than diff clock?

generators/beagle/src/main/scala/BeaglePeriphery.scala
abejgonzalez: It serves as a reminder that you need to space it out correctly. But no, this is not strictly necessary.

generators/beagle/src/main/scala/BeaglePeriphery.scala
abejgonzalez: Isn't the normal `reset` the reset that the periphery is set to... aka it is the uncore reset. Then the `rst_async` is the reset gotten from offchip.

generators/beagle/src/main/scala/BeaglePeriphery.scala
abejgonzalez: These are setup from the MMIO and the clocks are "clock_rx"'ed at the ChipTop and passed into here.

generators/beagle/src/main/scala/BeaglePeriphery.scala
abejgonzalez: All the select signals for the muxes come from offchip except for the passthrough mux. That signal is given by a `AsyncResetReg` in the MMIO section, so it should come up correctly?

generators/beagle/src/main/scala/BeaglePeriphery.scala
```diff
@@ -192,42 +265,141 @@ trait HasPeripheryBeagleBundle {
...
+  val bh_clk_mux_out = bh_clk_mux.io.clockOut
+  val bh_clk_mux_rst = ResetCatchAndSync(bh_clk_mux.io.clockOut, reset.asBool)
+
+  val rs_clk_mux = testchipip.ClockMutexMux(single_clks ++ diff_clks)
+  rs_clk_mux.io.sel := rs_clk_sel
+  rs_clk_mux.io.resetAsync := rst_async
+  val rs_clk_mux_out = rs_clk_mux.io.clockOut
+  val rs_clk_mux_rst = ResetCatchAndSync(rs_clk_mux.io.clockOut, reset.asBool)
+
+  // setup clock dividers
```
jwright6323: There's probably a way to make this more DRY with some functions you apply to generate clocks.

abejgonzalez: DRY? Do you mean wrap these in functions so it is easier to read? That should be easy enough to do.

abejgonzalez:
@jwright6323 I dont think those functions are as helpful as the ones I put in (if you are refering to the object functions). I think you really want to, you can add another function to return both a clk and divided clk.

jwright6323:
I mean, do whatever you want- you asked for my input, so I gave it...

abejgonzalez:
I appreciate the input! I'll go ahead and add another function. It'll make it look cleaner at the expense of a couple of lines at the top (which is arguably better than a congested mess at the bottom).

jwright6323:
Yeah, I agree with that line of thought.

PR #118, opened 2019-06-10
have boom depend on testchipip

abejgonzalez:
Have BOOM now depend on `testchipip` so that BOOM and can do TSI based bringup.

Needed for https://github.com/riscv-boom/riscv-boom/pull/283


abejgonzalez:
Actually, after discussion with the BOOM folks. This is not needed.

PR #125, opened 2019-06-21
Rename to chipyard

abejgonzalez:
Rename all instances of REBAR in this repository to the new new "Chipyard".


jerryz123:
IMO `ChipYard` (with that capitalization) looks best.

jwright6323:
`Chipyard` > `ChipYard` IMO. Shipyard is one word.

abejgonzalez:
Since, this seems like there is a bit of controversy around it. We can decide on this in the next Chip(yY)ard meeting.

PR #126, opened 2019-07-01
RC Bump - 50de8a

abejgonzalez:
Bump BOOM, RC, testchipip, firrtl for newer RC version. This will allow us to use the L2 that is newly released.


abejgonzalez:
Should work... merging for now then will get buildbot up and running

PR #128, opened 2019-07-02
Add SiFive InclusiveCache

zhemao:


generators/example/src/main/scala/Configs.scala
```diff
@@ -247,3 +247,12 @@ class RV32BoomAndRocketConfig extends Config(
   new freechips.rocketchip.subsystem.WithRV32 ++
   new freechips.rocketchip.subsystem.WithNBigCores(1) ++
   new freechips.rocketchip.system.BaseConfig)
+
+class RocketL2Config extends Config(
```
abejgonzalez: Can these be put in the BOOM, Rocket and (BOOM and Rocket) areas in the config file respectively? I think we should avoid having alot of configs spread out in a bunch of places.

colinschmidt: Putting all rocket-chip configs in BOOM seems like its going to cause trouble eventually.

abejgonzalez: Sorry if this was not clear. I mean just put these new configs in the sections related to boom/rocket/etc in this example config file.

abejgonzalez: I pushed a fix.

generators/example/src/main/scala/Configs.scala
```diff
@@ -247,3 +253,6 @@ class RV32BoomAndRocketConfig extends Config(
   new freechips.rocketchip.subsystem.WithRV32 ++
   new freechips.rocketchip.subsystem.WithNBigCores(1) ++
   new freechips.rocketchip.system.BaseConfig)
+
+class DualCoreRocketL2Config extends Config(
```
colinschmidt: Can we also add a config that sets parameters in the L2 config?
`Rocket4BankL2` or `Rocket1MBL2` and just have them build in CI?
I think its important to test the number of banks and the number of memory channels. Having both of those parameters set has caused problems in the past and I think was the reason for Howie's rocket-chip PR.

abejgonzalez: Sure. Thats fine with me.

abejgonzalez:
Merging for now with knowing that BOOM is for some reason broken on this bump. Will address later.

colinschmidt:
I think we should file issues about things that we will address later. I filed one about my L2 config complaint, and you should file one about fixing boom with the l2, IMO. Otherwise its very easy to lose track of things.

davidbiancolin:
Issues or it never happened.

abejgonzalez:
Thanks to @jerryz123 this: https://github.com/riscv-boom/riscv-boom/pull/315 should address the issue.

PR #133, opened 2019-07-04
Bump boom for fetch buffer fixes

jerryz123:
Thanks to @bkorpan for the fix


davidbiancolin:
[I added write protection to dev]

PR #134, opened 2019-07-06
Verilator Install Location | Re-add HTIF arg parsing | Pass Errors Correctly in *.out

abejgonzalez:
Added support to change the Verilator install location. Should help with CI efforts.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -76,7 +76,7 @@ verilog: $(sim_vsrcs)
 # helper rules to run simulator
 #########################################################################################
 run-binary: $(sim)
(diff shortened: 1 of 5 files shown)


abejgonzalez:
Updated the PR to include fixing the CI so that Verilator runs correctly and so that tests properly error out (instead of being masked by spike-dasm). Need to check the CI passes AND that there is reasonable output.

abejgonzalez:
#129 Is the PR being reverted here until Verilator can support `+permissive/+permissive-off`

PR #136, opened 2019-07-08
Allow esp-tools to get updated after init script

abejgonzalez:


abejgonzalez:
Note: This is will pass CI (but actually fail it if you look at the tests run) until #134 is added in.

PR #137, opened 2019-07-09
Remote CI Builds

abejgonzalez:
Update CI to build remotely on ferry.millennium.berkeley.edu. Also add hwacha CI tests.


abejgonzalez:
Ok so this is now ready to merge. The CI fails on the "commit on master check" which is fine and on Hwacha tests (which is not fine). Can @a0u @colinschmidt look into the Hwacha bug separately from this PR and we will just merge it in?

PR #140, opened 2019-07-10
remove duplicate blackbox files | general grep

abejgonzalez:
Fixes error in VCS when there are duplicate blackbox files in the simulation.

### DIFF
--- a/sims/vsim/Makefile
+++ b/sims/vsim/Makefile
@@ -58,8 +58,7 @@ VCS_NONCC_OPTS = \
 	+v2k \
 	+vcs+lic+wait \
 	+vc+list \
-	-f $(sim_top_blackboxes) \
-	-f $(sim_harness_blackboxes) \
+	-f $(sim_vcs_blackboxes) \
 	-f $(sim_dotf) \
 	-sverilog \
 	+incdir+$(build_dir) \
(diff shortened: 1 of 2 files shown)


sims/vsim/Makefile
```diff
@@ -75,14 +74,22 @@ VCS_NONCC_OPTS = \
 
 VCS_OPTS = -notice -line $(VCS_CC_OPTS) $(VCS_NONCC_OPTS)
 
+########################################################################################
+# remove duplicate blackboxes
+########################################################################################
+sim_vcs_blackboxes ?= $(build_dir)/firrtl_black_box_resource_files.vcs.f
+
+$(sim_vcs_blackboxes): $(sim_top_blackboxes) $(sim_harness_blackboxes)
+	join -a 1 -a 2 <(sort $(sim_top_blackboxes)) <(sort $(sim_harness_blackboxes)) > $@
```
zhemao: There's a much simpler way to do this that doesn't require parallel pipes.

```
cat $^ | sort | uniq > $@
```

zhemao: LOL, I guess that doesn't work because the blackbox files don't have ending newlines.

jwright6323: You can use `awk '{print $1;}'` to get around the trailing newline issue. Also, `sort -u` would remove the extra `| uniq`, so

```
awk '{print $1;}' $^ | sort -u > $@
```

should work.

jwright6323: @abejgonzalez I confirmed this worked, it's a bit cleaner than the anonymous pipes.

zhemao: Nice

abejgonzalez:
I just want to confirm that CI works then I think its mergable

PR #142, opened 2019-07-12
Update FireSim Instructions

davidbiancolin:
Also rework the simulation section a little bit.


docs/Simulation/FPGA-Accelerated-Simulators.rst
```diff
@@ -0,0 +1,64 @@
+FPGA-Accelerated Simulators
+==============================
+
+FireSim
+-----------------------
+
+`FireSim <https://fires.im/>`__ is an open-source cycle-accurate FPGA-accelerated full-system hardware simulation platform that runs on cloud FPGAs (Amazon EC2 F1).
```
alonamid: Have we decided if we're trying to stick with "cycle-exact" or "cycle-accurate"?

docs/Simulation/FPGA-Accelerated-Simulators.rst
```diff
@@ -0,0 +1,64 @@
...
+    ./scripts/firesim-setup.sh --fast
+
+
+``firesim-setup.sh`` initializes additional submodules and then invokes
+firesim's ``build-setup.sh`` script. ``firesim-setup.sh`` accepts all of the same arguments and
+passes them through to ``build-setup.sh``, adding ``--library`` to properly
+initialize FireSim as a library submodule in chipyard. You may run
+``./sims/firesim/build-setup.sh --help`` to see more options.
+
+In order to build bitstreams, run simulations, or to generate MIDAS-transformed RTL for your
```
alonamid: bitstreams -> FPGA bitstreams ?
Also, I'm not sure we've mentioned MIDAS in this page, so the term "MIDAS-transformed RTL" might seem weird to a new user

docs/Simulation/FPGA-Accelerated-Simulators.rst
```diff
@@ -0,0 +1,64 @@
...
+simulator, you'll need to source one of the following three environments:
```
alonamid: This might be too much information. 
I think it's best too tell them to source `sourceme-f1-manager`, and then half a note/comment paragraph explaining about the other two source file.
I generally advocate for the approach of leading a user through one clear recommended path (rather than describing all the options), and having links/comments describing more expanded options.

docs/Simulation/FPGA-Accelerated-Simulators.rst
```diff
@@ -0,0 +1,64 @@
...
+
+    # OR A minimal environment to run recipes out of sim/ (to invoke MIDAS; run MIDAS-level RTL simulation)generate RTL; transform At the root of your chipyard repo
+    source env.sh
+
+    # OR A complete environment to run local FPGA builds with Vivado
+    source sourceme-f1-full.sh
+
+At this point you're ready to use FireSim with Chipyard. If you're not already
+familiar with FireSim, please refer to the `FireSim Docs <http://docs.fires.im/>`__, and proceed
+through the rest of the tutorial.
```
alonamid: We might want to link them directly to this page:
https://docs.fires.im/en/dev/Running-Simulations-Tutorial/Running-a-Single-Node-Simulation.html

So they will know to skip the setup parts?

docs/Simulation/index.rst
```diff
@@ -1,15 +1,20 @@
...
+visability.
 
 .. toctree::
    :maxdepth: 2
    :caption: Simulators:
 
-   Open-Source-Simulators
-   Commercial-Simulators
-   FPGA-Based-Simulators
+   SW-RTL-Simulators
```
alonamid: nit: SW -> Software ?

docs/index.rst
```diff
@@ -4,7 +4,7 @@
    contain the root `toctree` directive.
 
 Welcome to Chipyard's documentation!
-=================================
+====================================
```
alonamid: lol, catching the important stuff :P

docs/Simulation/index.rst
```diff
@@ -1,15 +1,20 @@
...
 
-Chipyard provides support and integration for multiple simulation flows, for various user levels and requirements.
-In the majority of cases during a digital design development process, a simple software RTL simulation will do.
-When more advanced full-system evaluation is required, with long running workloads, FPGA-accelerated simulation will then become a preferable solution.
-The following pages provide detailed information about the simulation possibilities within the Chipyard framework.
+Chipyard supports two classes of simulation:
+
+#. Software RTL simulation using commercial or open-source (Verilator) RTL simulators 
+#. FPGA-accelerated full-system simulation using FireSim
+
```
alonamid: I would still advocate for including the example use-cases for SW vs. FPGA, because I have a hunch some people will see the O(100 MHz) and try to jump to FPGA, while not realizing that for 90% of use-cases (because many people will just test bare-metal assembly tests) SW sim will do fine.

docs/Simulation/FPGA-Accelerated-Simulators.rst
```diff
@@ -0,0 +1,64 @@
...
+Then clone your full Chipyard repository onto your Amazon EC2 FireSim manager
+instance, and setup your Chipyard repository as you would normally.
+
+When you are ready to use FireSim, initalize it as library in Chipyard by running:
+
+.. code-block:: shell
+
+    # At the root of your chipyard repo
+    ./scripts/firesim-setup.sh --fast
+
```
abejgonzalez: Extra space

docs/Simulation/FPGA-Accelerated-Simulators.rst
abejgonzalez: I would not explain it... rather just make this a work link to a place that does explain it (maybe the GitHub repo)?

docs/Simulation/FPGA-Accelerated-Simulators.rst
abejgonzalez: I think pointing to the tutorial is good.

abejgonzalez:
Seems good to me. Did you confirm that this builds the docs correctly?

docs/index.rst
davidbiancolin: Too many warnings :|

docs/Simulation/index.rst
davidbiancolin: I'm not sure we should hand hold people that don't already know that. I'll make the costs more obvious though.

This was a bit of a knee jerk reaction to the amount of text i've seen reading the docs that explain things that either should be common knowledge (i.e., a new student could ask their non-berkeley colleague about it) or highly google-able. I think if were going to burn words on docs they should probably air on side of being more technical.

davidbiancolin:
> Seems good to me. Did you confirm that this builds the docs correctly?

Ze docs build.

docs/Simulation/index.rst
alonamid: We'll agree to disagree here :).
I've had conversations with at least two different group trying get FireSim setup just to run assembly tests (clearly, FireSim in unnecessary for that). You could argue that "this is common knowledge", or "they should know this", but since we are promoting FireSim (and I don't think anyone is currently promoting the benefits of SW RTL Sim in the outside world), I think people tend to gravitate to the product that was promoted, even if they don't need it.

davidbiancolin: To be clear, i'm not saying necessarily that they should know it, but rather it's information readily available elsewhere and so have other means to learn it.

I worry if we have too much text people won't actually read the docs closely enough.

PR #145, opened 2019-07-12
Stop *.h pollution | BlockDev CI

abejgonzalez:
Quick fix to filter out unwanted files. Though we probably don't want a makefile fix for the long run.


abejgonzalez:
It looks like it is passing CI. The `sim_files.c` that is generated from `GenerateSimFiles` step adds the necessary `.cc` files. Then this just makes sure that the `.h` and `.cc`s are filtered out.

abejgonzalez:
@colinschmidt is there any other concerns with this?

jerryz123:
Can you add a CI test that builds the block device config?

abejgonzalez:
This is actually going to change. The proper way to fix this is to remove the extra calls to add resources in `testchipip` then removing extra files from `GenerateSimFiles`. This should suffice for now though for those who need it. And sure... I can build a block device test to CI. Is there a specific set of tests that I should run?

common.mk
```diff
@@ -53,9 +53,9 @@ $(VERILOG_FILE) $(SMEMS_CONF) $(TOP_ANNO) $(TOP_FIR) $(sim_top_blackboxes): $(FI
 	cd $(base_dir) && $(SBT) "project tapeout" "runMain barstools.tapeout.transforms.GenerateTop -o $(VERILOG_FILE) -i $(FIRRTL_FILE) --syn-top $(TOP) --harness-top $(MODEL) -faf $(ANNO_FILE) -tsaof $(TOP_ANNO) -tsf $(TOP_FIR) $(REPL_SEQ_MEM) -td $(build_dir)"
 	cp $(build_dir)/firrtl_black_box_resource_files.f $(sim_top_blackboxes)
 
-$(HARNESS_FILE) $(HARNESS_ANNO) $(HARNESS_FIR) $(sim_harness_blackboxes): $(FIRRTL_FILE) $(ANNO_FILE) $(sim_top_blackboxes)
```
colinschmidt: I think the reason this dependency was there was to avoid the race condition on both this rule and the one above generating conflicting `.f` files.

abejgonzalez: Ah. I was unaware of this. I'll revert this.

common.mk
```diff
@@ -53,9 +53,10 @@ $(VERILOG_FILE) $(SMEMS_CONF) $(TOP_ANNO) $(TOP_FIR) $(sim_top_blackboxes): $(FI
...
+# note: this depends on sim_top_blackboxes to avoid race condition where both GenerateX calls are done at the same time
+$(HARNESS_FILE) $(HARNESS_SMEMS_CONF) $(HARNESS_ANNO) $(HARNESS_FIR) $(sim_harness_blackboxes): $(FIRRTL_FILE) $(ANNO_FILE) $(sim_top_blackboxes)
 	cd $(base_dir) && $(SBT) "project tapeout" "runMain barstools.tapeout.transforms.GenerateHarness -o $(HARNESS_FILE) -i $(FIRRTL_FILE) --syn-top $(TOP) --harness-top $(VLOG_MODEL) -faf $(ANNO_FILE) -thaof $(HARNESS_ANNO) -thf $(HARNESS_FIR) $(HARNESS_REPL_SEQ_MEM) -td $(build_dir)"
-	grep -v ".*\.cc" $(build_dir)/firrtl_black_box_resource_files.f > $(sim_harness_blackboxes)
+	cp $(build_dir)/firrtl_black_box_resource_files.f $(sim_harness_blackboxes)
```
abejgonzalez: Ok so this is actually an issue. Currently... with the grep it removes any `.cc` files that conflict with Rocket Chip so that the simulator won't complain about name collision. Before the `.*\.cc` the grep just removed `SimSerial` `DTM` `JTAG`. However, I don't think adding a grep here is the right thing to do. I think we should remove the additions in the `GenerateSimFiles` step. What was the initial reasoning in the `GenerateSimFiles` step for adding things like... `SimDTM` `SimJTAG`, etc @grebe

colinschmidt: Do we still have that makefile bit that makes sure the two blackbox `.f` files are unique? If so then the only issue with the `Sim` components is that `emulator.cc` includes all of the possible interfaces and so if the config elaborated doesn't include all interfaces then you will not be able to build a simulator.

abejgonzalez: Ahh. I was unaware of that since I was running VCS sim. Hmm. The simple approach is to "dedup" the two `.f` files... but I think that is something that needs to be cleaner in the end.

colinschmidt: Yeah IIRC there was some github code-golf on the deduping. Maybe its just in some other PR?

abejgonzalez: Yes there was. I will update this PR with the changes.

common.mk
```diff
@@ -33,39 +33,46 @@ $(FIRRTL_JAR): $(call lookup_scala_srcs, $(CHIPYARD_FIRRTL_DIR)/src/main/scala)
...
-REPL_SEQ_MEM = --infer-rw --repl-seq-mem -c:$(MODEL):-o:$(SMEMS_CONF)
+REPL_SEQ_MEM = --infer-rw --repl-seq-mem -c:$(MODEL):-o:$(TOP_SMEMS_CONF)
 HARNESS_REPL_SEQ_MEM = --infer-rw --repl-seq-mem -c:$(MODEL):-o:$(HARNESS_SMEMS_CONF)
 
-$(VERILOG_FILE) $(SMEMS_CONF) $(TOP_ANNO) $(TOP_FIR) $(sim_top_blackboxes): $(FIRRTL_FILE) $(ANNO_FILE)
-	cd $(base_dir) && $(SBT) "project tapeout" "runMain barstools.tapeout.transforms.GenerateTop -o $(VERILOG_FILE) -i $(FIRRTL_FILE) --syn-top $(TOP) --harness-top $(MODEL) -faf $(ANNO_FILE) -tsaof $(TOP_ANNO) -tsf $(TOP_FIR) $(REPL_SEQ_MEM) -td $(build_dir)"
-	cp $(build_dir)/firrtl_black_box_resource_files.f $(sim_top_blackboxes)
+$(TOP_FILE) $(TOP_SMEMS_CONF) $(TOP_ANNO) $(TOP_FIR) $(sim_top_blackboxes): $(FIRRTL_FILE) $(ANNO_FILE)
+	cd $(base_dir) && $(SBT) "project tapeout" "runMain barstools.tapeout.transforms.GenerateTop -o $(TOP_FILE) -i $(FIRRTL_FILE) --syn-top $(TOP) --harness-top $(MODEL) -faf $(ANNO_FILE) -tsaof $(TOP_ANNO) -tsf $(TOP_FIR) $(REPL_SEQ_MEM) -td $(build_dir)"
+	grep -v ".*\.h" $(build_dir)/firrtl_black_box_resource_files.f > $(sim_top_blackboxes)
```
abejgonzalez: You need to only filter `.h` files from this since `.h` files are included in `sim_common_files`.

common.mk
```diff
@@ -33,39 +33,46 @@ $(FIRRTL_JAR): $(call lookup_scala_srcs, $(CHIPYARD_FIRRTL_DIR)/src/main/scala)
...
 
-$(HARNESS_FILE) $(HARNESS_ANNO) $(HARNESS_FIR) $(sim_harness_blackboxes): $(FIRRTL_FILE) $(ANNO_FILE) $(sim_top_blackboxes)
+# note: this depends on sim_top_blackboxes to avoid race condition where firrtl_black_box_resource_files.f is created at the same time
+$(HARNESS_FILE) $(HARNESS_SMEMS_CONF) $(HARNESS_ANNO) $(HARNESS_FIR) $(sim_harness_blackboxes): $(FIRRTL_FILE) $(ANNO_FILE) $(sim_top_blackboxes)
 	cd $(base_dir) && $(SBT) "project tapeout" "runMain barstools.tapeout.transforms.GenerateHarness -o $(HARNESS_FILE) -i $(FIRRTL_FILE) --syn-top $(TOP) --harness-top $(VLOG_MODEL) -faf $(ANNO_FILE) -thaof $(HARNESS_ANNO) -thf $(HARNESS_FIR) $(HARNESS_REPL_SEQ_MEM) -td $(build_dir)"
-	grep -v ".*\.cc" $(build_dir)/firrtl_black_box_resource_files.f > $(sim_harness_blackboxes)
+	grep -v ".*\.h" $(build_dir)/firrtl_black_box_resource_files.f > $(sim_harness_blackboxes)
```
abejgonzalez: Same as the other comment "You need to only filter .h files from this since .h files are included in sim_common_files."

sims/verilator/verilator.mk
```diff
@@ -6,7 +6,7 @@
 # verilator version, binary, and path
 #########################################################################################
 VERILATOR_VERSION      = 4.016
-VERILATOR_INSTALL_DIR ?= verilator
+VERILATOR_INSTALL_DIR ?= verilator_install
```
abejgonzalez: Renamed since the directory this resides in is called `verilator` so the path would be IMO a bad looking path `verilator/verilator/`

sims/verilator/Makefile
```diff
@@ -56,20 +56,29 @@ model_mk_debug = $(model_dir_debug)/V$(VLOG_MODEL).mk
 # build makefile fragment that builds the verilator sim rules
 #########################################################################################
 LDFLAGS := $(LDFLAGS) -L$(RISCV)/lib -Wl,-rpath,$(RISCV)/lib -L$(sim_dir) -lfesvr -lpthread
+SHARED_FLAGS := \
```
colinschmidt: How did you decide what to include in this variable? It appears there are several other "shared" flags between the two invocations. Why not go all in?

abejgonzalez: Oh. So I could put in the `VERILATOR_FLAGS` into the `SHARED_FLAGS`. However, the other flags are different (different directories, one has `--trace`, different `-o` files, different `-include`s). I can make this cleaner if you want.

colinschmidt: Yeah I'd also consider having two `-CFLAGS` arguments so that the shared flags can also absorb the first two flags there.

sims/verilator/Makefile
```diff
@@ -40,6 +40,31 @@ debug: $(sim_debug)
 include $(base_dir)/common.mk
 include $(sim_dir)/verilator.mk
 
+#########################################################################################
```
abejgonzalez: Tried to model how the VCS makefile does things. I think I did most of the split correctly (CC v non-CC).

PR #148, opened 2019-07-14
Add verbose to debug runs

colinschmidt:
If you are generating a vpd you aren't going to go much slower with verbose on, and its a useful debugging tool.

### DIFF
--- a/sims/vsim/Makefile
+++ b/sims/vsim/Makefile
@@ -28,6 +28,8 @@ sim_debug = $(sim_dir)/$(sim_prefix)-$(MODEL_PACKAGE)-$(CONFIG)-debug
 PERMISSIVE_ON=+permissive
 PERMISSIVE_OFF=+permissive-off
 
+WAVEFORM_FLAG=+vcdplusfile=$(sim_out_name).vpd
+
 .PHONY: default debug
 default: $(sim)
 debug: $(sim_debug)
@@ -94,16 +96,10 @@ $(sim_debug) : $(sim_vsrcs) $(sim_dotf) $(sim_vcs_blackboxes)
 	+define+DEBUG -debug_pp
 
 #########################################################################################
-# helper rules to run simulator with debug
-#########################################################################################
-run-binary-debug: $(sim_debug)
-	(set -o pipefail && $(sim_debug) $(PERMISSIVE_ON) $(SIM_FLAGS) +verbose +vcdplusfile=$(sim_out_name).vpd $(PERMISSIVE_OFF) $(BINARY) 3>&1 1>&2 2>&3 | spike-dasm > $(sim_out_name).out)
-
-#########################################################################################
 # create a vcs vpd rule
 #########################################################################################
 $(output_dir)/%.vpd: $(output_dir)/% $(sim_debug)
-	$(sim_debug) $(PERMISSIVE_ON) +verbose +vcdplusfile=$@ +max-cycles=$(timeout_cycles) $(PERMISSIVE_OFF) $<
+	$(sim_debug) $(PERMISSIVE_ON) +max-cycles=$(timeout_cycles) $(SIM_FLAGS) $(VERBOSE_FLAGS) +vcdplusfile=$@ $(PERMISSIVE_OFF) $<
 
 #########################################################################################
 # general cleanup rule
--- a/sims/vsim/Makefile
+++ b/sims/vsim/Makefile
@@ -97,13 +97,13 @@ $(sim_debug) : $(sim_vsrcs) $(sim_dotf) $(sim_vcs_blackboxes)
 # helper rules to run simulator with debug
 #########################################################################################
 run-binary-debug: $(sim_debug)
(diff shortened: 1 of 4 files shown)


sims/vsim/Makefile
```diff
@@ -97,13 +97,13 @@ $(sim_debug) : $(sim_vsrcs) $(sim_dotf) $(sim_vcs_blackboxes)
 # helper rules to run simulator with debug
 #########################################################################################
 run-binary-debug: $(sim_debug)
-	(set -o pipefail && $(sim_debug) $(PERMISSIVE_ON) $(SIM_FLAGS) +vcdplusfile=$(sim_out_name).vpd $(PERMISSIVE_OFF) $(BINARY) 3>&1 1>&2 2>&3 | spike-dasm > $(sim_out_name).out)
+	(set -o pipefail && $(sim_debug) $(PERMISSIVE_ON) $(SIM_FLAGS) +verbose +vcdplusfile=$(sim_out_name).vpd $(PERMISSIVE_OFF) $(BINARY) 3>&1 1>&2 2>&3 | spike-dasm > $(sim_out_name).out)
```
abejgonzalez: I think this should go into the `SIM_FLAGS` variable instead of here.

jerryz123: Why? I don't really understand the purpose of the `SIM_FLAGS` variable. Are they meant to be a common set of flags for all sim runs?

abejgonzalez: No its supposed to be for people to override the variables. They can change the amount of cycles the prog takes to run, add +verbose, remove other variables, add +jtag_rbb_enable, etc.

colinschmidt: The problem with putting it there is that we would need to change `SIM_FLAGS` based on what target the user is running, which is messy.
Is that what you want, or am I misunderstanding?

abejgonzalez: So the default would be to set `SIM_FLAGS=+max-cycles=$(timeout_cycles) +verbose` (like this: https://github.com/ucb-bar/chipyard/blob/20d072c513af7c2273a8f59934449e9bfa01f989/variables.mk#L151). Then both `run-binary` and `run-binary-debug` would have those flags. However, lets say I am doing a JTAG run and I want `+jtag_rbb_enable=1`... all I need to do is run the following `make SIM_FLAGS="+verbose +jtag_rbb_enable=1" run-binary`. IMO `SIM_FLAGS` makes it easier for others to set their own flags rather than change the `Makefile` directly.

colinschmidt: There are 3 levels of verbosity that are important for rocket-chip based simulations. Fast - no output at all. Default - only `+verbose` without a waveform, here you can see what instructions your test executes and what values they are writing/reading. Debug - both `+verbose` and a waveform are generated for full debugging capabilities. 
These levels are orthogonal to flags that affect the function of your simulation such as `+max-cycles` or `+jtag_rbb_enable` therefore these two concepts should not be mixed together.

abejgonzalez: Its not a question of getting them mixed up... I think its a question of allowing someway to override it if the user (in this case us) doesn't want verbose... wants to change amt of timeout cycles... etc. Otherwise, how else is a person supposed to run a simulation without changing the `Makefile`? They would have to manually type the simulator, flags, without the `Makefile`. If that is the intended behavior then we can just remove the helper rules. Am I understanding your point correctly?

jwright6323: If someone wants to turn off verbose in a debug situation, then yes, they should modify the Makefile. It's easy to do and will never happen. There's a reason this feature has existed in project-template without complaint for a long time. Apologies for the ad absurdum argument but if you keep following this pattern the makefile will eventually look like:

```
something:
    $(DO_SOMETHING)
```

and will be no more useful than running the commands directly on the command line.

colinschmidt: But you are making the easy things hard. There are certain expectations users will have that I believe we should meet. If I request a `debug` run I want as much detail as possible. If I ask you to run a test it should have a reasonable timeout. If I want to change these things it should be possible but it being slightly more difficult is ok.
Here is a proposed solution:
We add a `VERBOSE_FLAG` variable that defaults to `+verbose` and is excluded in the `fast` modes.
We have always add timeout_cycles to the flags and note that a timeout_cycles of (0/-1) corresponds to no timeout.
We have `SIM_FLAGS` for other assorted flags the user wishes to pass to the simulator, which now is empty by default.

abejgonzalez: @jwright6323 I hope it never reaches that point but I can see why that might be the fear. @colinschmidt I think we can just do what you initially said. Just hardcode the `+verbose`, keep `SIM_FLAGS=+max_cycles`. And the user can override `SIM_FLAGS` if they want to add more flags to the sim.

jwright6323: I think the `VERBOSE_FLAG` is a good solution.

abejgonzalez: Go ahead and change the PR to what you propose (although I thought that you all were opposed to adding more variables?)

jwright6323: I am opposed to making the user specify `+verbose` when it is the sane default. I am in general opposed to additional make variables, but this seems like a reasonable compromise that adds value, even if it is slight.

colinschmidt: I will add the VERBOSE_FLAG option later and I'd like to remind everyone about this:
https://xkcd.com/1205/

colinschmidt:
Almost got away with more features in fewer lines of code :(

Issue #152, opened 2019-07-15
Re-add `-q` flag to vcs invocation


colinschmidt:
We don't need to see all that garbage, so we should quiet it down.

Issue #151, opened 2019-07-15
Re-add run-fast target


colinschmidt:
`run-fast` was a convenience target for
`make run-asm-tests-fast run-bmark-tests-fast`
We should re-add it.

Issue #144, opened 2019-07-12
Rename sims/vsim -> sims/vcs; sims/verisim -> sims/verilator


jwright6323:
Now that `sims` live in their own directory, it makes more sense to rename `vsim` to `vcs` and `verisim` to `verilator`. `vsim` just meant verilog simulation, but both simulators are doing that.

PR #154, opened 2019-07-15
Address misc. issues

abejgonzalez:
Addresses #152 #151 #144


abejgonzalez:
I think #148 and #137 should be merged in first so that people don't have to port changes over to a new folder and so that I only have to change the CI once.

PR #155, opened 2019-07-15
Temp Fix to Harness Mem Rename Issue

alonamid:
A temporary fix to the renaming issue in test harness memories.
A more permanent fix will likely require barstools and firrtl modifications (see issue https://github.com/ucb-bar/barstools/issues/60 of barstools)


abejgonzalez:
This doesn't change the generated verilog and thus still errors. Closing for now until a better solution is found.

colinschmidt:
I'm surprised this didn't fix the problem. Are you sure you had a clean enough rebuild?

abejgonzalez:
This fix changes the `mems.conf` file but doesn't change the generated verilog. So the `mems.conf` generates an `inTestHarness` version of the mems... but the verilog doesn't match. Looking at the CI this is also the case. So Im fairly certain this is not a "clean" directory fix.

colinschmidt:
Ok the point of this was so that when you pass the modified conf to macrocompiler it will name the memories the "right" thing. But if thats not happening there clearly is some bug.

abejgonzalez:
The macrocompiler is fine. We just need something else to name the memories in the the FIRRTL step. Howie seems to have a solution that he is going to put up.

jwright6323:
@abejgonzalez can you send me a path offline?

PR #156, opened 2019-07-15
Don't replace memories in TestHarness RTL

zhemao:
Having an additional `*.harness.mems.v` file creates naming conflicts, and there's no reason to separate mems in the TestHarness verilog anyway, as it's not meant to be synthesized.


jwright6323:
This was done intentionally because you can't have memories larger than 2GiB in VCS.

zhemao:
Rats. I guess we'll have to figure out some other workaround then.

PR #157, opened 2019-07-16
Toolchains build update

alonamid:
Revamps the toolchains build scripts to build individual components of the toolchains, and play better with FireSim:
-Build linux gnu toolchain, which is required for FireSim.
-Check for conflict with FireSim pre-built toolchain
-Don't build OpenOCD if this is a FireSim build.
-Move from riscv-tools to rocket-tools (since riscv-tools does not include gnu-toolchain anymore)

Depends on FireSim PR 317 (https://github.com/firesim/firesim/pull/317)

- [x] Need to make decisions regarding fesvr (rocket-tools/riscv-tools have merged it into spike. We have been holding on to an older commit of riscv-tools)


alonamid:
@davidbiancolin , currently a user must specify `--firesim` in the `build-toolchains.sh` script if they would like to get the pre-compiled toolchains. Is that reasonable, or should we try to embedded it in `firesim-setup.sh`?

scripts/build-toolchains.sh
```diff
@@ -8,15 +10,83 @@ unamestr=$(uname)
...
-  if [ $1 == "riscv" ]; then
-    TOOLCHAIN="riscv-tools"
-  elif [ $1 == "hwacha" ]; then
-    TOOLCHAIN="esp-tools"
-  fi
-else
-  TOOLCHAIN="riscv-tools"
+function usage
+{
+    echo "usage: ./scripts/build-toolchains.sh [riscv] [hwacha] [ firesim | --firesim] [--submodules-only]"
```
abejgonzalez: missing `ec2fast` in the usage?

abejgonzalez: also do you say ... `firesim` at any point in the args

scripts/build-toolchains.sh
```diff
@@ -26,13 +96,61 @@ RISCV="$(pwd)/$INSTALL_DIR"
...
-echo "export RISCV=$RISCV" > env.sh
+echo "export CHIPYARD_TOOLCHAIN_SOURCED=1" > env.sh
+echo "export RISCV=$RISCV" >> env.sh
 echo "export PATH=$RISCV/bin:$RDIR/$DTCversion:\$PATH" >> env.sh
 echo "export LD_LIBRARY_PATH=$RISCV/lib" >> env.sh
 echo "Toolchain Build Complete!"
+
+
+if [ "$FASTINSTALL" = "false" ]; then
+    # commands to run only on EC2
```
davidbiancolin: Comment needs to be updated

scripts/build-toolchains.sh
```diff
@@ -8,15 +10,83 @@ unamestr=$(uname)
...
+
+#taken from riscv-tools to check for open-ocd autoconf versions
+check_version() {
+    $1 --version | awk "NR==1 {if (\$NF>$2) {exit 0} exit 1}" || (
+        echo $3 requires at least version $2 of $1. Aborting.
+        exit 1
+    )
+}
+
+if [ "$1" == "--help" -o "$1" == "-h" -o "$1" == "-H" ]; then
```
abejgonzalez: Isn't this extra since your `while` statement does this?

scripts/build-toolchains.sh
davidbiancolin: It's going to be removed now.

scripts/build-toolchains.sh
```diff
@@ -8,15 +10,83 @@ unamestr=$(uname)
 RDIR=$(pwd)
```
abejgonzalez: nit: can we call `RDIR` something more descriptive like `INSTALL_DIR`

scripts/build-toolchains.sh
alonamid: Forgot that from an earlier iteration. I pushed a fix

abejgonzalez: do you ever use `--submodules-only`

scripts/build-toolchains.sh
```diff
@@ -8,15 +10,83 @@ unamestr=$(uname)
...
+    shift
+done
+
+
+if [ "$EC2FASTINSTALL" = "true" ]; then
+    if [ "$TOOLCHAIN" = "riscv-tools" ]; then
+      cd $RDIR
+      git clone https://github.com/firesim/firesim-riscv-tools-prebuilt.git
+      cd firesim-riscv-tools-prebuilt
+      git checkout 56a40961c98db5e8f904f15dc6efd0870bfefd9e
```
abejgonzalez: nit: Can you move this hash to an env. variable

scripts/build-toolchains.sh
alonamid: Trying to keep it consistent with the firesim build-setup script?

scripts/init-submodules-no-riscv-tools.sh
```diff
@@ -11,14 +11,14 @@ scripts_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 # ignore riscv-tools for submodule init recursive
 # you must do this globally (otherwise riscv-tools deep
 # in the submodule tree will get pulled anyway
-git config --global submodule.riscv-tools.update none
+git config --global submodule.toolchains/riscv-tools.update none
```
davidbiancolin: this can be handled out of band, but it would be best if we tried not to use  `--global` where possible

.gitmodules
```diff
@@ -13,9 +13,6 @@
 [submodule "tools/firrtl"]
 	path = tools/firrtl
 	url = https://github.com/freechipsproject/firrtl
-[submodule "riscv-tools"]
-	path = toolchains/riscv-tools
-	url = https://github.com/riscv/riscv-tools.git
 [submodule "esp-tools"]
```
abejgonzalez: nit: if `riscv-tools` is now labeled as `toolchains/riscv-tools` can we switch `esp-tools` to match the same `folder/name` structure (and other similar submodules that are under a folder)

scripts/init-submodules-no-riscv-tools.sh
```diff
@@ -11,14 +11,14 @@ scripts_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
...
 git config --global submodule.esp-tools.update none
```
abejgonzalez: nit: see my other nit on "path to git submodules"

scripts/build-toolchains.sh
abejgonzalez: I personally like to error on verbosity rather than something like this... but if this doesn't change that is completely fine.

scripts/build-toolchains.sh
alonamid: fixed

scripts/build-toolchains.sh
alonamid: I think in this case staying consistent is fine

scripts/build-toolchains.sh
alonamid: done

scripts/build-toolchains.sh
alonamid: fixed

.gitmodules
alonamid: done

scripts/init-submodules-no-riscv-tools.sh
alonamid: done

scripts/init-submodules-no-riscv-tools.sh
alonamid: better for and out-of-band discussion with the git experts

jwright6323: Isn't this a pretty simple change / can't you just remove the `--global`?

alonamid: It could be, but then, why was it there to the beginwith? I'm assuming there was some good reason...

jwright6323: In the comments:
```
 # you must do this globally (otherwise riscv-tools deep
 # in the submodule tree will get pulled anyway
```
Where is `riscv-tools` in the submodule tree though? Is this vestigial from when rocket-chip included it?

jwright6323: FWIW I just tried it without global and it behaved as desired.

davidbiancolin: It is vestigial. At one point there three instances of riscv-tools in FireSim's submodule hierarchy and that was the quick and dirty way to deal with it.

We should avoid using `--global` though since it's exactly that -- global. And it pollutes the user's gitconfig.

alonamid: ok, ok, I changed it

scripts/build-static-libfesvr.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
colinschmidt: Please add `usr/bin/env`

scripts/build-static-libfesvr.sh
```diff
@@ -0,0 +1,17 @@
...
+
+if [ "x$RISCV" = "x" ]
+then
+  echo "Please set the RISCV environment variable to your preferred install path."
+  exit 1
+fi
+
+set -e
+
+objs=$(head -n 1 <(make -f <( echo -e 'include Makefile\n$(info $(value fesvr_objs))') -n))
```
colinschmidt: Wow this really is gross.

scripts/build-toolchains.sh
```diff
@@ -8,31 +10,149 @@ unamestr=$(uname)
...
+            ;;
+        hwacha)
+            TOOLCHAIN="esp-tools"
+            ;;
+        ec2fast | --ec2fast) # I don't want to break this api
+            EC2FASTINSTALL=true
+            ;;
+        -h | -H | --help)
+            usage
+            exit
```
colinschmidt: If we're gonna use weird exit values at least make them consistent for the same failure case. I.e. this should be `exit 3` based on the if statement before this while.

scripts/build-toolchains.sh
```diff
@@ -8,31 +10,149 @@ unamestr=$(uname)
...
+    cp HASH "$RISCV"
+    cd $RDIR
+    rm -rf firesim-riscv-tools-prebuilt
+else
+    mkdir -p "$RISCV"
+    git -C $CHIPYARD_DIR submodule update --init --recursive toolchains/$TOOLCHAIN #--jobs 8
+    cd "$CHIPYARD_DIR/toolchains/$TOOLCHAIN"
+    export MAKEFLAGS="-j16"
+    #build the actual toolchain
+    #./build.sh
```
colinschmidt: We inlined this to avoid building openocd?

scripts/build-static-libfesvr.sh
alonamid: This was just pulled from firesim into chipyard. I don't have any intention of touching this :)

scripts/build-toolchains.sh
alonamid: Yes, and since we need to build the linux version of the toolchain as well anyways

scripts/build-static-libfesvr.sh
alonamid: done

scripts/build-toolchains.sh
alonamid: done

colinschmidt:
Are you trying to fix the esp-toolchain CI failure?

PR #159, opened 2019-07-16
Make verilator wget safe

abejgonzalez:


abejgonzalez:
Thanks @dkohlbre for the tip

abejgonzalez:
Passed all tests (- hwacha tests) which is known to break.

PR #161, opened 2019-07-18
Add Hwacha generator in example subproject

abejgonzalez:
This is my quick solution to generating Hwacha tests for a particular config. Just create a new `Generator` that needs to be called with the `make` variables to instantiate Hwacha tests in the new setup.

For ex: `make CONFIG=HwachaBoomConfig GENERATOR=GeneratorWithHwacha`.

If anyone has recommendations on how to do this better that would be great! I basically want to create a generator that is able to detect what RoCC accelerator is on `hartId == 0` and bring in the tests for that accelerator (if any). The problem with doing this right now is that you have to code around whether you are using `MultiRoCC` or just `BuildRoCC` to instantiate the design. I can do something like... if BuildRoCC is a map (aka using `MultiRoCC`)... find out what `hart0` is assigned to and add tests, otherwise, normally check what `BuildRoCC` is. However, this seems like it can get very gnarly...

Maybe have the RoCC accelerators add the tests if they are instantiated? For that matter why not have the cores do this?

Was there a reason why the target design didn't add these tests? I feel like this is a decision that was made in the past.

### DIFF
--- a/generators/example/src/main/scala/Generator.scala
+++ b/generators/example/src/main/scala/Generator.scala
@@ -4,6 +4,7 @@ import chisel3._
 
 import freechips.rocketchip.config.{Parameters}
 import freechips.rocketchip.util.{GeneratorApp}
+import freechips.rocketchip.system.{TestGeneration}
 
 import boom.system.{BoomTilesKey, TestSuiteHelper}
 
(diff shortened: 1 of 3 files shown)


colinschmidt:
Did we lose `GENERATOR_PROJECT`? If not then hwacha's generator is setup to "just work" if you set `GENERATOR_PROJECT=hwacha`

colinschmidt:
As to the why we don't add the tests in the target design, I believe there is a difficult to resolve control flow problem there, but perhaps its worth investigating again.

abejgonzalez:
`GENERATOR_PROJECT` == `GENERATOR_PACKAGE` since project is not super apparent what it means (I changed this). The problem is that when we do this... we also want to get BOOM's tests (Hwacha only uses Rockets tests). Thus, a new generator was created.

jwright6323:
@abejgonzalez you gotta stop renaming things just because you don't like the names. Although you could argue that they make about the same amount of sense, "Project" is the terminology SBT uses, not package, which is where this variable gets its name. Moving between equally sane names is a net productivity loss.

colinschmidt:
Ok. FYI you didn't use the new generator in this PR yet. Is that intentional?

abejgonzalez:
@jwright6323 of course. I had no opposition at the time to argue against it hence the change. Again, if people feel strongly about names I can revert this. @colinschmidt Yes. The current Hwacha CI uses Rocket Chip, so your generator is just fine (it uses your `SUB_PROJECT` flag). However, if I were to use a `CONFIG` with BOOM + Hwacha things would not work.

jwright6323:
I think they're equally fine names; don't bother changing them back for me. I only commented because this is not the first instance of such a change.

generators/example/src/main/scala/Generator.scala
```diff
@@ -24,3 +25,34 @@ object Generator extends GeneratorApp {
...
+    import hwacha.HwachaTestSuites._
+    TestGeneration.addSuites(rv64uv.map(_("p")))
+    TestGeneration.addSuites(rv64uv.map(_("vp")))
+    TestGeneration.addSuite(rv64sv("p"))
+    TestGeneration.addSuite(hwachaBmarks)
+  }
+
+  override def generateTestSuiteMakefrags {
+    addTestSuites
+    var frag = TestGeneration.generateMakefrag + "\nSRC_EXTENSION = $(base_dir)/hwacha/$(src_path)/*.scala" + "\nDISASM_EXTENSION = --extension=hwacha"
```
abejgonzalez: Need to double check this line... but I think this breaks in Chipyard since `hwacha` is under the `generators` section and not in the root of the repo.

abejgonzalez:
Remove in favor of #284

PR #168, opened 2019-07-20
Coordinate Top and Harness generation

albert-magyar:
This avoids name conflicts between the design top and the test harness by coordinating the two FIRRTL compiler runs into one invocation of the main generator app. It also makes Verilog simulator generation somewhat faster.

Coordinated with ucb-bar/barstools#63

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -47,16 +47,13 @@ $(FIRRTL_FILE) $(ANNO_FILE): $(SCALA_SOURCES) $(sim_files)
 # create verilog files rules and variables
 #########################################################################################
 REPL_SEQ_MEM = --infer-rw --repl-seq-mem -c:$(MODEL):-o:$(TOP_SMEMS_CONF)
-HARNESS_REPL_SEQ_MEM = --infer-rw --repl-seq-mem -c:$(MODEL):-o:$(HARNESS_SMEMS_CONF)
+HARNESS_CONF_FLAGS = -thconf $(HARNESS_SMEMS_CONF)
 
-$(TOP_FILE) $(TOP_SMEMS_CONF) $(TOP_ANNO) $(TOP_FIR) $(sim_top_blackboxes): $(FIRRTL_FILE) $(ANNO_FILE)
-	cd $(base_dir) && $(SBT) "project tapeout" "runMain barstools.tapeout.transforms.GenerateTop -o $(TOP_FILE) -i $(FIRRTL_FILE) --syn-top $(TOP) --harness-top $(MODEL) -faf $(ANNO_FILE) -tsaof $(TOP_ANNO) -tsf $(TOP_FIR) $(REPL_SEQ_MEM) -td $(build_dir)"
-	grep -v ".*\.h" $(build_dir)/firrtl_black_box_resource_files.f > $(sim_top_blackboxes)
+TOP_TARGETS = $(TOP_FILE) $(TOP_SMEMS_CONF) $(TOP_ANNO) $(TOP_FIR) $(sim_top_blackboxes)
+HARNESS_TARGETS = $(HARNESS_FILE) $(HARNESS_SMEMS_CONF) $(HARNESS_ANNO) $(HARNESS_FIR) $(sim_harness_blackboxes)
 
-# note: this depends on sim_top_blackboxes to avoid race condition where firrtl_black_box_resource_files.f is created at the same time
-$(HARNESS_FILE) $(HARNESS_SMEMS_CONF) $(HARNESS_ANNO) $(HARNESS_FIR) $(sim_harness_blackboxes): $(FIRRTL_FILE) $(ANNO_FILE) $(sim_top_blackboxes)
-	cd $(base_dir) && $(SBT) "project tapeout" "runMain barstools.tapeout.transforms.GenerateHarness -o $(HARNESS_FILE) -i $(FIRRTL_FILE) --syn-top $(TOP) --harness-top $(VLOG_MODEL) -faf $(ANNO_FILE) -thaof $(HARNESS_ANNO) -thf $(HARNESS_FIR) $(HARNESS_REPL_SEQ_MEM) -td $(build_dir)"
-	grep -v ".*\.h" $(build_dir)/firrtl_black_box_resource_files.f > $(sim_harness_blackboxes)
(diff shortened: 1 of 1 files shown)


common.mk
```diff
@@ -47,16 +47,14 @@ $(FIRRTL_FILE) $(ANNO_FILE): $(SCALA_SOURCES) $(sim_files)
 # create verilog files rules and variables
 #########################################################################################
 REPL_SEQ_MEM = --infer-rw --repl-seq-mem -c:$(MODEL):-o:$(TOP_SMEMS_CONF)
-HARNESS_REPL_SEQ_MEM = --infer-rw --repl-seq-mem -c:$(MODEL):-o:$(HARNESS_SMEMS_CONF)
+HARNESS_REPL_SEQ_MEM = -thconf $(HARNESS_SMEMS_CONF)
```
jwright6323: I think this variable needs to change now that it's being used differently.

variables.mk
```diff
@@ -125,10 +125,9 @@ HARNESS_SMEMS_FILE ?= $(build_dir)/$(long_name).harness.mems.v
 HARNESS_SMEMS_CONF ?= $(build_dir)/$(long_name).harness.mems.conf
 HARNESS_SMEMS_FIR  ?= $(build_dir)/$(long_name).harness.mems.fir
 
-sim_files                  ?= $(build_dir)/sim_files.f
-sim_top_blackboxes         ?= $(build_dir)/firrtl_black_box_resource_files.top.f
-sim_harness_blackboxes     ?= $(build_dir)/firrtl_black_box_resource_files.harness.f
-sim_common_files           ?= $(build_dir)/sim_files.common.f
+sim_files          ?= $(build_dir)/sim_files.f
+sim_blackboxes     ?= $(build_dir)/firrtl_black_box_resource_files.sim.f
+sim_common_files   ?= $(build_dir)/sim_files.common.f
```
jwright6323: Don't we still need a harness blackbox file? E.g. with a hbwif module you still have non-SRAM blackboxes in the harness.

abejgonzalez: My impression is that they would just be added in the `sim.f` file. The current `blackbox.f` file seems to match the `sim.f` version, so I think they are similar if not the same.

abejgonzalez:
Also ran this with the systolic project that initially caused the issue and this seemed to work.

variables.mk
jwright6323: It's not clear to me which file is `sim.f` because there are multiple files ending in `.f` that contain `sim`. `firrtl_black_box_resource_files.sim.f` ?

What is the difference between `$(sim_files)` and `$(sim_blackboxes)` ?

abejgonzalez: Sorry. By `sim.f` I mean `.sim.f`. So I am referring to the `firrtl_black_box_resource_files.sim.f`. 

The difference between the two is that the `sim_blackboxes` is the file generated by the RTL during elaboration/firrtl. The `sim_files` is generated by the `GenerateSimFiles` step (run before elaboration). It adds all necessary files needed for either Verilator or VCS in the `sim_files.f`.

However, these two files (`sim_blackboxes` and `sim_files`) can conflict so thats why we need the dedup step.

jwright6323: So what lives in `$(sim_common_files)` then? Is it things that are common to both the top and harness or things that are in both `$(sim_files)` and `$(sim_blackboxes)`?

We separate out the top and harness blackboxes because we pass (some of) the top blackboxes to synthesis. Depending on the synthesis settings used, the hierarchy may be preserved and there would be module conflicts between the modules in the top blackbox .f list and the post-synthesis netlist, which is why we want to have a separate harness .f.

abejgonzalez: `$(sim_common_files)` is the all unique files needed for simulation in VCS or Verilator (`.cc`, `.v` files). So it is the second option you listed... things in both `$(sim_files)` and `$(sim_blackboxes)`.

As for the separation, I think that needs to be discussed in the corresponding BARSTOOLS PR.

abejgonzalez:
Adding the "DO NOT MERGE" until the harness/top `.f` discussion is completed.

variables.mk
colinschmidt: @jwright6323 Is there a problem if we pass all of the verilog files in the `.f` files into synthesis? If they aren't instantiated the tools shouldn't have a problem (and if they do we can add appropriate `ifdef SYNTHESIS`).

jwright6323: @colinschmidt I think it's probably ok to pass them all to synthesis, but @albert-magyar highlighted the actual problem- we need a version of the .f for things that exist _only_ in the harness so that we can avoid conflicts with the post-syn netlist.

common.mk
albert-magyar: I updated it, thanks

abejgonzalez:
If I remember correctly... bumping FIRRTL cause a bunch of BOOM "firrtlation" issues where we would run out of memory. I don't know if this is a similar issue to the memoization problem that BOOM encountered in the past.

colinschmidt:
> If I remember correctly... bumping FIRRTL cause a bunch of BOOM "firrtlation" issues where we would run out of memory. I don't know if this is a similar issue to the memoization problem that BOOM encountered in the past.

Do you mean the firrtl issue I worked around here: https://github.com/ucb-bar/hwacha/commit/ff4605f5d10d91b7ae8c4a006b965d4706009a06

abejgonzalez:
Yup. That looks about right. To be honest, I wasn't familiar with the issue since Albert/David solved the issue before the BOOM team saw it.

albert-magyar:
So this seems like just a "bump FIRRTL to head of master" issue. I can reproduce the OOM issues locally, but if I cherry-pick the FIRRTL commit that adds my `.f` features, things build fine.

I need to bisect this.

abejgonzalez:
Can we merge `dev` into this to rerun CI with the new FIRRTL/barstool changes? (You probably also need to combine the `barstools` branches also)

variables.mk
albert-magyar: I forgot to circle back to this comment, but this generates two .f files once again.

abejgonzalez:
This LGTM. I would prefer that the `barstools` PR is merged in first though.

albert-magyar:
So the barstools PR is merged and this is passing all CI checks now.

PR #173, opened 2019-07-24
Switch submodules to https

abejgonzalez:


.gitmodules
```diff
@@ -42,7 +42,7 @@
 	url = https://github.com/ucb-bar/esp-tools.git
 [submodule "vlsi/hammer"]
 	path = vlsi/hammer
-	url = git@github.com:ucb-bar/hammer.git
+	url = https://github.com/ucb-bar/hammer.git
 [submodule "vlsi/hammer-cad-plugins"]
 	path = vlsi/hammer-cad-plugins
-	url = git@github.com:ucb-bar/hammer-cad-plugins
```
jwright6323: This is a nit but we should be consistent in including or omitting `.git`

jwright6323: (this should have been on the other side of the diff)

abejgonzalez: Fixed

PR #176, opened 2019-07-25
[boom] move boom integration into chipyard

abejgonzalez:
This PR moves BOOM integration and configs that involve integration into Chipyard. Configs that are only BOOM related will stay in BOOM. It is up to the BOOM user to use Chipyard to make a full SoC config (specifying top, bootrom, etc).


jerryz123:
Is the plan to merge this into boom-ci, then merge boom-ci into dev?

abejgonzalez:
`boom-ci` is the branch that BOOM CI runs on. It doesn't run on Chipyard `dev`. So if it runs with `boom-ci` then it should be good to merge into `dev`

jerryz123:
Oh sorry misread the merge target branch

generators/example/src/main/scala/BoomConfigs.scala
```diff
@@ -0,0 +1,97 @@
+//******************************************************************************
+// Copyright (c) 2015 - 2019, The Regents of the University of California (Regents).
+// All Rights Reserved. See LICENSE and LICENSE.SiFive for license details.
+//------------------------------------------------------------------------------
+// Author: Christopher Celio, Abraham Gonzalez, Ben Korpan, Jerry Zhao
+//------------------------------------------------------------------------------
```
bkorpan: Nothing else in example has this style of header.

jerryz123:
This reenables DTM configs in chipyard as well.

This LGTM.

abejgonzalez:
@alonamid Do we want to merge this first with the DTM docs before doing your documentation changes. IMO this is the last thing that should be merged in before the dev->master merge.

jerryz123:
I would like 1 more boom bump before the dev->master merge, specifically the breakpoint stuff which lets us remove chip-specific test suite generation.

docs/Advanced-Usage/DTM-Debugging.rst
```diff
@@ -0,0 +1,48 @@
+Debugging with DTM/JTAG
+===============================
+
+By default, Chipyard is not setup to use the Debug Test Module (DTM) to bringup the core.
+Instead, Chipyard uses TSI commands to bringup the core (which normally results in a faster simulation).
```
jerryz123:
```suggestion
Instead, Chipyard uses TSI commands to bringup the core (which normally results in a faster simulation).
TSI simulations use the SimSerial interface to directly write the test binary into memory, while the DTM 
executes a small loop of code to write the test binary byte-wise into memory.
```

PR #177, opened 2019-07-25
Do not clobber LD_LIBRARY_PATH in env.sh

abejgonzalez:
When you run `env.sh` now, it clobbers the `LD_LIBRARY_PATH`. Instead, it should just append to it. Is there a specific reason for not being an append?

### DIFF
--- a/scripts/build-toolchains.sh
+++ b/scripts/build-toolchains.sh
@@ -138,7 +138,7 @@ cd $RDIR
 echo "export CHIPYARD_TOOLCHAIN_SOURCED=1" > env.sh
 echo "export RISCV=$RISCV" >> env.sh
 echo "export PATH=$RISCV/bin:$RDIR/$DTCversion:\$PATH" >> env.sh
-echo "export LD_LIBRARY_PATH=$RISCV/lib" >> env.sh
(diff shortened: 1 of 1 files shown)


scripts/build-toolchains.sh
```diff
@@ -138,7 +138,7 @@ cd $RDIR
 echo "export CHIPYARD_TOOLCHAIN_SOURCED=1" > env.sh
 echo "export RISCV=$RISCV" >> env.sh
 echo "export PATH=$RISCV/bin:$RDIR/$DTCversion:\$PATH" >> env.sh
-echo "export LD_LIBRARY_PATH=$RISCV/lib" >> env.sh
+echo "export LD_LIBRARY_PATH=$RISCV/lib:\$LD_LIBRARY_PATH" >> env.sh
```
a0u: If `LD_LIBRARY_PATH` is not already defined, the new value contains an empty component due to the trailing `:`.  This causes the dynamic linker to also search within the current working directory, which may inadvertently result in a different library file being loaded in some situations.  A safer alternative would be:
```sh
export LD_LIBRARY_PATH="${RISCV}/lib${LD_LIBRARY_PATH:+":${LD_LIBRARY_PATH}"}"
```

PR #181, opened 2019-07-26
Switch to auto-generated hammer makefile fragment

colinschmidt:


vlsi/Makefile
```diff
@@ -15,108 +15,91 @@ sim_dir=$(abspath .)
...
+#########################################################################################
+# AUTO BUILD FLOW
+#########################################################################################
 
-###################################################### DRC ############################################################
-# TODO unimplemented
-.PHONY: drc
-drc: $(OBJ_DIR)/drc-rundir/drc_results.db
+.PHONY: buildfile
+buildfile: $(OBJ_DIR)/hammer.d
```
jwright6323:
```suggestion
# Tip: Set HAMMER_D_DEPS to an empty string to avoid unnecessary RTL rebuilds
# TODO: make this dependency smarter so that we don't need this at all
HAMMER_D_DEPS ?= $(GENERATED_CONFS)
```

colinschmidt:
Good catch @jwright6323

jerryz123:
Is this good to merge? We'd like to start using this for beagle stuff

PR #184, opened 2019-07-30
Only check commit-on-master on master

jerryz123:
We decided this was a bad idea, and that submodules should just avoid rebase-commits.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -439,10 +439,20 @@ jobs:
 # Order and dependencies of jobs to run
 workflows:
     version: 2
+    submodules-on-master:
+        jobs:
+            # Check to make sure submodule commits are on master branches
+            - commit-on-master-check
+        triggers:
+            - schedule:
+                cron: "0 0 * * *"
+                filters:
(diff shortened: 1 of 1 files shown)


colinschmidt:
I thought we decided this was a bad idea until dev was on master, and this should be added back in when that happens?

jerryz123:
> I thought we decided this was a bad idea until dev was on master, and this should be added back in when that happens?

I can redo this so the commit-on-master check only runs on master branch

colinschmidt:
Would it then apply to PRs targeting master as well?

jerryz123:
I'm not sure... Should we ever be bumping submodules in a PR straight to master? That seems dangerous.

I was planning to make this a nightly task on the master branch... I need to do some experimentation to see what the behavior is.

abejgonzalez:
I agree. I think this should only happen on dev->master PR's

jerryz123:
This job will now run only on PRs to master (you can see it did not run for this job).
Once this is merged into master, it will also run once every day according to the cron schedule.

PR #186, opened 2019-08-02
Add dsptools

grebe:


harrisonliew:
I merged this into dev branch upon which I am rebasing the Hydra repository, and it compiles/elaborates. I would suggest also updating the documentation to include references to these new repos, but the plumbing LGTM.

abejgonzalez:
I agree with Harrison, I think this looks good from a technical standpoint. I think there should be some quick outline of docs or something if possible before merging it in (or an issue made so this is kept track of)

grebe:
I added some light documentation (hopefully I didn't screw it up!)

.gitmodules
```diff
@@ -46,3 +46,12 @@
 [submodule "vlsi/hammer-cad-plugins"]
 	path = vlsi/hammer-cad-plugins
 	url = https://github.com/ucb-bar/hammer-cad-plugins.git
+[submodule "tools/dsptools"]
```
alonamid: I was under the impression that dsptools was more on the generators side rather than the tools side. Was I mistaken?

build.sbt
```diff
@@ -43,6 +43,22 @@ def conditionalDependsOn(prj: Project): Project = {
...
 
+/**
+  * It has been a struggle for us to override settings in subprojects.
+  * An example would be adding a dependency to rocketchip on midas's targetutils library,
+  * or replacing dsptools's maven dependency on chisel with the local chisel project.
+  *
+  * This function works around this by specifying the project's root at src/ and overriding
+  * scalaSource and resourceDirectory.
+  */
+def freshProject(name: String, dir: File): Project = {
```
alonamid: I feel like we've tried this before, and it hasn't completely worked for all use-cases. Maybe @davidbiancolin has some additional insights?

.gitmodules
grebe: It's mostly a library, with a few blocks that I'm probably the only user of. I think it belongs closer to chisel than to rocket, but either place would be OK

build.sbt
davidbiancolin: No, this is just a method wrapping around the gloriously hacky hack that we used for rocket-chip. LGTM

abejgonzalez:
I went ahead and put in the `dev` -> `master` PR. So this LGTM

grebe:
Err, this merges into dev. Should it go into master instead?

abejgonzalez:
No this should go into `dev`. The `dev` to `master` merge was to just fix bugs that were in `master`. We will release this with the rest of the repo in Oct. for MICRO.

PR #204, opened 2019-08-16
Make waveforms precious

colinschmidt:


jerryz123:
Can you make `%.vpd` `.PRECIOUS` as well? Those are the output waves from `run-binary-debug`

colinschmidt:
I made the outputs for run-binary-debug for both simulators precious.

PR #206, opened 2019-08-22
Fix firesim test suite generation

zhemao:
These commits are needed in order to fix firesim-as-library compilation. Now example and firechip share the assembly test and benchmark generation.


abejgonzalez:
LGTM once the CI passes

PR #207, opened 2019-08-22
Make devices connect to fbus/pbus instead of sbus

zhemao:
In sifive-blocks, they use fbus and pbus to connect to the memory network. This commit changes the peripherals in icenet and testchipip to do the same.


zhemao:
I just checked this. IceNIC and SerialAdapter are agnostic to the interface widths. The BlockDevice has a requirement that fires if the interface width doesn't match what it expects. We can probably fix that though.

Issue #147, opened 2019-07-13
Make -jN fails


jerryz123:
`make -j2` fails with 
```
[jerryz123@a5:/nscratch/jerryz/rebar2-proj/project-template/sims/vsim]$ make -j2
cd /nscratch/jerryz/rebar2-proj/project-template && java -Dsbt.ivy.home=/scratch/jerryz/.ivy2 -Dsbt.global.base=/scratch/jerryz/.sbt -Dsbt.boot.directory=/scratch/jerryz/.sbt/boot/ -Xmx16G -Xss16M -XX:MaxPermSize=256M -jar /nscratch/jerryz/rebar2-proj/project-template/generators/rocket-chip/sbt-launch.jar ++2.12.4 "project utilities" "runMain utilities.GenerateSimFiles -td /nscratch/jerryz/rebar2-proj/project-template/sims/vsim/generated-src/example.BoomRocketTestHarness.DefaultRocketConfig -sim vcs"
make: *** No rule to make target '/nscratch/jerryz/rebar2-proj/project-template/sims/vsim/generated-src/example.BoomRocketTestHarness.DefaultRocketConfig/example.BoomRocketTestHarness.DefaultRocketConfig.harness.mems.conf', needed by '/nscratch/jerryz/rebar2-proj/project-template/sims/vsim/generated-src/example.BoomRocketTestHarness.DefaultRocketConfig/example.BoomRocketTestHarness.DefaultRocketConfig.harness.mems.v'.  Stop.
make: *** Waiting for unfinished jobs....
```

jwright6323:
The Makefiles need to more explicitly list out all of the collateral that each step produces

jwright6323:
This will probably be fixed (or partially fixed) by @albert-magyar's barstools PR ucb-bar/barstools#63

PR #211, opened 2019-08-26
Cleanup configs | take ownership of Heterogeneous subsystem | improve verilator compile time

jerryz123:
- Boom/Rocket/Hetero configs now reside in correct files
- Configs are flattened and all extend `BaseConfig`
- Config mixins use the full classpath
- SubSystem/System are moved from boom into chipyard
- `BoomRocket` naming is removed
- L2 included by default
- FireSim defaults to LargeBoom
- Use the HasHierarchicalBusTopology mixin
- Verilator now generates multiple small cpp files, instead of few large cpp files


sims/verilator/Makefile
```diff
@@ -96,11 +97,12 @@ $(model_mk_debug): $(sim_vsrcs) $(sim_common_files) $(INSTALLED_VERILATOR)
 #########################################################################################
 # invoke make to make verilator sim rules
 #########################################################################################
+VERILATOR_MAKEFLAGS=-j8
```
abejgonzalez: Can we share this flag with the Verilator build step and `make -j*` the Verilator binary... since that also takes a long time in CI.

jerryz123: No, because verilator compile is just a few minutes, and this flag exists only to deal with the fact that `make -jN` at the top-level makefile is broken (Issue #147)

abejgonzalez: 10mins is a long time!

jerryz123: Isn't the installed verilator cached between runs anyways?

abejgonzalez: Yes it is. Its really not an issue.

sims/verilator/Makefile
```diff
@@ -59,7 +59,8 @@ VERILATOR_NONCC_OPTS = \
 	+define+PRINTF_COND=\$$c\(\"verbose\",\"\&\&\"\,\"done_reset\"\) \
 	+define+STOP_COND=\$$c\(\"done_reset\"\) \
 	--assert \
-	--output-split 20000 \
+        --output-split 10000 \
```
abejgonzalez: Nit: This is off by a space. Also is there any intuition behind these numbers? Or is it just arbitrary?

jerryz123: `output-split` roughly specifies how much to split the simulator cpp into different files (lower number means more files)

`output-split-cfunc` specifies how aggressively to split individual c functions into different functions (lower number means more functions)

abejgonzalez: I'm assuming you just tested this with the `MegaBoomConfig` and chose random numbers to see if it worked?

jerryz123: I actually tested with LargeBoom, but yeah you can do some fine tuning to improve it even further, but I don't see the need for that.

zhemao:
FireSim and FireBoom configurations seem to compile properly.

zhemao:
They also both pass the regression test suite.

PR #215, opened 2019-08-29
Add TraceGen project

zhemao:
This PR adds support for the TraceGen tool, which is used to verify cache coherence. You can now use TraceGen in VCS/Verilator SW simulation or in FireSim SW simulation.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -6,11 +6,10 @@ SHELL=/bin/bash
 #########################################################################################
 # variables to get all *.scala files
 #########################################################################################
-lookup_scala_srcs = $(shell find -L $(1)/ -iname "*.scala" 2> /dev/null)
+lookup_scala_srcs = $(shell find -L $(1)/ -name target -prune -o -iname "*.scala" -print 2> /dev/null)
 
-PACKAGES=$(addprefix generators/, rocket-chip testchipip boom hwacha sifive-blocks sifive-cache example) \
-		 $(addprefix sims/firesim/sim/, . firesim-lib midas midas/targetutils)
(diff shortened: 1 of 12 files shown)


common.mk
```diff
@@ -6,11 +6,10 @@ SHELL=/bin/bash
 #########################################################################################
 # variables to get all *.scala files
 #########################################################################################
-lookup_scala_srcs = $(shell find -L $(1)/ -iname "*.scala" 2> /dev/null)
+lookup_scala_srcs = $(shell find -L $(1)/ -name target -prune -o -iname "*.scala" -print 2> /dev/null)
 
-PACKAGES=$(addprefix generators/, rocket-chip testchipip boom hwacha sifive-blocks sifive-cache example) \
-		 $(addprefix sims/firesim/sim/, . firesim-lib midas midas/targetutils)
-SCALA_SOURCES=$(foreach pkg,$(PACKAGES),$(call lookup_scala_srcs,$(base_dir)/$(pkg)/src/main/scala))
+SOURCE_DIRS=$(addprefix $(base_dir)/,generators sims/firesim/sim)
```
abejgonzalez: If this changes, the documentation needs to be updated that you don't have to add to the `Makefile` the name of the new folder.

abejgonzalez:
I'm hesitant to do an RC bump in this PR, especially since this was not telegraphed to the other project leads. While CI is supposed to check if it "works" we should probably get other project leads on the review list.

zhemao:
Oh oops. Yeah, I definitely didn't mean to change the rocket-chip version. I'll fix that.

zhemao:
@abejgonzalez Can you take another look?

zhemao:
The automatic build had to be restarted because of a transient error. The succeeding CI check is here: https://circleci.com/gh/ucb-bar/chipyard/4971.

PR #216, opened 2019-08-29
Integrate sha3 accelerator

colinschmidt:


abejgonzalez:
So are we tied to using the name `rocc_template` or can we switch the repository name to `sha3`. Is there anything general that is in `rocc_template` that should be moved out of it and to Chipyard i.e. any specific integration items? (I know this is a nitpick but I think we want it to be apparent that the `sha3` accelerator lives in that repo)

harrisonliew:
Just want to give a heads up that as part of the ASAP7 HAMMER demo, I'm pushing this accelerator through the flow and will therefore suggest that these changes be reviewed in the ASAP7 PR (which would have this branch merged in & rebased on dev). Let me know if there are any objections - I can always wait for this to merge first and then rebase.

jerryz123:
Is there any opposition to merging this in now? We need to start finalizing material Lisa's east coast tutorial this week.

colinschmidt:
Assuming we are ok having a cleanup PR later I'm fine with merging this now.

I've moved changed the name as per Abe's suggestion.

PR #220, opened 2019-08-30
Partition build/ into long_name/ segments

### LINKED ISSUE #187: Partition vlsi/build by long_name
We should partition the `vlsi/build` dir by `long_name`, similar to how the `generated-src` and `output` dirs are partitioned.

abejgonzalez:
Fixes #187


abejgonzalez:
Untested... but I believe this is the solution.

PR #223, opened 2019-08-30
L2 CI

### LINKED ISSUE #131: Complex L2 config for CI
Can we also add a config that sets parameters in the L2 config?
`Rocket4BankL2` or `Rocket1MBL2` and just have them build in CI?
I think its important to test the number of banks and the number of memory channels. Having both of those parameters set has caused problems in the past and I think was the reason for Howie's rocket-chip PR.

_Originally posted by @colinschmidt in https://github.com/ucb-bar/chipyard/pull/128/files_

abejgonzalez:
Added L2 CI to test a larger L2 and multiple banks. 

Do we want to test multiple mem. channels?

Dependent on #211 being merged
Fixes #131


jerryz123:
IMO this test won't really affect our RTL. The L2 RTL and its integration is all SiFive-code.

colinschmidt:
We want multiple channels. That isn't tested upstream and you can see Howie had to make a PR to fix something upstream recently.

zhemao:
Multi-channel memory configurations are broken in rocket-chip. There's a PR to fix it, but it hasn't been merged yet.

https://github.com/chipsalliance/rocket-chip/pull/1789

abejgonzalez:
Closing since this needs to be dealt with after https://github.com/chipsalliance/rocket-chip/pull/1789

PR #224, opened 2019-08-31
Improve robustness/portability of toolchain shell scripts

a0u:
Included are miscellaneous changes to address some less known edge cases with the shell scripting and improve clarity and maintainability.

* Parse command-line options with the `getopts` builtin.
* Fix quoting to tolerate paths with whitespace characters.
* Set make jobs based on hardware thread count instead of a hard-coded value (16).

### DIFF
--- a/scripts/build-toolchains.sh
+++ b/scripts/build-toolchains.sh
@@ -12,12 +12,16 @@ RDIR=$(pwd)
 
 PRECOMPILED_REPO_HASH=56a40961c98db5e8f904f15dc6efd0870bfefd9e
 
-function usage
-{
-    echo "usage: ./scripts/build-toolchains.sh [riscv-tools] [esp-tools] [ec2fast | --ec2fast] "
+usage() {
+    echo "usage: ${0} [riscv-tools | esp-tools | ec2fast]"
     echo "   riscv: if set, builds the riscv toolchain (this is also the default)"
     echo "   hwacha: if set, builds esp-tools toolchain"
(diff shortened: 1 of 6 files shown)


scripts/build-toolchains.sh
```diff
@@ -153,8 +152,19 @@ if [ "$FASTINSTALL" = "false" ]; then
...
+        build_project riscv-openocd --prefix="${RISCV}" --enable-remote-bitbang --enable-jtag_vpi --disable-werror
         echo -e "\\nRISC-V OpenOCD installation completed!"
-        cd $RDIR
+        cd "$RDIR"
     fi
 fi
+
+cd "$RDIR"
+
+{
```
alonamid: The reason I had these before the EC2 check was in-case the ec2 check fails because of the some corner case, the sourceme file is still generated and the build essentially finished without a problem

a0u: I made this change since I took `env.sh` being generated is a indication of success, which would not be true if OpenOCD was supposed to build but failed.  Not having `env.sh` is a clear sign to a user not necessarily paying attention to the logs that something went wrong.

I can see the value in having `env.sh` produced earlier since there's no way to resume without rebuilding everything entirely from scratch (which is another issue).  The commit has to be amend anyway since I just noticed `$DTCversion` is never defined.

alonamid: Yeah, I think OpenOCD is used in a very restricted set of use-cases (and most user will not need it), so indicating that the entire build process failed just because of OpenOCD would be any overkill. There is definitely a clear error message when OpenOCD fails

a0u:
Verified on EC2.

PR #226, opened 2019-09-02
Regression test cleanup (resolves #217)

### LINKED ISSUE #217: FireChip TestSuite Cleanup
https://github.com/ucb-bar/chipyard/blob/c63fbfa4602dab11e71e1cba564d14f2e8766905/generators/firechip/src/main/scala/Generator.scala#L24-L67

I believe this can now be refactored into the Generator itself. So now there is no need to create a new trait that gets added.

zhemao:
Delete some of the redundant regression test lists. They are now generated by the Generator code in the utilities package.


colinschmidt:
I don't care about this but the use of this variable is for rocket its a good subset to run that hits nearly all the paths. This was originally intended to speed up things like CI but if our new CI system doesn't need this performance improvement than its fine to remove.

PR #227, opened 2019-09-02
Firesim CI scripts

zhemao:
Add CI jobs to build FireSim MIDAS simulator using Verilator and run regression tests.


.circleci/defaults.sh
```diff
@@ -44,3 +47,5 @@ mapping["boom"]="SUB_PROJECT=example CONFIG=SmallBoomConfig"
 mapping["rocketchip"]="SUB_PROJECT=rocketchip"
 mapping["blockdevrocketchip"]="SUB_PROJECT=example CONFIG=SimBlockDeviceRocketConfig TOP=TopWithBlockDevice"
 mapping["hwacha"]="SUB_PROJECT=example CONFIG=HwachaRocketConfig GENERATOR_PACKAGE=hwacha"
+mapping["firesim"]="DESIGN=FireSim TARGET_CONFIG=FireSimRocketChipConfig PLATFORM_CONFIG=FireSimDDR3FRFCFSLLC4MBConfig"
+mapping["fireboom"]="DESIGN=FireBoom TARGET_CONFIG=FireSimSmallBoomConfig PLATFORM_CONFIG=FireSimDDR3FRFCFSLLC4MBConfig"
```
davidbiancolin: Why not the large one?

.circleci/run-firesim-tests.sh
```diff
@@ -0,0 +1,21 @@
...
+# get remote exec variables
+SCRIPT_DIR="$( cd "$( dirname "$0" )" && pwd )"
+source $SCRIPT_DIR/defaults.sh
+
+export PATH=$LOCAL_VERILATOR_DIR/install/bin:$PATH
+export FIRESIM_ENV_SOURCED=1
+
+SIMULATION_ARGS="${mapping[$1]}"
+
+run_test_suite () {
```
davidbiancolin: Why not the ScalaTests here?

.circleci/defaults.sh
zhemao: Takes too long

.circleci/run-firesim-tests.sh
zhemao: This way of running it is easier to fit into the existing CI method, which is to separate building the verilator model from running the tests. But I suppose I should make the set of tests run here match what's run in ScalaTests.

.circleci/defaults.sh
davidbiancolin: i'd rather use a config that we're actually building into AGFIs. Can we run a smaller subset of tests or something?

.circleci/run-firesim-tests.sh
davidbiancolin: The scala tests check additional things (e.g., that the trace widget is properly capturing the instruction trace. 

Maybe the best approach is to modify the scala tests to use an existing simulator if it's already compiled? That way we can still have the CI generation and simulator stages?

davidbiancolin: We resolved to modify the scala tests in the future to permit reusing a precompiled simulator.

.circleci/do-firesim-build.sh
```diff
@@ -0,0 +1,63 @@
...
+
+clean
+
+# copy over riscv/esp-tools, verilator, and chipyard to remote
+run "mkdir -p $REMOTE_CHIPYARD_DIR"
+run "mkdir -p $REMOTE_VERILATOR_DIR"
+copy $LOCAL_CHIPYARD_DIR/ $SERVER:$REMOTE_CHIPYARD_DIR
+copy $LOCAL_VERILATOR_DIR/ $SERVER:$REMOTE_VERILATOR_DIR
+
+run "cp -r ~/.ivy2 $REMOTE_WORK_DIR"
```
abejgonzalez: Why is this being done... shouldn't the `.ivy` and `.sbt` directories be created from scratch when you invoke `sbt`?

.circleci/do-firesim-build.sh
```diff
@@ -0,0 +1,63 @@
...
+run "cp -r ~/.sbt  $REMOTE_WORK_DIR"
+
+TOOLS_DIR=$REMOTE_RISCV_DIR
+LD_LIB_DIR=$REMOTE_RISCV_DIR/lib
+VERILATOR_BIN_DIR=$REMOTE_VERILATOR_DIR/install/bin
+
+if [ $1 = "hwacha" ]; then
```
abejgonzalez: Will you need to run Hwacha items on FireSim regressions? If so this `if` statement can be removed and the `else` portion can be kept.

.circleci/run-firesim-tests.sh
abejgonzalez: You don't have to follow the model... The original purpose of splitting up the build from the tests was so that BOOM could split its assembly tests and bmark tests and run them in parallel in two containers. Since it looks like Chipyard is using CI as a smoke test then we can just have this script run the build and tests.

.circleci/do-firesim-build.sh
zhemao: The problem is that would cause all the external dependencies to be downloaded again. I'm trying to avoid doing unnecessary work.

.circleci/do-firesim-build.sh
zhemao: We may want to later. Leaving this in here in case we want to add it later. Although the comparison may be different then.

PR #228, opened 2019-09-03
ASAP7 Hammer Demo

harrisonliew:
This adds an example Hammer VLSI flow for the SHA-3 accelerator + a dummy hard macro in the ASAP7 process.

Things to do before this can be merged:
- [x] Update CI with new submodules
- [x] Make DRC/LVS work for ASAP7 (submodule bumps)
- [x] Add to Chipyard's documentation (incl. a tutorial slide deck)
- [x] Fix ns vs ps time unit bug
- [x] Need more elegant way of separating physical top from elaborated top

Coauthored with @dpgrubb13.


scripts/init-vlsi.sh
```diff
@@ -1,11 +1,15 @@
 #!/usr/bin/env bash
-# exit script if any command fails 
+# exit script if any command fails
 set -e
 set -o pipefail
 
 # Initialize HAMMER and CAD-plugins
 git submodule update --init --recursive vlsi/hammer
-git submodule update --init --recursive vlsi/hammer-cad-plugins
```
jerryz123: You need to modify exception for the cad-plugins to the `init-submodules-no-riscv-tools.sh` script, so base installations will not attempt to clone these private repos

vlsi/README.md
```diff
@@ -1,9 +1,72 @@
...
+source $HAMMER_HOME/sourceme.sh
+```
+
+The configuration for the example design is contained in `example.yml` and the entry script with hooks is contained in `example-vlsi`. You may go through Hammer's readme to learn about the supported configuration options and how to write hooks.
+
+In order to install the process, download (and optionally extract) the ASAP7 PDK tarball. Then, edit the key `vlsi.technology.asap7.tarball_dir` if you want Hammer to extract for you or `vlsi.technology.asap7.install_dir` if you have already extracted it.
+
+To synthesize the just the SHA-3 accelerator with the hard macro we have to change the physical top module (this step is not necessary if you are pushing the entire Rocket-chip through the VLSI flow):
+```shell
+export TOP=Sha3AccelwBB
```
jerryz123: Is this really the best way to to the BlackBox example? Why not create a SHA3 config that uses a blackbox, and add the blackbox to the resource files in the SHA3 package?

It would be nice if `make TOP=Sha3Accel CONFIG=Sha3RocketConfig` and `make TOP=Sha3AccelwBB CONFIG=Sha3BBRocketConfig` did what they implied,

vlsi/bwrc-env.yml
```diff
@@ -0,0 +1,7 @@
+mentor.mentor_home: "/tools/mentor"
```
jerryz123: Can we put links to the license servers in the public repo?
Can these be moved to the cad-plugins repos?

vlsi/README.md
harrisonliew: So, a couple issues:

1. I ran into a problem when doing a `make TOP=Sha3Accel CONFIG=Sha3RocketConfig`, basically that firrtl was looking for a `firrtl_black_box_resources.f` file which didn't exist for `TOP=Sha3Accel`, so this was a workaround.

2. I can do the blackboxing in Chisel-land, but the point of doing it this way is help audiences who are using Hammer without Chisel/Firrtl. 

3. The black box we instantiate has nothing to do with the SHA-3 accelerator, so I don't think it belongs in that package.

Let me know your thoughts.

vlsi/bwrc-env.yml
harrisonliew: Well, these should not work for anyone who doesn't have BWRC computing resource access. I don't think paths to license servers is violating any NDA (@jwright6323 please weigh in)? 

The cad plugins should default to blank for these fields so that the user can override.

vlsi/README.md
jerryz123: 1. Hmmm I'm not sure why that file does not exist. 

2. Can we do something like `make syn TOP_FILE=customverilog.v` for a flow which does not use Chisel/FIRRTL at all? I guess there are really three ways to use HAMMER. 1) Ingest only Chisel-generated Verilog. 2) Ingest Chisel-generated Verilog with Chisel-specified black-boxes. 3) Ingest raw verilog. 

3. I think this goes with the previous point. Should we have examples for 1 and 2, 1 and 3, or 1-3.

harrisonliew: 1. There aren't any blackboxes in the Sha3Accel module, so I guess it doesn't get emitted (but I don't know why)

2. `customverilog.v` is already added to `extra_v_includes`, so I'd suggest having a separate `TOP` and `VLSI_TOP ?= $(TOP)` or similar which allows elaborated top to be different from the physical top module? Then all use cases can be covered by pushing through a different module.

jerryz123: 1. @zhemao @abejgonzalez did you deal with this stuff recently? Can you weigh in on the behavior here?

2. This makes sense to me. So if I did `make syn VLSI_TOP=my_custom_top CUSTOM_VERILOG=my_verilog.v` no Chisel/FIRRTL would be invoked?

harrisonliew: I guess we can introduce a `CUSTOM_VERILOG` variable that if non-empty would break all dependencies on Chisel/FIRRTL/SRAM compiler.

scripts/init-vlsi.sh
```diff
@@ -1,11 +1,15 @@
...
+git submodule update --init --recursive vlsi/hammer-cadence-plugins
+git submodule update --init --recursive vlsi/hammer-synopsys-plugins
+git submodule update --init --recursive vlsi/hammer-mentor-plugins
 
 # Initialize HAMMER tech plugin
-git submodule update --init --recursive vlsi/hammer-"$1"-plugin
+if [[ $1 != *asap7* ]] && [[ $1 != *saed32* ]]; then
```
jwright6323: We need to get rid of saed32

harrisonliew: Ok, scrub from main Hammer as well?

jwright6323: Yes- I'll file a separate issue.

vlsi/bwrc-env.yml
abejgonzalez: I think it would be ideal to not have this be in Chipyard. I think from the outside users will think "what is BWRC" and ask questions about what this file is. 

To me this should be something that a downstream repo should have. Brainstorming... maybe have a submodule that holds BWRC configurations... or have this be an env variable that needs to be set where the env variable points to this file that is located somewhere and tell the Chipyard user that they need to point to an environment config file?

jerryz123: I think having this in the cad-plugins directories makes sense, since each cad-plugin corresponds to a license server,

vlsi/README.md
abejgonzalez: 1. I have never built anything without a BB... so I don't know what the behavior is for that... however that should be an easy fix.

2. IMO having Chipyard support a non-Chisel design is not the goal of this project. To me, that arbitrary vlog synthesis/par Makefile should exist in a different top level repo that we can point to in this README or the docs. We should discuss this at the Chipyard meeting. However, if we are to do this I think Jerry's idea makes sense where `CUSTOM_VLOG` is a list of Verilog files to include rather than one file.

vlsi/example.v
```diff
@@ -0,0 +1,154 @@
+// Extracted Sha3Accel w/ a blackbox (a dummy DCO) included inside
```
colinschmidt: Does this work? Not all instances have implementations in here so I would have thought the tools would choke on it.

vlsi/example.yml
```diff
@@ -0,0 +1,139 @@
+# Technology Setup
+# Technology used is ASAP7
+vlsi.core.technology: asap7
+# Specify dir with ASAP7 tarball
+technology.asap7.tarball_dir: "SPECIFY DIR WITH ASAP7 TARBALL"
+
+vlsi.core.max_threads: 12
```
colinschmidt: I think this is rather aggressive for such a simple example. How about 4?

docs/VLSI/HAMMER.rst
```diff
@@ -24,8 +24,8 @@ Hooks are modifications to steps or actions that are programmatically defined in
 Tool Plugins
 ============
 
-Hammer supports separately managed plugins for different CAD tool vendors.
-The types of tools (in there hammer names) supported currently include:
+Hammer supports separately managed plugins for different CAD tool vendors. You may be able to acquire access to the included Cadence, Synopsys, and Mentor Graphics plugins with permission from the respective CAD tool vendor.
```
jwright6323:
```suggestion
Hammer supports separately managed plugins for different CAD tool vendors. You may be able to acquire access to the included Cadence, Synopsys, and Mentor plugins with permission from the respective CAD tool vendor.
```
They dropped the "Graphics" when they were bought by Siemens.

docs/VLSI/HAMMER.rst
```diff
@@ -24,8 +24,8 @@ Hooks are modifications to steps or actions that are programmatically defined in
...
+The types of tools (by HAMMER names) supported currently include:
```
jwright6323: Style is "Hammer", not all-caps "HAMMER"

vlsi/example.yml
```diff
@@ -0,0 +1,139 @@
...
+        GND: "0 V"
+
+# Because the DCO is a dummy layout, we treat it as a physical-only cell
+par.inputs.physical_only_cells_mode: append
+par.inputs.physical_only_cells_list:
+  - ExampleDCO
+
+# SRAM Compiler compiler options
+vlsi.core.sram_generator_tool: "sram_compiler"
+vlsi.core.sram_generator_tool_path: ["SPECIFY LOCATION OF SRAM GENERATOR IN TECH PLUGIN"]
```
jwright6323: Can we use a comment for this?

vlsi/extra_libraries/dco/ExampleDCO_PVT_0P77V_0C.lib
```diff
@@ -0,0 +1,142 @@
...
+                related_power_pin : VDD;
+                related_ground_pin : VSS;
+            }
+        }
+
+        bus ( row_sel_b ) {
+            bus_type : bus_15_to_0;
+            direction : input;
+            capacitance : 0.006;
+            max_transition : 0.039999999999999994;
```
jwright6323: Nit: can we clean up this float string mangling in whatever generated this file?

vlsi/Makefile
```diff
@@ -18,22 +18,30 @@ include $(base_dir)/variables.mk
...
 #########################################################################################
 # general rules
 #########################################################################################
-ALL_RTL = $(TOP_FILE) $(TOP_SMEMS_FILE) $(extra_v_includes)
-extra_v_includes = $(build_dir)/EICG_wrapper.v
+ALL_RTL = $(TOP_FILE) $(TOP_SMEMS_FILE)
+extra_v_includes = $(build_dir)/EICG_wrapper.v $(vlsi_dir)/example.v
+ifneq ($(CUSTOM_VLOG), )
+	VLSI_RTL = $(CUSTOM_VLOG)
+	VLSI_BB = /dev/null
```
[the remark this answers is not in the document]
harrisonliew: I tried doing an `ifdef $(CUSTOM_VLOG)` but it didn't seem to work in the case where `CUSTOM_VLOG` was not set in the shell. Help appreciated here.

vlsi/example.v
harrisonliew: I'm not sure what you mean. This `example.v` is included along with all the other Verilog files elaborated from Chisel, so all the implementations are defined. That said, this does not track any changes to the Sha3 generator.

docs/VLSI/HAMMER.rst
```diff
@@ -34,25 +43,191 @@ The types of tools (in there hammer names) supported currently include:
...
+
+The ``defaults.yml`` file contains tool-specific configuration variables. The defaults may be overridden as necessary.
 
 Technology Plugins
 ==================
 
-Hammer supports separately managed plugins for different technologies.
+Hammer supports separately managed technology plugins to satisfy NDAs. You may be able to acquire access to certain pre-built technology plugins with permission from the technology vendor. Or, to build your own tech plugin, you need at least a ``<tech_name>.tech.json`` and ``defaults.yml``. An ``__init__.py`` is optional if there are any technology-specific methods or hooks to run.
+
+The `ASAP7 plugin <https://github.com/ucb-bar/hammer/tree/master/src/hammer-vlsi/technology/asap7>`__ is a good starting point for setting up a technology plugin. Refer to Hammer's documentation for the schema and setup instructions.
```
alonamid: It might be worth mentioning that it's a good start for setting up a tech plugin because it's an open-source example, and not because you would actually use it in a chip

harrisonliew: I think it goes without saying that you would not be able to tape out anything in ASAP7 since it's purely a predictive PDK.

alonamid: You'd be surprised how many people in the architecture world would require that to be said. It doesn't hurt us to write that sentence, right?

docs/VLSI/Tutorial.rst
```diff
@@ -0,0 +1,137 @@
...
+To elaborate the Sha3RocketConfig (Rocketchip w/ the accelerator) and set up all prerequisites for the build system to push just the accelerator + hard macro through the flow:
+
+.. code-block:: shell
+
+    export MACROCOMPILER_MODE='--mode synflops'
+    export CONFIG=Sha3RocketConfig
+    export VLSI_TOP=Sha3AccelwBB
+    make buildfile
+
+Note that because the ASAP7 process does not yet have a memory compiler, flip-flop arrays are used instead.
```
alonamid: If might be worth structuring this in a way that exlains why each environemnt variable is set:

The `MACROCOMPILER_MODE='--mode synflops' environment variable is required since the ASAP7 process does not yet have a memory compiler. Therefore, flip-flop arrays (`synflops`) are used instead.

The `CONFIG=Sha3RocketConfig` selects the target  generator config in the same manner as the rest of the chipyard framework.

The `VLSI_TOP=Sha3AccelwBB` environment variable indicates to hammer that we are interested in the physical design of only the SHA3 accelerator block (rather than the entire RTL SoC).

docs/VLSI/Tutorial.rst
```diff
@@ -0,0 +1,137 @@
...
+
+example-vlsi
+^^^^^^^^^^^^
+This is the entry script with placeholders for hooks. In the ``ExampleDriver`` class, a list of hooks is passed in the ``get_extra_par_hooks``. Hooks are additional snippets of python and TCL (via ``x.append()``) to extend the Hammer APIs. Hooks can be inserted using the ``make_pre/post/replacement_hook`` methods.
+
+example.yml
+^^^^^^^^^^^
+This contains the Hammer configuration for this example project. Example clock constraints, power straps definitions, placement constraints, and pin constraints are given. Additional configuration for the extra libraries and tools are at the bottom.
+
+First, set ``technology.asap7.tarball_dir`` to where you downloaded the ASAP7 PDK.
```
alonamid: relative path or absolute path? can you give an example with code context here?

docs/VLSI/Advanced-Usage.rst
```diff
@@ -0,0 +1,54 @@
...
+
+Alternative RTL Flows
+---------------------
+The Make-based build system provided supports using Hammer without using RTL generated by Chipyard. To push a custom verilog module through, one only needs to export the following environment variables before ``make buildfile``.
+
+.. code-block:: shell
+
+    export CUSTOM_VLOG=<your verilog files>
+    export VLSI_TOP=<your top module>
+
```
jerryz123: This is not the normal way to assign make variables. 
The way we describe it everywhere else is `make VARIABLE=x`. I think we should keep this consistent.

harrisonliew: Sure... but these two variables in particular only need to be exported once for the design you want to be pushed through? Should I instead ask the user to populate them in the Makefile instead?

vlsi/example.v
alonamid: I would prefer to use the sha3 verilog that is generated by the chipyard generators. I guess that means instantiating the `Sha3Accel` module?

docs/VLSI/Advanced-Usage.rst
jerryz123: Yes. Leaving it in the Makefiles is more consistent with documentation for the rest of the flows.

vlsi/Makefile
```diff
@@ -18,22 +18,30 @@ include $(base_dir)/variables.mk
...
 SMEMS_CACHE        ?= $(tech_dir)/sram-cache.json
 SMEMS_HAMMER       ?= $(build_dir)/$(long_name).mems.hammer.json
 MACROCOMPILER_MODE ?= -l $(SMEMS_CACHE) -hir $(SMEMS_HAMMER)
 OBJ_DIR            ?= $(vlsi_dir)/build
-ENV_YML            ?= $(vlsi_dir)/bwrc-env.yml
-INPUT_CONFS        ?= example.yml $(dir $(tech_dir))/bwrc.yml
+ENV_YML            ?= $(vlsi_dir)/env.yml
+INPUT_CONFS        ?= example.yml
 HAMMER_EXEC        ?= ./example-vlsi
+VLSI_TOP           ?= $(TOP)
```
jerryz123: Out of curiosity, is there a reason we need a separate `VLSI_TOP` variable? IIRC specifying `make TOP=MyBlock` should do the right thing. 

Its fine if we resolve this in a separate PR.

harrisonliew: You can't set `TOP` to a top_module that's not described in the Chisel you're trying to elaborate.

alonamid: (if this comes up as a questions, it's worth writing it explicitly in the docs :) )

docs/VLSI/Tutorial.rst
```diff
@@ -0,0 +1,141 @@
+.. _tutorial:
+
+ASAP7 Tutorial
+==============
+The ``vlsi`` folder of this repository contains an example HAMMER flow with the SHA-3 accelerator and a dummy hard macro in the ASAP7 PDK. It is intended for use with the Cadence and Mentor tool plugins.
```
alonamid: Maybe it would be good to use explicit hammer terminology here:
"This example tutorial uses an ASAP7 Technology plug-in and a Cadence CAD plug-in."
Where does the Mentor tool plug-in fit here?

harrisonliew: Sure. The Mentor tool plug-in is needed for DRC & LVS.

alonamid: So let's say that. Because I think it's reasonable to assume that some people may have Cadence tools but won't have Mentor tools, and they should know how far in the flow they can get

docs/VLSI/HAMMER.rst
```diff
@@ -1,10 +1,14 @@
+.. _hammer:
+
 Core HAMMER
```
abejgonzalez: A general comment is that there is an inconsistency with using "Hammer" and "HAMMER". We should grep/sed and choose a single version of this throughout the docs.

docs/VLSI/HAMMER.rst
```diff
@@ -21,11 +25,18 @@ Hooks
 
 Hooks are modifications to steps or actions that are programmatically defined in a Hammer configuration.
 
+Configuration (Hammer IR)
+=========================
+
+To configure a Hammer flow, supply a set yaml or json configuration files that chooses the tool and technology plugins and versions as well as any design specific configuration options. Collectively, this configuration API is referred to as Hammer IR and can be generated from higher-level abstractions.
```
abejgonzalez:
```suggestion
To configure a Hammer flow, supply a set ``yaml`` or ``json`` configuration files that chooses the tool and technology plugins and versions as well as any design specific configuration options. Collectively, this configuration API is referred to as Hammer IR and can be generated from higher-level abstractions.
```

docs/VLSI/Tutorial.rst
```diff
@@ -0,0 +1,138 @@
...
+
+Prerequisites
+-------------
+
+* Python 3.4+
+* numpy and gdspy packages
+* Genus, Innovus, and Calibre licenses
+* For ASAP7 specifically:
+
+  * Download the `ASAP7 PDK <http://asap.asu.edu/asap/>`__ tarball and do not extract it
```
abejgonzalez: Download to where?

abejgonzalez: Maybe a brief description of how ASAP7 is extracted... when/where?

docs/VLSI/Tutorial.rst
```diff
@@ -0,0 +1,138 @@
...
+
+.. code-block:: shell
+
+    cd vlsi
+    export HAMMER_HOME=$PWD/hammer
+    source $HAMMER_HOME/sourceme.sh
+
+Building the Design
+-------------------
+To elaborate the Sha3RocketConfig (Rocketchip w/ the accelerator) and set up all prerequisites for the build system to push just the accelerator + hard macro through the flow:
```
abejgonzalez:
```suggestion
To elaborate the ``Sha3RocketConfig`` (Rocket Chip w/ the accelerator) and set up all prerequisites for the build system to push just the accelerator + hard macro through the flow:
```

docs/VLSI/Tutorial.rst
```diff
@@ -0,0 +1,138 @@
...
+
+.. code-block:: shell
+
+    make buildfile MACROCOMPILER_MODE='--mode synflops' CONFIG=Sha3RocketConfig VLSI_TOP=Sha3AccelwBB
+
+The ``MACROCOMPILER_MODE='--mode synflops'`` is needed because the ASAP7 process does not yet have a memory compiler. Therefore, flip-flop arrays are used instead.
+
+The ``CONFIG=Sha3RocketConfig`` selects the target generator config in the same manner as the rest of the Chipyard framework. This elaborates a Rocket-chip with the Sha3Accel module.
```
abejgonzalez:
```suggestion
The ``CONFIG=Sha3RocketConfig`` selects the target generator config in the same manner as the rest of the Chipyard framework. This elaborates a Rocket Chip with the Sha3Accel module.
```

docs/VLSI/Tutorial.rst
```diff
@@ -0,0 +1,138 @@
...
+The ``VLSI_TOP=Sha3AccelwBB`` indicates that we are only interested in physical design of the accelerator block. If this variable is not set, the entire SoC will be pushed through physical design. Note that you should not set the ``TOP`` variable because it is used during Chisel elaboration.
+
+For the curious, ``make buildfile`` generates a set of Make targets in ``build/hammer.d``. It needs to be re-run if environment variables are changed. It is recommended that you edit these variables directly in the Makefile rather than exporting them to your shell environment.
+
+Running the VLSI Flow
+---------------------
+
+example-vlsi
+^^^^^^^^^^^^
+This is the entry script with placeholders for hooks. In the ``ExampleDriver`` class, a list of hooks is passed in the ``get_extra_par_hooks``. Hooks are additional snippets of python and TCL (via ``x.append()``) to extend the Hammer APIs. Hooks can be inserted using the ``make_pre/post/replacement_hook`` methods.
```
abejgonzalez: Just to make sure... is "pre/post/replacement" described somewhere? If not, you might want a quick description. Something like "where these methods inject the statements before, after or completely replace the hook"

docs/VLSI/Tutorial.rst
```diff
@@ -0,0 +1,138 @@
...
+
+First, set ``technology.asap7.tarball_dir`` to the absolute path of where the downloaded the ASAP7 PDK tarball lives.
+
+Synthesis
+^^^^^^^^^
+.. code-block:: shell
+
+    ``make syn``
+
+Post-synthesis logs and collateral are in ``build/syn-rundir``. The Raw QoR data is available at ``build/syn-rundir/reports``, and methods to extract this information for design space exploration are a WIP.
```
abejgonzalez:
```suggestion
Post-synthesis logs and collateral are in ``build/syn-rundir``. The raw QoR data is available at ``build/syn-rundir/reports``, and methods to extract this information for design space exploration are a WIP.
```

docs/VLSI/Advanced-Usage.rst
```diff
@@ -0,0 +1,54 @@
+.. _advanced-usage:
+
+Advanced Usage
+==============
+
+Alternative RTL Flows
+---------------------
+The Make-based build system provided supports using Hammer without using RTL generated by Chipyard. To push a custom verilog module through, one only needs to give the following environment variables ``make buildfile`` (or edit them directly in the Makefile).
```
abejgonzalez: Is this sentence wrong? Delete the ``make buildfile``?

vlsi/bwrc-env.yml
edwardcwang: bwrc-env should just live somewhere internal to Berkeley/BWRC IMO.

harrisonliew: Yeah, it's no longer part of this PR.

PR #234, opened 2019-09-08
[Docs] Add source env.sh to quick start.

davidbiancolin:
1. Adds a `source ./env.sh` to quick start guide
2. Adds a gitignore on docs generated output
3. Makes the _What's Next_ section items fit on a single line each.


davidbiancolin:
I think we should consider moving the note about hwacha somewhere else. It probably provides too much detail in this context and makes the quick start less quick.

docs/Quick-Start.rst
```diff
@@ -28,22 +28,27 @@ To build the toolchains, you should run:
 .. Note:: If you are planning to use the Hwacha vector unit, or other RoCC-based accelerators, you should build the esp-tools toolchains by adding the ``esp-tools`` argument to the script above.
   If you are running on an Amazon Web Services EC2 instance, intending to use FireSim, you can also use the ``--ec2fast`` flag for an expedited installation of a pre-compiled toolchain.
 
+Finally, set up Chipyard environment varibles and put the newly built toolchain on your path:
```
abejgonzalez: Spelling: variables

abejgonzalez: Also, I think there should be a "the" after the "set up".

abejgonzalez:
LGTM Other than the small fix. On the next commit could you add a `[skip ci]` to the commit to make sure it doesn't run the Ci.

PR #235, opened 2019-09-10
Updates and additions to documentation

zhemao:
This adds some of the documentation I promised the last Chipyard meeting. 

1. I noticed that the "Adding an Accelerator" section is really out of date, so I updated that to conform to the current API. 
2. The ToC markup in the top-level index was giving warnings to the Sphinx compiler, so I fixed those issues.
3. Filled out the "Memory Hierarchy" section
4. Added an overview of RocketChip as a whole with the diagram I made for CCC.
5. Added detailed documentation on how TileLink and Diplomacy work and what all the different RocketChip widgets do.
6. Added an explanation of the boot process.


docs/TileLink-Diplomacy-Reference/index.rst
```diff
@@ -0,0 +1,29 @@
+TileLink and Diplomacy Reference
+================================
+
+TileLink is the cache coherence and memory protocol used by RocketChip and
+other Chipyard generators. It is how different modules like caches, memories,
+peripherals, and DMA devices communicate with each other.
+
+TileLink is built on top of diplomacy, a framework for exchanging
```
alonamid: Shouldn't TileLink the protocol be independent of diplomacy (or other implementations)? (or isn't that at least the intention of the current revision)

zhemao: True. I guess it should be "RocketChip's TileLink implementation is built on top of diplomacy".

docs/Customization/Memory-Hierarchy.rst
```diff
@@ -1,4 +1,104 @@
...
-TODO: Talk about SiFive Cache, and integration with L1 and backing main memory models
-(maybe even Tilelink)
+
+The L1 Caches
+--------------
+
+Each CPU tile has an L1 instruction cache and L1 data cache. The size and
+associativity of these caches can be configured. The default ``RocketConfig``
+uses 16 KiB, 4-way set-associative instruction and data caches. However,
+if you use the ``NMediumCores`` or ``NSmallCores`` configurations, you can
```
alonamid: Do we want to mention the specific parameters you can change to change to L1 cache sizes to something custom? (i.e., not the NMediumCores, but rather DCacheParams)

davidbiancolin:
I'm a little bit worried about the amount of code-in-docs in this PR. It's going to go obsolete very easy unless we stay on top of it.

docs/Customization/Memory-Hierarchy.rst
```diff
@@ -1,4 +1,104 @@
...
+
+If you do not want to use the L2 cache (say, for a resource-limited embedded
+design), you can create a configuration without it. Instead of using the L2
+cache, you will instead use RocketChip's TileLink broadcast hub.
+
+.. code-block:: scala
+
+    import freechips.rocketchip.subsystem.{WithNBigCores, BaseConfig}
+
+    class CachelessRocketConfig extends Config(
```
alonamid: This is meant to demonstrate that the example *lacks* an additional mix-in for a cache?

docs/Customization/Memory-Hierarchy.rst
```diff
@@ -1,4 +1,104 @@
...
+
+.. code-block:: scala
+
+    import freechips.rocketchip.subsystem.WithBufferlessBroadcastHub
+
+    class BufferlessRocketConfig extends Config(
+        new WithBufferlessBroadcastHub ++
+        new CachelessRocketConfig)
+
+The Outer Memory System
```
alonamid: Should we talk about how the outer memory is modeled in the different simulators (verisim/firesim)? Or at least link a pointer that talk about that?

docs/Generators/RocketChip.rst
```diff
@@ -0,0 +1,66 @@
+RocketChip
+==========
+
+RocketChip is an SoC generator supported by SiFive. Chipyard uses RocketChip
```
alonamid: developed in Berkeley and currently maintained by SiFive?

docs/Generators/RocketChip.rst
```diff
@@ -0,0 +1,66 @@
...
+as the basis for producing a RISC-V SoC including Rocket, BOOM, and/or Hwacha.
+
+A detailed diagram of a typical RocketChip system is shown below.
+
+.. image:: ../images/rocketchip-diagram.png
+
+Tiles
+-----
+
+This is a dual-core ``Rocket`` system. Each ``Rocket`` core is grouped with a
```
alonamid: By "This", you mean the example diagram?

docs/Generators/RocketChip.rst
```diff
@@ -0,0 +1,66 @@
+RocketChip
+==========
+
+RocketChip is an SoC generator supported by SiFive. Chipyard uses RocketChip
+as the basis for producing a RISC-V SoC including Rocket, BOOM, and/or Hwacha.
```
alonamid: Maybe it's worth mentioning here explicitly the difference between the RocketChip generator and the Rocket core?

docs/TileLink-Diplomacy-Reference/index.rst
```diff
@@ -0,0 +1,29 @@
+TileLink and Diplomacy Reference
+================================
+
+TileLink is the cache coherence and memory protocol used by RocketChip and
+other Chipyard generators. It is how different modules like caches, memories,
+peripherals, and DMA devices communicate with each other.
+
+RocketChip's TileLink implementation is built on top of Diplomacy, a framework
```
alonamid: Is it worth adding a link to Henry's CARRV paper/slides about diplomacy? (for those who want something formal)

docs/TileLink-Diplomacy-Reference/index.rst
```diff
@@ -0,0 +1,29 @@
...
+for exchanging configuration information among Chisel generators in a two-phase
+elaboration scheme.
+
+A brief overview of how to connect simple TileLink widgets can be found
+in the :ref:`Adding-an-Accelerator` section. This section will provide a
+detailed reference for the TileLink and Diplomacy functionality provided by
```
alonamid: Is this a good place to talk about the GraphML file that helps visualize the diplomacy graph? @a0u

abejgonzalez:
We should add this to the README. https://groups.google.com/forum/#!forum/chipyard

docs/Customization/Memory-Hierarchy.rst
zhemao: But the only way to get at those parameters is to override RocketTilesKey. I feel like if they want to do something tricky like that, they can just look at the code for `NMediumCores` and `NSmallCores`.

docs/Customization/Memory-Hierarchy.rst
zhemao: Yes, that's correct.

docs/Customization/Memory-Hierarchy.rst
alonamid: Then maybe it's worth adding a sentence saying "for custom cache sizes, associativity, etc., you will need to modify the RocketTilesKey parameters, as is specified in the the Parameter System <link> section (and then we should add something about this in the parameter system section).

docs/Customization/Memory-Hierarchy.rst
alonamid: Then maybe it's worth clarifying that in the text (because usually when you give these types of example, I'm looking for the specific relevant mix-in, whereas here, the point is that there is no mix-in)

jerryz123: I think its also confusing that you've redefined `CachelessRocketConfig` here, while in the actual code its defined as `RocketConfig`.

zhemao: @jerryz123 But it's not a redefinition. In the actual code, `RocketConfig` include the L2 cache, which is expressly what this is **not** doing.

docs/Customization/Memory-Hierarchy.rst
zhemao: But that section talks about the config system more generally and not about specific keys. Though we could add something about how to use `site` and `up` to reference other keys.

alonamid: We could and we should

zhemao:
@alonamid I already added the configurations for the L1 cache. It turns out you don't have to do the RocketTilesKey override thing, because there are already configs for changing number of sets and ways.

I've moved the NodeTypes examples into the `example` project as well. I'll merge this once CI completes.

PR #236, opened 2019-09-10
Make firrtl-interpreter a submodule instead of depending on external snapshot

zhemao:
Recently, the sonatype URL we were using to pull in firrtl-interpreter stopped working. This commit adds firrtl-interpreter as a submodule and switches to including it as a project in the build.sbt instead.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -142,7 +142,7 @@ lazy val sha3 = (project in file("generators/sha3"))
   .settings(commonSettings)
 
 lazy val tapeout = conditionalDependsOn(project in file("./tools/barstools/tapeout/"))
-  .dependsOn(firrtl_interpreter)
+  .dependsOn(firrtl_interpreter, `chisel-testers`)
   .settings(commonSettings)
 
 lazy val mdf = (project in file("./tools/barstools/mdf/scalalib/"))
--- a/build.sbt
+++ b/build.sbt
@@ -68,15 +68,16 @@ def isolateAllTests(tests: Seq[TestDefinition]) = tests map { test =>
 
 // Subproject definitions begin
 //
+// NB: FIRRTL should not be a managed dependency of chisel or rocketchip.
+// Instead, they will get firrtl from the JAR file placed in /lib
+lazy val chisel  = (project in rocketChipDir / "chisel3")
+
 lazy val firrtl = (project in file("tools/firrtl"))
   .settings(commonSettings)
 
 lazy val firrtl_interpreter = (project in file("tools/firrtl-interpreter"))
   .dependsOn(firrtl)
   .settings(commonSettings)
-
-// NB: FIRRTL dependency is unmanaged (and dropped in sim/lib)
-lazy val chisel  = (project in rocketChipDir / "chisel3")
 
 lazy val treadle = freshProject("treadle", file("tools/treadle"))
   .dependsOn(firrtl)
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -67,15 +67,23 @@ def isolateAllTests(tests: Seq[TestDefinition]) = tests map { test =>
   } toSeq
 
 // Subproject definitions begin
+//
```
abejgonzalez: We need to keep FIRRTL unmanaged. IIRC it is because RC also has it unmanaged.

zhemao: chisel-testers seems to break if we don't add firrtl as a dependency for it, though

zhemao: Oh nvm, I was reading that wrong. Seems like the problem is just that I didn't give the right dependencies to sha3.

zhemao: Nope, I was right the first time. RocketChip doesn't have a dependency on firrtl in its build.sbt, but firrtl-interpreter, chisel-testers, and treadle do, so we will have to manage a dependency for them it seems.

seldridge:
Just a heads up, is there a reason why this doesn't use Treadle? 

FIRRTL Interpreter has some weird quirks that result in differing behavior vs. verilator/Treadle and is pushing towards getting deprecated. E.g., https://github.com/freechipsproject/chisel3/issues/993

zhemao:
@seldridge We don't use firrtl-interpreter directly, but `chisel-testers` and some other submodules depend on it, so we have to include it here even if we aren't using it.

seldridge:
@zhemao: Got it. It's just a build issue. As the interpreter gets deprecated, it'll get pulled out of here, too. Sounds great.

build.sbt
```diff
@@ -67,15 +67,24 @@ def isolateAllTests(tests: Seq[TestDefinition]) = tests map { test =>
...
 
 // Subproject definitions begin
-
-// NB: FIRRTL dependency is unmanaged (and dropped in sim/lib)
+//
+// NB: FIRRTL should not be a managed dependency of chisel or rocketchip.
+// Instead, they will get firrtl from the JAR file placed in /lib
 lazy val chisel  = (project in rocketChipDir / "chisel3")
 
+lazy val firrtl = (project in file("tools/firrtl"))
```
davidbiancolin: this shouldn't be required.

davidbiancolin: The problem is probably that these tools are requesting a different version of firrtl than the one we are providing in lib

zhemao: But the FIRRTL jar we are providing in lib was generated from tools/firrtl, wasn't it?

grebe: What uses tools/firrtl? I would expect the firrtl in rocketchip to get used.

abejgonzalez: Everything should use the FIRRTL in `tools/firrtl`. We want to decouple RC's FIRRTL from the FIRRTL that we use so we could be ahead/behind what they provide.

build.sbt
```diff
@@ -13,8 +13,8 @@ lazy val commonSettings = Seq(
   libraryDependencies += "org.scalatest" %% "scalatest" % "3.0.5" % "test",
   libraryDependencies += "org.json4s" %% "json4s-jackson" % "3.6.1",
   libraryDependencies += "org.scala-lang" % "scala-reflect" % scalaVersion.value,
-  libraryDependencies += "edu.berkeley.cs" %% "firrtl-interpreter" % "1.2-SNAPSHOT",
   libraryDependencies += "com.github.scopt" %% "scopt" % "3.7.0",
+  libraryDependencies += "org.scala-lang.modules" % "scala-jline" % "2.12.1",
```
jwright6323: What is this?

project/plugins.sbt
```diff
@@ -13,5 +13,6 @@ addSbtPlugin("org.scalastyle" %% "scalastyle-sbt-plugin" % "1.0.0")
 addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.14.6")
 addSbtPlugin("com.simplytyped" % "sbt-antlr4" % "0.8.1")
 addSbtPlugin("com.github.gseitz" % "sbt-protobuf" % "0.6.3")
+addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.9.4")
```
abejgonzalez: Similar to John's question... why is this needed?

build.sbt
zhemao: It's a dependency that's needed by firrtl-interpreter.

https://github.com/freechipsproject/firrtl-interpreter/blob/master/build.sbt#L66

project/plugins.sbt
zhemao: FIRRTL uses this plugin

https://github.com/freechipsproject/firrtl/blob/master/project/plugins.sbt#L23

PR #239, opened 2019-09-13
Remove Berkeley-Managed Library Dependencies from Subprojects

davidbiancolin:
This PR does three things:
1) cleans up the interpreter change 
2) prevents sub-projects from fetching libraries from maven
3) uses the chisel3 in tools/ instead of the submodule in RC

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -79,7 +79,7 @@ def isolateAllTests(tests: Seq[TestDefinition]) = tests map { test =>
 // before launching sbt if any of the firrtl source files has been updated
 // The jar is dropped in chipyard's lib/ directory, which is used as the unmanagedBase
 // for all subprojects
-lazy val chisel  = (project in rocketChipDir / "chisel3")
+lazy val chisel  = (project in file("tools/chisel3"))
 
 lazy val firrtl_interpreter = (project in file("tools/firrtl-interpreter"))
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -15,7 +19,10 @@ lazy val commonSettings = Seq(
   libraryDependencies += "org.scala-lang" % "scala-reflect" % scalaVersion.value,
   libraryDependencies += "com.github.scopt" %% "scopt" % "3.7.0",
   libraryDependencies += "org.scala-lang.modules" % "scala-jline" % "2.12.1",
+  libraryDependencies += "com.typesafe.play" %% "play-json" % "2.6.10",
   addCompilerPlugin("org.scalamacros" % "paradise" % "2.1.0" cross CrossVersion.full),
+  unmanagedBase := (chipyardRoot / unmanagedBase).value,
+  allDependencies := allDependencies.value.filterNot(_.organization == "edu.berkeley.cs"),
```
grebe: Does this work? I think some things (is it barstools?) use "berkeley". This is fragile but it's hard to think of a better alternative.

davidbiancolin: The robust way would be to just create a blacklist of libraries (org %% name) and filter on those. We can punt on that though.

PR #240, opened 2019-09-16
Fix Sha3RocketConfig ordering

jerryz123:


davidbiancolin:
what was the behavior of the previous order? this could be a bug in the sha3 config no?

jerryz123:
Both orderings worked, it was just a OCD/consistency thing.

PR #247, opened 2019-09-25
Firrtl Transforms Documentation

abejgonzalez:


docs/Customization/Firrtl-Transforms.rst
```diff
@@ -0,0 +1,73 @@
+.. _firrtl-transforms:
+
+Adding a Firrtl Transform
+=========================
+
+After writing the Chisel RTL, you can make further modifications by adding transforms during the FIRRTL compilation phase.
```
jerryz123: Should make analogy to LLVM transforms for software.
"Similar to how LLVM IR passes can perform transformations and optimizations on software, FIRRTL transforms can modify Chisel-elaborated RTL."

docs/Customization/Firrtl-Transforms.rst
```diff
@@ -0,0 +1,73 @@
...
+``firechip`` project and uses an annotation to mark BOOM's register file for optimization:
+
+.. literalinclude:: ../../generators/firechip/src/main/scala/TargetMixins.scala
+    :language: make
+    :start-after: DOC include start: ChiselAnnotation
+    :end-before: DOC include end: ChiselAnnotation
+
+Here is an example of grouping a series of modules by specifying the ``.json`` file.
+
+.. code-block:: json
```
jerryz123: Can we do an example that starts with Chisel annotations? Why don't we just explain the DontTouch annotation? That is something that a lot of users will use anyways.

docs/Customization/Firrtl-Transforms.rst
```diff
@@ -0,0 +1,73 @@
...
+In Chipyard, the FIRRTL compiler is called multiple times to create a "Top" file that has the DUT and a "Harness" file that has all harness collateral.
+This done by the ``tapeout`` SBT project (located in ``tools/barstools/tapeout``) which calls ``GenerateTopAndHarness`` (a function that wraps the multiple FIRRTL compiler calls and extra transforms).
+
+.. literalinclude:: ../../common.mk
+    :language: make
+    :start-after: DOC include start: FirrtlCompiler
+    :end-before: DOC include end: FirrtlCompiler
+
+If you look inside of the `tools/barstools/tapeout/src/main/scala/transforms/Generate.scala <https://github.com/ucb-bar/barstools/blob/master/tapeout/src/main/scala/transforms/Generate.scala>`__ file,
+you can see that FIRRTL invoked twice, once for the "Top" and once for the "Harness". If you want to add transforms to just modify the DUT, you can add them to ``topTransforms``.
```
jwright6323:
```suggestion
you can see that FIRRTL is invoked twice, once for the "Top" and once for the "Harness". If you want to add transforms to just modify the DUT, you can add them to ``topTransforms``.
```

docs/Customization/Firrtl-Transforms.rst
```diff
@@ -0,0 +1,73 @@
...
+=========================
+
+After writing the Chisel RTL, you can make further modifications by adding transforms during the FIRRTL compilation phase.
+As mentioned in Section <>, transforms are modifications that happen on the FIRRTL IR that can modify a circuit.
+Transforms are a powerful tool to take in the FIRRTL IR that is emitted from Chisel and run analysis (https://www.youtube.com/watch?v=FktjrjRVBoY) or convert the circuit into a new form.
+
+Where to add transforms
+-----------------------
+
+In Chipyard, the FIRRTL compiler is called multiple times to create a "Top" file that has the DUT and a "Harness" file that has all harness collateral.
```
jwright6323: > a "Harness" file that has all harness collateral
This is kind of confusing/misleading. There's not one file that has all harness collateral- the verilog, annotations, and .fir are all separate files. If we're talking strictly about verilog, then "collateral" is a bit confusing.
```suggestion
In Chipyard, the FIRRTL compiler is called multiple times to create a "Top" file containing the DUT and a "Harness" file containing the test harness, which instantiates the DUT. The "Harness" file does not contain the DUT's module definition or any of its submodules.
```

docs/Customization/Firrtl-Transforms.rst
```diff
@@ -0,0 +1,73 @@
+.. _firrtl-transforms:
+
+Adding a Firrtl Transform
+=========================
+
+After writing the Chisel RTL, you can make further modifications by adding transforms during the FIRRTL compilation phase.
+As mentioned in Section <>, transforms are modifications that happen on the FIRRTL IR that can modify a circuit.
+Transforms are a powerful tool to take in the FIRRTL IR that is emitted from Chisel and run analysis (https://www.youtube.com/watch?v=FktjrjRVBoY) or convert the circuit into a new form.
```
colinschmidt: I would either remove this link or add one for the other use case.

docs/Customization/Firrtl-Transforms.rst
```diff
@@ -0,0 +1,73 @@
...
+               "BeagleChipTop.BeagleChipTop.instance2",
+           ],
+           "newModule": "NewModule",
+           "newInstance": "newModInst"
+       }
+   ]
+
+In this case, the specific syntax depends on the type of annotation. The best way to figure out
+what the ``json`` file should contain is to first try to annotate in the Chisel
+source and then see what the ``json`` output gives and copy that format.
```
jwright6323: I don't know that this is true. It's pretty easy to infer what a case class is going to serialize to. It's always 
```
{"class": "package.AnnotationCaseClassName", "caseClassField": "value (string, int, list, etc.)", ...}
```

docs/Customization/Firrtl-Transforms.rst
```diff
@@ -0,0 +1,73 @@
...
+----------------------
+
+There are multiple examples of transforms that you can apply and are spread across the FIRRTL ecosystem.
+Within FIRRTL there is a default set of supported transforms located in https://github.com/freechipsproject/firrtl/tree/master/src/main/scala/firrtl/transforms.
+This includes transforms that can flatten modules (``Flatten``), group modules together (``GroupAndDedup``), and more.
+
+Transforms can be standalone or can take annotations as input. Annotations are used to pass information between FIRRTL transforms. This includes information on
+what modules to flatten, group, and more. Annotations can be added to the code by
+adding them to your Chisel source or by creating the ``.json`` file and adding it to the FIRRTL compiler (note: adding to the Chisel source will add the ``json`` snippet into the build system for you).
+**The recommended way to annotate something is to do it in the Chisel source**.
```
jwright6323:
```suggestion
**The recommended way to annotate something is to do it in the Chisel source, but not all annotation types have Chisel APIs**.
```

(something like this)

docs/Customization/Firrtl-Transforms.rst
```diff
@@ -0,0 +1,73 @@
...
+Examples of transforms
+----------------------
+
+There are multiple examples of transforms that you can apply and are spread across the FIRRTL ecosystem.
+Within FIRRTL there is a default set of supported transforms located in https://github.com/freechipsproject/firrtl/tree/master/src/main/scala/firrtl/transforms.
+This includes transforms that can flatten modules (``Flatten``), group modules together (``GroupAndDedup``), and more.
+
+Transforms can be standalone or can take annotations as input. Annotations are used to pass information between FIRRTL transforms. This includes information on
+what modules to flatten, group, and more. Annotations can be added to the code by
+adding them to your Chisel source or by creating the ``.json`` file and adding it to the FIRRTL compiler (note: adding to the Chisel source will add the ``json`` snippet into the build system for you).
```
colinschmidt: `serialized annotation file as a json file`

docs/Customization/Firrtl-Transforms.rst
colinschmidt: I think it would be good if this was an actual serialized annotation that came from an example config.

docs/Customization/Firrtl-Transforms.rst
```diff
@@ -0,0 +1,73 @@
...
+           "components": [
+               "BeagleChipTop.BeagleChipTop.instance1",
+               "BeagleChipTop.BeagleChipTop.instance2",
+           ],
+           "newModule": "NewModule",
+           "newInstance": "newModInst"
+       }
+   ]
+
+In this case, the specific syntax depends on the type of annotation. The best way to figure out
```
colinschmidt: I don't think this is the best way. The transforms/annotations should have documentation right?

docs/Customization/Firrtl-Transforms.rst
```diff
@@ -0,0 +1,73 @@
...
+what the ``json`` file should contain is to first try to annotate in the Chisel
+source and then see what the ``json`` output gives and copy that format.
+
+Once ``yourAnnoFile.json`` is created then you can add ``-faf yourAnnoFile.json`` to the FIRRTL compiler invocation in ``common.mk``.
```
colinschmidt: Do we want to support this more cleanly in the build system?

docs/Customization/Firrtl-Transforms.rst
abejgonzalez: The problem with this is that it is non-intuitive when you try to figure out some of the fields:

Example:
``case class FirrtlExtractionAnnotation(target: ModuleTarget) extends
    SingleTargetAnnotation[ModuleTarget]``
means you annotation the circuit like the following
``"target":"~TopModule|SomeModule"``

How did you know to add the `~` or the `|`... I struggled with this so I figured others would?

docs/Customization/Firrtl-Transforms.rst
abejgonzalez: Look at my comment that I gave to John. Its not always intuitive when to use `~` or `>` or `|` to deliminate things when doing paths, etc.

docs/Customization/Firrtl-Transforms.rst
abejgonzalez: To me, it seems like you all want the main form of adding annotations to be in Chisel. If thats the case, the Makefile doesn't necessarily need to support this...

The easiest way to support this is to add a `FIRRTL_COMPILER_FLAGS` variable that is empty and could be overridden.

docs/Customization/Firrtl-Transforms.rst
```diff
@@ -0,0 +1,88 @@
...
+modify Chisel-elaborated RTL.
+As mentioned in Section :ref:`firrtl`, transforms are modifications that happen on the FIRRTL IR that can modify a circuit.
+Transforms are a powerful tool to take in the FIRRTL IR that is emitted from Chisel and run analysis or convert the circuit into a new form.
+
+Where to add transforms
+-----------------------
+
+In Chipyard, the FIRRTL compiler is called multiple times to create a "Top" file that contains the DUT and a "Harness" file containing the test harness, which instantiates the DUT.
+The "Harness" file does not contain the DUT's module definition or any of its submodules.
+This done by the ``tapeout`` SBT project (located in ``tools/barstools/tapeout``) which calls ``GenerateTopAndHarness`` (a function that wraps the multiple FIRRTL compiler calls and extra transforms).
```
alonamid: This *is* done?

docs/Customization/Firrtl-Transforms.rst
```diff
@@ -0,0 +1,88 @@
...
+
+.. literalinclude:: ../../common.mk
+    :language: make
+    :start-after: DOC include start: FirrtlCompiler
+    :end-before: DOC include end: FirrtlCompiler
+
+If you look inside of the `tools/barstools/tapeout/src/main/scala/transforms/Generate.scala <https://github.com/ucb-bar/barstools/blob/master/tapeout/src/main/scala/transforms/Generate.scala>`__ file,
```
alonamid: Maybe add a reference to the barstools section? (which will hopefully explain what barstools is, once it is written)

docs/Customization/Firrtl-Transforms.rst
```diff
@@ -0,0 +1,88 @@
...
+    class TopModule extends Module {
+        ...
+        val submod = Module(new Submodule)
+        ...
+    }
+
+    class Submodule extends Module {
+        ...
+        val some_signal := ...
+        annotate(new ChiselAnnotation {
```
alonamid: Is there no Chisel API for DontTouch? This seems like a very weird way to annotate...

seldridge: https://www.chisel-lang.org/api/SNAPSHOT/chisel3/dontTouch$.html

abejgonzalez: I did have the ChiselAPI right underneath. I just figured that it would be clearer to show how the ChiselAnnotation is written... then show the nice Chisel wrapper for it.

alonamid: Then I think it might be worth to explicitly state in the text the most FIRRTL annotations have an equivalent Chisel annotation. However, if there doesn't exist an equivalent Chisel annotation, the way to use a FIRRTL annotation within chisel would be as follow

alonamid: You're making the Chisel wrapper seems very secondary here, while it should be the main way of using this annotation

docs/Customization/Firrtl-Transforms.rst
alonamid: This can be addressed in a later PR (after the tutorial). You're invited to open an issue

PR #250, opened 2019-09-25
Notes that building with macOS doesn't really work

mehnadnerd:


docs/Chipyard-Basics/Initial-Repo-Setup.rst
```diff
@@ -1,6 +1,21 @@
...
 ========================================================
 
+Requirements
+-------------------------------------------
+
+Using Linux is recommended.
+The provided scripts will not work on macOS out of the box, but they may be able to be modified to support it.
+Working under Windows is not recommended.
+
+* dtc (design-tree-compiler)
```
abejgonzalez: I would get rid of the particular requirements (dtc, ...). It should be self-explanatory when it breaks that you need X thing.

docs/Chipyard-Basics/Initial-Repo-Setup.rst
```diff
@@ -1,6 +1,21 @@
 Initial Repository Setup
 ========================================================
 
+Requirements
+-------------------------------------------
+
+Using Linux is recommended.
```
abejgonzalez: I would reword to... "Chipyard developed and tested on Linux based systems." That emphasizes that we haven't thought of macOS support and don't really care about it.

PR #255, opened 2019-09-25
add firrtl interpreter to the docs [ci skip]

abejgonzalez:


seldridge:
I would prefer if the FIRRTL interpreter was de-emphasized here as Treadle is its replacement. Including both (while reasonable) is confusing without context. I.e., why are there two tools that do the same thing?

abejgonzalez:
Hopefully that is better.

seldridge:
Thanks for addressing my feedback @abejgonzalez!

PR #261, opened 2019-09-26
Overhaul toolchain organization

a0u:
This eliminates one level of submodules to make it easier to maintain toolchain components individually.

### riscv-tools

* The flattening of riscv-tools and esp-tools preserves the same directory structure.
* All components have been bumped to the latest master revisions as of this week.
  * I specifically aimed for a more recent binutils that supports the `.insn` pseudo-op, which provides a cleaner way to implement the RoCC macros than `.word`.
* riscv-openocd is treated as optional since no design flow requires it (yet).  Run `scripts/build-openocd.sh` to install it.
* riscv-fesvr is now part of riscv-isa-sim, which means that `libfesvr.so` no longer exists.  VCS compilation has been adjusted accordingly.

### esp-tools

* The esp-tools versions generally track their mainline counterparts (GCC 9.2, Binutils 2.32).
* The exception is `esp-isa-sim`, which is held back at the "v1.0.0" release to avoid intermingling with the RVV extension support.  Xhwacha and RVV can coexist (Hwacha's `vsetvl` just needs to be renamed), but it does not seem worthwhile presently.

### Binary distribution

* As discussed, only riscv-gnu-toolchain is being pre-built to simplify versioning.  There is a new packaging flow similar to FireSim: http://github.com/chipyard/chipyard-toolchain-prebuilt
* The distfiles are handled as a shallow submodule in `toolchains/riscv-tools/riscv-gnu-toolchain-prebuilt` instead of an untracked clone.
* Binaries built on CentOS 7 work on the `a*` machines.

### Portability

* `scripts/build-toolchains.sh` should conform strictly to the POSIX shell.  The only feature that prevents running under pure `/bin/sh` is `set -o pipefail`, which is beneficial enough for user-friendliness to retain Bash, but this could be worked around with some effort.
* I've tested building on FreeBSD 12.  Besides the typical GNU dependencies (gawk, gsed, etc.), the main caveat is that gcc is needed to compile gcc.  Building with clang is possible but requires some inconvenient patches.  I assume that OS X is similarly affected.
https://svnweb.freebsd.org/ports/head/lang/gcc9/files/?pathrev=509718


a0u:
CI is failing because glibc now requires Python 3.4+ to build.  I don't have the credentials to update the riscv-boom Docker image.

alonamid:
Waiting for FireSim testing confirmation from @a0u and/or @NathanTP , and BOOM testing confirmation from @jerryz123 (Verilator and FireSim)

a0u:
I have built and tested Linux 5.3 buildroot images with the new toolchain.  Both "disk" and "nodisk" configurations  boot successfully on the stock `firesim-quadcore-nic-ddr3-llc4mb` AGFI.

@NathanTP: Once firesim/FireMarshal#46 lands, buildroot needs to be bumped to the `2019.08` release.  For now I have hacked the buildroot Kconfig to recognize GCC 9.2 and Linux 5.0 headers.

scripts/build-util.sh
```diff
@@ -0,0 +1,89 @@
...
+
+module_build() ( # <submodule> [configure-arg..]
+    set -e -o pipefail
+    name=$1
+    shift
+
+    cd "${SRCDIR}/${name}"
+
+    if [ -e build ] ; then
+        echo "==>  Removing existing ${name}/build directory"
```
jerryz123: Why did the old `build-tools.sh` always blow out the build directory on a rebuild?

a0u: It ensures reproducibility in the least complicated (if sometimes inconvenient) manner.  For example, autoconf could be using cached results of previous configure tests, or insufficiently specified dependencies in makefiles might not guarantee that everything is rebuilt as necessary.

NathanTP:
I tested on marshal (post linux bump) and everything builds and runs. @a0u I didn't have any issue building BR on the prebuilt toolchain. I think master is on an old buildroot but @davidbiancolin bumped buildroot a while ago when he last updated the toolchain.

a0u:
> I tested on marshal (post linux bump) and everything builds and runs. @a0u I didn't have any issue building BR on the prebuilt toolchain. I think master is on an old buildroot but @davidbiancolin bumped buildroot a while ago when he last updated the toolchain.

If buildroot already has the external toolchain set up, it won't complain when switching compilers.  The sysroot is then potentially out of sync, as glibc likely hasn't been copied over again.

However, from a pristine state, one encounters:
```
>>> toolchain-external-custom  Extracting
>>> toolchain-external-custom  Patching
>>> toolchain-external-custom  Configuring
Incorrect selection of kernel headers: expected 4.13.x, got 5.0.x
make: *** [/mnt/firesim/src/chipyard/sims/firesim/sw/firesim-software/wlutil/br/buildroot/output/build/toolchain-external-custom/.stamp_configured] Error 1
```

The current buildroot release is `2019.02.2`, which predates both Linux 5.x and GCC 9.x series.
Interestingly, unsetting certain Kconfig knobs suppresses the check even if the new ones aren't defined:

```diff
diff --git a/wlutil/br/buildroot-config b/wlutil/br/buildroot-config
index b5d88dd..6e673f9 100644
--- a/wlutil/br/buildroot-config
+++ b/wlutil/br/buildroot-config
@@ -159,6 +159,7 @@ BR2_TOOLCHAIN_EXTERNAL_PREFIX="$(ARCH)-unknown-linux-gnu"
 BR2_TOOLCHAIN_EXTERNAL_CUSTOM_PREFIX="$(ARCH)-unknown-linux-gnu"
 # BR2_TOOLCHAIN_EXTERNAL_GCC_8 is not set
 BR2_TOOLCHAIN_EXTERNAL_GCC_7=y
+BR2_TOOLCHAIN_EXTERNAL_HEADERS_5_0=y
 # BR2_TOOLCHAIN_EXTERNAL_HEADERS_4_20 is not set
 # BR2_TOOLCHAIN_EXTERNAL_HEADERS_4_19 is not set
 # BR2_TOOLCHAIN_EXTERNAL_HEADERS_4_18 is not set
@@ -166,7 +167,7 @@ BR2_TOOLCHAIN_EXTERNAL_GCC_7=y
 # BR2_TOOLCHAIN_EXTERNAL_HEADERS_4_16 is not set
 # BR2_TOOLCHAIN_EXTERNAL_HEADERS_4_15 is not set
 # BR2_TOOLCHAIN_EXTERNAL_HEADERS_4_14 is not set
-BR2_TOOLCHAIN_EXTERNAL_HEADERS_4_13=y
+# BR2_TOOLCHAIN_EXTERNAL_HEADERS_4_13 is not set
 # BR2_TOOLCHAIN_EXTERNAL_HEADERS_4_12 is not set
 # BR2_TOOLCHAIN_EXTERNAL_HEADERS_4_11 is not set
 # BR2_TOOLCHAIN_EXTERNAL_HEADERS_4_10 is not set
@@ -267,7 +268,8 @@ BR2_TOOLCHAIN_HEADERS_AT_LEAST_4_10=y
 BR2_TOOLCHAIN_HEADERS_AT_LEAST_4_11=y
 BR2_TOOLCHAIN_HEADERS_AT_LEAST_4_12=y
 BR2_TOOLCHAIN_HEADERS_AT_LEAST_4_13=y
-BR2_TOOLCHAIN_HEADERS_AT_LEAST="4.13"
+BR2_TOOLCHAIN_HEADERS_AT_LEAST_5_0=y
+BR2_TOOLCHAIN_HEADERS_AT_LEAST="5.0"
 BR2_TOOLCHAIN_GCC_AT_LEAST_4_3=y
 BR2_TOOLCHAIN_GCC_AT_LEAST_4_4=y
 BR2_TOOLCHAIN_GCC_AT_LEAST_4_5=y
@@ -278,7 +280,8 @@ BR2_TOOLCHAIN_GCC_AT_LEAST_4_9=y
 BR2_TOOLCHAIN_GCC_AT_LEAST_5=y
 BR2_TOOLCHAIN_GCC_AT_LEAST_6=y
 BR2_TOOLCHAIN_GCC_AT_LEAST_7=y
-BR2_TOOLCHAIN_GCC_AT_LEAST="7"
+BR2_TOOLCHAIN_GCC_AT_LEAST_9=y
+BR2_TOOLCHAIN_GCC_AT_LEAST="9"
 BR2_TOOLCHAIN_HAS_MNAN_OPTION=y
 BR2_TOOLCHAIN_HAS_SYNC_1=y
 BR2_TOOLCHAIN_HAS_SYNC_2=y
```

NathanTP:
Alright, I see the error now on a fresh marshal clone. I assume this would also fail on current Marshal dev branch due to the "BR2_TOOLCHAIN_GCC_AT_LEAST_9=y" option. If that's the case, I'd recommend holding off on merging this until I bump buildroot (I can do that as soon as firesim/FireMarshal#46 goes through).

a0u:
I don't think it's worth delaying this.   To bump marshal, we need to also bump firesim, which is even more complicated.  Bumping buildroot and updating the config also means that users will see the version mismatch error when using the old toolchain.

Everything has to be synced before the MICRO release, but until then I think we can accept some decoupling to not lengthen the critical path.

jerryz123:
Works with BOOM w. MegaBoom+L2+4mbllc.

NathanTP:
I just tested this on the dev branch of marshal and it doesn't work there either. What good is merging this if it doesn't work?

a0u:
As discussed, the `firesim` submodule now points to FireSim dev with the Buildroot update merged.

PR #263, opened 2019-10-03
Bugfixes for ASAP7 demo

harrisonliew:
Turns out ASAP7 can't be used with Innovus 19.1, so reverted down to 18.1 instead. Fixed the GDS hacking scripts.


vlsi/example-vlsi
```diff
@@ -24,21 +24,25 @@ def example_tool_settings(x: hammer_vlsi.HammerTool) -> bool:
...
     x.append('''
 # Write script out to a temporary file and execute it
 set fp [open "{script_file}" "w"]
 puts -nonewline $fp "{script_text}"
 close $fp
-if {{ [catch {{ exec python3 {script_file} }} msg] }} {{
-    puts "$::errorInfo"
-}}
+
+# Innovus <19.1 appends some bad LD_LIBRARY_PATHS, so remove them before executing python
```
colinschmidt: Should this also include an example of the checking versions API?

vlsi/example-vlsi
```diff
@@ -50,16 +54,21 @@ class ExampleDriver(CLIDriver):
...
-            hammer_vlsi.HammerTool.make_pre_insertion_hook("route_design", example_add_fillers),  # SYNTAX: make_pre_insertion_hook("EXISTING_STEP", INSERTED_HOOK)
+            # SYNTAX: make_pre_insertion_hook("EXISTING_STEP", INSERTED_HOOK)
+            # hammer_vlsi.HammerTool.make_pre_insertion_hook("route_design", example_add_fillers),
+
             # make_post_insertion_hook will execute the custom hook after the specified step
             hammer_vlsi.HammerTool.make_post_insertion_hook("init_design", example_tool_settings),
+
             # make_replacement_hook will replace the specified step with a custom hook
-            hammer_vlsi.HammerTool.make_replacement_hook("place_tap_cells", example_place_tap_cells),
+            # hammer_vlsi.HammerTool.make_replacement_hook("place_tap_cells", example_place_tap_cells),
```
colinschmidt: Its unclear which of these hooks is just here to show syntax and which are actually needed.

vlsi/example-vlsi
harrisonliew: Yeah, I think I could check for version 181 and below or error out. This particular comment doesn't need that because P&R should not be attempted with version 191 anyway.

vlsi/example-vlsi
harrisonliew: Ok, I'll move the unused ones that are showing syntax to a different section in the list.

PR #265, opened 2019-10-03
Bump SHA-3 accelerator for tutorial enhancements

a0u:
* ucb-bar/sha3#14: Conform to the finalized FIPS 202 standard.
* ucb-bar/sha3#15: Fix virtual memory interaction with interrupts.

This was tested on FireSim.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -77,9 +77,9 @@
 [submodule "tools/treadle"]
 	path = tools/treadle
 	url = https://github.com/freechipsproject/treadle.git
-[submodule "generators/rocc-template"]
+[submodule "generators/sha3"]
 	path = generators/sha3
-	url = https://github.com/ucb-bar/rocc-template.git
+	url = https://github.com/ucb-bar/sha3.git
 [submodule "tools/firrtl-interpreter"]
 	path = tools/firrtl-interpreter
 	url = https://github.com/freechipsproject/firrtl-interpreter.git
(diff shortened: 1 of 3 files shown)


generators/firechip/src/main/scala/TargetConfigs.scala
```diff
@@ -138,6 +138,11 @@ class FireSimRocketChipOctaCoreConfig extends Config(
   new WithNDuplicatedRocketCores(8) ++
   new FireSimRocketChipSingleCoreConfig)
 
+// SHA3 accelerator
+class FireSimRocketChipSha3Config extends Config(
```
alonamid: what about the L2?

.gitmodules
```diff
@@ -79,7 +79,7 @@
 	url = https://github.com/freechipsproject/treadle.git
 [submodule "generators/rocc-template"]
```
abejgonzalez:
```suggestion
[submodule "generators/sha3"]
```

a0u:
I made this PR in the meeting before the L2 was discussed.  Time for a rebase.

a0u:
I'm a little ambivalent about having a random `FireSimRocketChipSha3L2Config` in mainline Chipyard.  Maybe this should go into a tutorial-only branch instead of `dev`, but it could be useful to keep around to reference in other educational materials.

Maybe @sagark or @davidbiancolin have an opinion.

alonamid:
Are we also adding the rocket-boom firechip config in this PR?

a0u:
I assumed that would be a separate PR once @abejgonzalez confirms that the heterogeneous config still works.

PR #267, opened 2019-10-04
Add marshal symlink

NathanTP:
Adds FireMarshal as a symlink at chipyard/software/firemarshal

This depends on:
firesim/FireMarshal#52
firesim/firesim#351

I can't be sure I got all the submodules right with the circular dependency.

Testing:
* I built and installed br-base.json from the symlink and everything seems to be in order.
* I'm now testing on EC2, just to be sure it all works end-to-end.


NathanTP:
One potential issue here:
Marshal relies on a fairly recent version of Qemu. FireSim installs this during build-setup.sh, but chipyard doesn't do that by default. We might consider moving Qemu into the chipyard toolchain installation procedure. I'm not sure what the best move is here.

sagark:
Although @a0u should look at the bash stuff.

scripts/init-submodules-no-riscv-tools.sh
```diff
@@ -6,6 +6,18 @@ set -o pipefail
 
 RDIR=$(git rev-parse --show-toplevel)
 
+NO_FIRESIM=false
+
+while test $# -gt 0
```
a0u: I prefer to use the standard `getopts` built-in for options parsing.  It's more flexible if we have more flags in the future, since it accounts for merging (`-ab` is equivalent to `-a -b`) and argument separation (`-f<arg>` is equivalent to `-f <arg>`).

It doesn't natively support GNU-style long options, but this can be worked around:
```sh
_usage() {
  echo "usage ${0} [--no-firesim]" >&2
  exit 1
}

while getopts '-:' opt ; do
  case ${opt} in
  -)
    case ${OPTARG} in
    no-firesim) NO_FIRESIM=true ;;
    *) echo "invalid option: --${OPTARG}" >&2 ; _usage ;;
    esac ;;
  *) echo "invalid option: -${opt}" >&2 ; _usage ;;
  esac
done
shift $((OPTIND - 1))
```

NathanTP: switched

.gitmodules
```diff
@@ -92,3 +92,6 @@
 [submodule "vlsi/hammer-mentor-plugins"]
 	path = vlsi/hammer-mentor-plugins
 	url = git@github.com:ucb-bar/hammer-mentor-plugins.git
+[submodule "toolchains/qemu"]
```
abejgonzalez: Probably want to write a doc on this... but we can just make this an issue and do this later.

NathanTP: What's there to document? It's installed by the build-toolchains script and is needed the same as other toolchain stuff (e.g. spike). I suppose mentioning that we override any system qemu they have already installed wouldn't hurt.

PR #268, opened 2019-10-04
Docs Cleanup

alonamid:
Add logos, and fix stuff I've been finding along the way


docs/conf.py
```diff
@@ -121,6 +121,7 @@
     ]
 }
 
+html_logo = ' _static/images/chipyard-logo.png'
```
abejgonzalez:
```suggestion
html_logo = '_static/images/chipyard-logo.png'
```

PR #271, opened 2019-10-05
Bump sha3 for RTL fixes/optimizations

a0u:
This fixes a TLB-related issue introduced in #265.

I am still waiting for the AGFI build.


a0u:
Bumped to include fixes for two more RTL bugs discovered along the way.

a0u:
Tested on F1 with the minimal baremetal/Linux benchmarks and John the Ripper.

PR #275, opened 2019-10-07
[system] Elaborate tiles in order of hartid

jerryz123:
When the harts are defined out-of-order, the DTS entry for the CLINT will also be misordered. It was difficult to track down where that was occuring. For now we can fix this by defining the tiles by order of hartid.

I believe the new code is also DRYer than the old implementation.


generators/utilities/src/main/scala/Subsystem.scala
```diff
@@ -34,49 +34,47 @@ trait HasBoomAndRocketTiles extends HasTiles
...
-
-    connectMasterPortsToSBus(rocket, crossing)
-    connectSlavePortsToCBus(rocket, crossing)
-
-    def treeNode: RocketTileLogicalTreeNode = new RocketTileLogicalTreeNode(rocket.rocketLogicalTree.getOMInterruptTargets)
-    LogicalModuleTree.add(logicalTreeNode, rocket.rocketLogicalTree)
-
-    rocket
-  }
+  // This MUST be performed in order of hartid
```
jwright6323: Probably worth explaining why

generators/utilities/src/main/scala/Subsystem.scala
```diff
@@ -34,49 +34,47 @@ trait HasBoomAndRocketTiles extends HasTiles
...
+  val tiles = allTilesInfo.sortWith(_._1.hartId < _._1.hartId).map {
+    case (param, crossing) => {
+      val (tile, rocketLogicalTree) = param match {
+        case r: RocketTileParams => {
+          val t = LazyModule(new RocketTile(r, crossing, PriorityMuxHartIdFromSeq(rocketTileParams), logicalTreeNode))
+          (t, t.rocketLogicalTree)
+        }
+        case b: BoomTileParams => {
+          val t = LazyModule(new BoomTile(b, crossing, PriorityMuxHartIdFromSeq(boomTileParams), logicalTreeNode))
+          (t, t.rocketLogicalTree) // TODO FIX rocketLogicalTree is not a member of the superclass, both child classes define it separately
```
abejgonzalez: Is there any way to reasonably fix this without a change to RC?

abejgonzalez:
LGTM. I still want to test this with a Linux boot. Additionally, we should wait until CI passes on it.

generators/utilities/src/main/scala/Subsystem.scala
jerryz123: Need to fix the LogicalTree representation for BOOM to not just use the Rocket one.

abejgonzalez:
I am fixing the error in firesim right now. I will push a fix once Linux works.

abejgonzalez:
Testing this as we speak on FireSim with a SmallBoomConfig and SmallBoom+RocketConfig

generators/utilities/src/main/scala/Subsystem.scala
jerryz123:
```suggestion
  // This MUST be performed in order of hartid
  // There is something weird with registering tile-local interrupt controllers to the CLINT.
  // TODO: investigate why
```

generators/firechip/src/main/scala/TargetMixins.scala
```diff
@@ -9,12 +9,15 @@ import freechips.rocketchip.tilelink._
 import freechips.rocketchip.subsystem._
 import freechips.rocketchip.amba.axi4._
 import freechips.rocketchip.util._
+import freechips.rocketchip.tile._
```
jerryz123:
```suggestion
import freechips.rocketchip.tile.RocketTile
```

jerryz123:
Superseded by #277

PR #279, opened 2019-10-10
Update SHA3 to have parameterized printfs

abejgonzalez:


abejgonzalez:
Note: Berkeley servers are down due to the power outage so CI will fail.

PR #284, opened 2019-10-10
Check and add Hwacha tests

abejgonzalez:
Basic example of checking for a Hwacha parameter to add its tests.

### DIFF
--- a/generators/example/src/main/scala/Generator.scala
+++ b/generators/example/src/main/scala/Generator.scala
@@ -1,10 +1,15 @@
 package example
 
+import scala.util.control.Exception
+
 import chisel3._
 
 import freechips.rocketchip.config.{Parameters}
 import freechips.rocketchip.util.{GeneratorApp}
-import utilities.TestSuiteHelper
+import freechips.rocketchip.tile.{BuildRoCC}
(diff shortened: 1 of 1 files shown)


abejgonzalez:
@a0u previously used a cleaner way to query the Hwacha parameters. I am using that instead.

https://github.com/firesim/firesim/blob/bf453c19e99993934e12294b88f02c808a03469a/sim/src/main/scala/firesim/Generator.scala#L203-L209

generators/example/src/main/scala/Generator.scala
```diff
@@ -1,17 +1,34 @@
 package example
 
+import scala.util.Try
+
 import chisel3._
 
 import freechips.rocketchip.config.{Parameters}
 import freechips.rocketchip.util.{GeneratorApp}
-import utilities.TestSuiteHelper
+import freechips.rocketchip.tile.{BuildRoCC}
```
abejgonzalez:
```suggestion
```

abejgonzalez:
I agree this mechanism is just to make things work. We need to think more about how to inject tests based on tiles, rocc modules, etc.

PR #287, opened 2019-10-11
[DO NOT MERGE] Final tutorial branch

sagark:
This is just so I can track changes. This contains a bunch of hacks that should never end up on dev.

Notes:

Done:
- [x] FireMarshal timebase freq hack @a0u
- [x] FireSim remove dram zeroing
- [x] build_recipes.ini with only one hwdb entry present
- [ ] existing hwdb entry in built-hwdb-entries/

TODO:
* script to run before lunch that does stuff like:
    * Do replace_rtl on the AMI for the design they are building
    * Marshal builds
* need to update the built-hwdb-entires included entry after lowering freq and regen


albert-magyar:
Should we close this?

PR #292, opened 2019-10-11
Use new website links

seldridge:
Use the new website links. `chisel.eecs.berkeley.edu` will properly redirect. However, `freechipsproject.github.io/firrtl` is legacy at this point.


davidbiancolin:
We'll merge this a little later tonight.

PR #297, opened 2019-10-12
Add Verilog blackbox FireSim simulation example

albert-magyar:
This provides an example design with a GCD MMIO peripheral with a Verilog blackbox in FireChip. It can be pushed through FPGA simulation with firesim/firesim#388 and ucb-bar/midas#150.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -189,7 +189,7 @@ lazy val midas      = ProjectRef(firesimDir, "midas")
 lazy val firesimLib = ProjectRef(firesimDir, "firesimLib")
 
 lazy val firechip = (project in file("generators/firechip"))
-  .dependsOn(boom, icenet, testchipip, sifive_blocks, sifive_cache, sha3, utilities, tracegen, midasTargetUtils, midas, firesimLib % "test->test;compile->compile")
+  .dependsOn(boom, icenet, testchipip, example, sifive_blocks, sifive_cache, sha3, utilities, tracegen, midasTargetUtils, midas, firesimLib % "test->test;compile->compile")
   .settings(
     commonSettings,
     testGrouping in Test := isolateAllTests( (definedTests in Test).value )
--- a/build.sbt
+++ b/build.sbt
@@ -189,7 +189,7 @@ lazy val midas      = ProjectRef(firesimDir, "midas")
 lazy val firesimLib = ProjectRef(firesimDir, "firesimLib")
 
 lazy val firechip = (project in file("generators/firechip"))
-  .dependsOn(boom, icenet, testchipip, example, sifive_blocks, sifive_cache, sha3, utilities, tracegen, midasTargetUtils, midas, firesimLib % "test->test;compile->compile")
+  .dependsOn(example, icenet, testchipip, tracegen, midasTargetUtils, midas, firesimLib % "test->test;compile->compile")
   .settings(
     commonSettings,
     testGrouping in Test := isolateAllTests( (definedTests in Test).value )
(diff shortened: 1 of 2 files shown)


generators/firechip/src/main/scala/Targets.scala
```diff
@@ -101,3 +101,15 @@ class FireSimTraceGenModuleImp(outer: FireSimTraceGen) extends BaseSubsystemModu
 
 // Supernoded-ness comes from setting p(NumNodes) (see DefaultFiresimHarness) to something > 1
 class FireSimSupernode(implicit p: Parameters) extends DefaultFireSimHarness(() => new FireSimDUT)
+
+// Verilog blackbox integration demo
```
davidbiancolin: In the future it would be good to have a more substantial demo for this. Maybe we can have a ugrad black-box a RoCC accelerator.

build.sbt
```diff
@@ -189,7 +189,7 @@ lazy val midas      = ProjectRef(firesimDir, "midas")
 lazy val firesimLib = ProjectRef(firesimDir, "firesimLib")
 
 lazy val firechip = (project in file("generators/firechip"))
-  .dependsOn(boom, icenet, testchipip, sifive_blocks, sifive_cache, sha3, utilities, tracegen, midasTargetUtils, midas, firesimLib % "test->test;compile->compile")
+  .dependsOn(boom, icenet, testchipip, example, sifive_blocks, sifive_cache, sha3, utilities, tracegen, midasTargetUtils, midas, firesimLib % "test->test;compile->compile")
```
abejgonzalez: Im fine with putting `example` here... however this means `firechip` will compile `hwacha`. Is that fine for FireSim folks? If we do add `hwacha` builds... it would only build once every so often so its not that bad.

abejgonzalez: Also, we could remove the `boom`, `sifive_blocks`, `sifive_cache`, `utilities`, `sha3` to make this cleaner

generators/firechip/src/main/scala/Targets.scala
abejgonzalez: +1

build.sbt
davidbiancolin: Adding the dependency example is fine; incremental recompilation works pretty well.

davidbiancolin: Adding the dependency on example is fine; incremental recompilation works pretty well.

generators/firechip/src/main/scala/Targets.scala
albert-magyar: This is why we need to keep teaching undergrads Verilog in 151 :)

build.sbt
abejgonzalez:
```suggestion
  .dependsOn(example, icenet, testchipip, tracegen, midasTargetUtils, midas, firesimLib % "test->test;compile->compile")
```

build.sbt
```diff
@@ -189,7 +189,7 @@ lazy val midas      = ProjectRef(firesimDir, "midas")
 lazy val firesimLib = ProjectRef(firesimDir, "firesimLib")
 
 lazy val firechip = (project in file("generators/firechip"))
```
abejgonzalez: Also while we are at it, could we change the `icenet` project and the `firechip` project to use `conditionalDependsOn` (and delete the `testchipip` dependency). This either pulls `testchipip` from maven or it just adds the dependency from the repo.

davidbiancolin: Why are we even pulling that conditionally from maven? These changes seem out of scope to the PR.

build.sbt
albert-magyar: There were already other redundant, transitive dependencies in the `build.sbt`. I leave finding them as an exercise to the reader.

build.sbt
abejgonzalez: Paul Rigge wants to support a case where you might use `testchipip` from maven. We can do this in a separate PR. Im going to go ahead and approve this PR.

albert-magyar:
@abejgonzalez @davidbiancolin I reverted the FireSim pointer to what's in `dev`, since this PR doesn't actually depend on any FireSim changes, as long as nobody is *using* the new target. For now, it will just add the new FireChip target.

PR #301, opened 2019-10-13
Fix tracegen target

zhemao:
This fixes the FireSimTraceGen target to conform to the new way of doing things.


zhemao:
Unfortunately, the way things are set up, the TraceGen tests can't finish cleanly, so adding it to scalatests would be a bit tricky.

zhemao:
I suppose I can fix that problem by adding a bridge for it.

davidbiancolin:
> Unfortunately, the way things are set up, the TraceGen tests can't finish cleanly, so adding it to scalatests would be a bit tricky.

Can you elaborate?

zhemao:
@davidbiancolin TraceGen doesn't use the HTIF serial. It signals completion by setting a "success" signal. I'm made a PR to add a bridge to firesim to poll this signal so that the simulation can exit cleanly.

https://github.com/firesim/firesim/pull/394

generators/firechip/src/main/scala/BridgeBinders.scala
```diff
@@ -72,5 +78,6 @@ class WithDefaultFireSimBridges extends Config(
   new WithUARTBridge ++
   new WithBlockDeviceBridge ++
   new WithFASEDBridge ++
-  new WithTracerVBridge
+  new WithTracerVBridge ++
+  new WithTraceGenBridge
```
davidbiancolin: I think it would probably be best in general to just add this to a TraceGen config but that's a nit.

.circleci/config.yml
```diff
@@ -516,6 +545,30 @@ jobs:
             - run:
                 name: Run hwacha tests
                 command: .circleci/run-tests.sh hwacha
+    tracegen-run-tests:
```
davidbiancolin: How long do these tests take to run?

zhemao: On CI it takes about 40 minutes to build and 4 minutes to run.

generators/firechip/src/main/scala/BridgeBinders.scala
zhemao: Fixed

PR #302, opened 2019-10-13
(master) Fix commit on master CI check

abejgonzalez:
Get the right name for the submodules in the `vlsi` folder


abejgonzalez:
Skipping CI on this since this doesn't run until its on `master`

PR #304, opened 2019-10-13
Add a specific toolchain `env.sh`

abejgonzalez:
This adds another `env.sh` that is specific to the toolchain that is built. This makes it so that the `env.sh` does not get clobbered.

### DIFF
--- a/scripts/build-toolchains.sh
+++ b/scripts/build-toolchains.sh
@@ -126,10 +126,14 @@ SRCDIR="$(pwd)/toolchains" module_all qemu --prefix="${RISCV}" --target-list=ris
 
 cd "$RDIR"
 
+# create specific env.sh
 {
     echo "export CHIPYARD_TOOLCHAIN_SOURCED=1"
     echo "export RISCV=$(printf '%q' "$RISCV")"
     echo "export PATH=\${RISCV}/bin:\${PATH}"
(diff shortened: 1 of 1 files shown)


alonamid:
Is this in reference to a specific bug? Why is it PR'd to `master`?

abejgonzalez:
No this is not in reference to a single bug, it is just a QoL improvement that I noticed.

scripts/build-toolchains.sh
```diff
@@ -126,10 +126,14 @@ SRCDIR="$(pwd)/toolchains" module_all qemu --prefix="${RISCV}" --target-list=ris
...
 {
     echo "export CHIPYARD_TOOLCHAIN_SOURCED=1"
     echo "export RISCV=$(printf '%q' "$RISCV")"
     echo "export PATH=\${RISCV}/bin:\${PATH}"
     echo "export LD_LIBRARY_PATH=\${RISCV}/lib\${LD_LIBRARY_PATH:+":\${LD_LIBRARY_PATH}"}"
-} > env.sh
+} > env-$TOOLCHAIN.sh
+
+# create general env.sh
+ln -s env-$TOOLCHAIN.sh env.sh
```
a0u: In general, I agree it is useful to support switching between installed toolchains.
However, `ln` without `-f` will fail if the symlink already exists.

PR #305, opened 2019-10-13
Get `master` CI fix

abejgonzalez:
See #302


abejgonzalez:
Skipping CI on this since it doesn't run unless its on `master`... However, we should wait a day to check that master CI works.

abejgonzalez:
@zhemao It looks like `icenet` is not pushed to master. Can we have `icenet` `dev` pushed to `master` and put into the `dev` and `master` branches of Chipyard?

abejgonzalez:
This is now running the CI check correctly. There are just issues with some commits not being on master.

PR #307, opened 2019-10-14
Fix `master` CI

abejgonzalez:
This does @sagark's recommendation to print out all the failures and exit only at the end of the script. Additionally, it deals with the flattened toolchain folders.

### DIFF
--- a/.circleci/check-commit.sh
+++ b/.circleci/check-commit.sh
@@ -21,29 +21,66 @@ git config submodule.vlsi/hammer-synopsys-plugins.update none
 git submodule update --init
 status=$(git submodule status)
 
+all_names=()
+
 search () {
     for submodule in "${submodules[@]}"
     do
         echo "Running check on submodule $submodule in $dir"
-        hash=$(echo "$status" | grep $submodule | awk '{print$1}' | grep -o "[[:alnum:]]*")
-        echo "Searching for $hash in origin/master of $submodule"
-        git -C $dir/$submodule branch -r --contains "$hash" | grep "origin/master" # needs init'ed submodules
+        hash=$(echo "$status" | grep "$dir.*$submodule " | awk '{print$1}' | grep -o "[[:alnum:]]*")
+        echo "Searching for $hash in origin/$branch of $submodule"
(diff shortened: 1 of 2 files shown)


abejgonzalez:
Cancelled CI on this PR since it can only be checked on `master`

davidbiancolin:
Is there a reason you're neglecting firesim?

abejgonzalez:
> Is there a reason you're neglecting firesim?

I forgot to add it. Thanks.

.circleci/check-commit.sh
```diff
@@ -21,29 +21,72 @@ git config submodule.vlsi/hammer-synopsys-plugins.update none
...
 search () {
     for submodule in "${submodules[@]}"
     do
         echo "Running check on submodule $submodule in $dir"
-        hash=$(echo "$status" | grep $submodule | awk '{print$1}' | grep -o "[[:alnum:]]*")
-        echo "Searching for $hash in origin/master of $submodule"
-        git -C $dir/$submodule branch -r --contains "$hash" | grep "origin/master" # needs init'ed submodules
+        hash=$(echo "$status" | grep "$dir.*$submodule " | awk '{print$1}' | grep -o "[[:alnum:]]*")
+        echo "Searching for $hash in origin/$branch of $submodule"
+        (git -C $dir/$submodule branch -r --contains "$hash" | grep "origin/$branch") && true # needs init'ed submodules
```
a0u: `&& true` is essentially a no-op, and the subshell is unnecessary.  If simply testing for a match, use `grep -q` to suppress output.  Also consider using `grep -F` for fixed-string patterns.

a0u: Okay, for the record: `&& true` suppresses `-o pipefail`, which seems to follow the standard `-e` behavior about when non-zero exit codes are ignored.

From POSIX:
> When this option is on, if a simple command fails for any of the reasons listed in Consequences of Shell Errors or returns an exit status value >0, and is not part of the compound list following a while, until, or if keyword, and is not a part of an **AND or OR list**, and is not a pipeline preceded by the ! reserved word, then the shell shall immediately exit.

PR #312, opened 2019-10-16
[docs][ci skip] Remove extra backticks in docs

### LINKED ISSUE #299: VLSI tutorial
https://github.com/ucb-bar/chipyard/blob/master/docs/VLSI/Tutorial.rst
1. There are extra markdown symbols in some commands.
e.g. \`\`./scripts/init-vlsi.sh asap7\`\`. 

2. In addition to `make syn`,  for the ASAP7 example without memory compiler, maybe add explicit instr like`make syn MACROCOMPILER_MODE='--mode synflops'`, so the synthesis won't fail with the error `hammer-vlsi/technology/asap7/sram-cache.json (No such file or directory)`.  It is unclear if the env var `MACROCOMPILER_MODE` needs to be explicitly passed to `make syn` again after `make buildfile`

colinschmidt:
Also fix sram-cache error. Fixes #299

### DIFF
--- a/docs/VLSI/Tutorial.rst
+++ b/docs/VLSI/Tutorial.rst
@@ -64,7 +64,7 @@ In the Chipyard root, run:
 
 .. code-block:: shell
 
-    ``./scripts/init-vlsi.sh asap7`` 
+    ./scripts/init-vlsi.sh asap7
     
 to pull the Hammer & plugin submodules. Note that for technologies other than ``asap7``, the tech submodule must be added in the ``vlsi`` folder first.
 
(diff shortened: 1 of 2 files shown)


docs/VLSI/Tutorial.rst
```diff
@@ -82,7 +82,8 @@ To elaborate the ``Sha3RocketConfig`` (Rocket Chip w/ the accelerator) and set u
 
 .. code-block:: shell
 
-    make buildfile MACROCOMPILER_MODE='--mode synflops' CONFIG=Sha3RocketConfig VLSI_TOP=Sha3AccelwBB
+    export MACROCOMPILER_MODE='--mode synflops'
```
harrisonliew: I was told that it's generally not good practice to export these variables to the shell environment, hence why I had taken this out from a previous iteration of the docs. Perhaps it's better to make it clearer that one has 3 options (in order of best practice): 1) append to make command, 2) change the Makefile, 3) export the variable?

colinschmidt: Ok. I'll fix this in a different way then.

colinschmidt:
As requested the makefile is fixed and docs updated to match.

PR #315, opened 2019-10-18
add large boom hwacha config

abejgonzalez:
Added a Hwacha config to the BOOM configurations. This is needed for BOOM CI.


abejgonzalez:
Canceled CI since this does not get tested with CI.

PR #318, opened 2019-10-25
Fix VLSI Makefile

abejgonzalez:
The `SMEMS_FILE` was supposed to be `TOP_SMEMS_FILE`


vlsi/Makefile
```diff
@@ -71,7 +71,7 @@ srams: sram_generator
 sram_generator: $(SRAM_CONF)
 
 # This should be built alongside $(SMEMS_FILE)
```
abejgonzalez:
```suggestion
# This should be built alongside $(TOP_SMEMS_FILE)
```

PR #327, opened 2019-11-05
Issue + PR Templates

abejgonzalez:
These are the current templates that BOOM uses for issues and PRs. I figure this would be a good time to add it so that future issues/PRs can be a bit more structured.

Do people want specific things added/removed?


.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,28 @@
+<!-- choose all that apply -->
+**Type of issue**: bug report | feature request | question | other enhancement
+
+<!-- if the issue is NOT a question, fill out the following sections -->
+
+<!-- choose one -->
+**Impact**: no rtl change | rtl refactoring | new rtl | unknown
```
alonamid: Chipyard is not BOOM. Not everything is RTL.
I think there is a low likelihood of someone filing an issue and knowing the impact ahead of time

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,28 @@
...
+
+<!-- choose one -->
+**Development Phase**: request | proposal
```
alonamid: Not sure if this is necessary

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,13 @@
+**Related issue**: <!-- if applicable -->
+
+<!-- choose one -->
+**Type of change**: bug fix | new feature | other enhancement
+
+<!-- choose one -->
+**Impact**: no rtl change | rtl refactoring | new rtl | unknown
```
alonamid: Same as before. Not everything is RTL

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,13 @@
...
+
+<!-- choose one -->
+**Development Phase**: proposal |  implementation
```
alonamid: Same as before. Not sure this is necessary

seldridge:
For the RTL/non-RTL that @alonamid is talking about... you could use multiple templates. This is [PR'd on FIRRTL](https://github.com/freechipsproject/firrtl/pull/1206/files) to split bug reports from feature requests. Users then get a prompt when they file an issue like https://github.com/seldridge/issue-pr-template-testing/issues/new/choose.

.github/ISSUE_TEMPLATE.md
davidbiancolin: I think RTL changes should be called out somehow though.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,28 @@
...
+**What is the expected behavior?**
+
+**Please tell us about your environment:**
+<!--    (examples)
+    - version: `chipyard git commit hash`
+    - os: `Linux knight 4.4.0-92-generic #115-Ubuntu SMP Thu Aug 10 09:04:33 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux`
+    - other: `prior steps taken/documentation followed/...`
+-->
+
+**What is the use case for changing the behavior?**
```
jwright6323: I never liked this sentence; it doesn't really make sense to me. Maybe something like this would be clearer?
```suggestion
**What is a motivating example for changing the behavior?**
```

abejgonzalez:
Ok I tried to split up the issue templates a bit according to @seldridges recommendation. Additionally, I made the "Impact" split based on if it is RTL/SW/Other/Unknown, I think that is a bit clearer although Im sure there is a better split.

PR #331, opened 2019-11-06
Partition `vlsi/build` on design

### LINKED ISSUE #187: Partition vlsi/build by long_name

abejgonzalez:
Splits `vlsi/build` on the type of design.

Covers #220. Fixes #187

### DIFF
--- a/vlsi/Makefile
+++ b/vlsi/Makefile
@@ -28,11 +28,16 @@ ifeq ($(tech_name),asap7)
 else
 	MACROCOMPILER_MODE ?= -l $(SMEMS_CACHE) -hir $(SMEMS_HAMMER)
 endif
-OBJ_DIR            ?= $(vlsi_dir)/build
-ENV_YML            ?= $(vlsi_dir)/env.yml
-INPUT_CONFS        ?= example.yml
+ENV_YML            ?= $(vlsi_dir)/bwrc-env.yml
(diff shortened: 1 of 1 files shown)


vlsi/Makefile
```diff
@@ -28,11 +28,16 @@ ifeq ($(tech_name),asap7)
 else
 	MACROCOMPILER_MODE ?= -l $(SMEMS_CACHE) -hir $(SMEMS_HAMMER)
 endif
-OBJ_DIR            ?= $(vlsi_dir)/build
-ENV_YML            ?= $(vlsi_dir)/env.yml
-INPUT_CONFS        ?= example.yml
+ENV_YML            ?= $(vlsi_dir)/bwrc-env.yml
```
harrisonliew: This should be left as `env.yml` for Chipyard. Ditto not including tech-plugin's `bwrc.yml` below.

harrisonliew: Probably can just go into VLSI documentation under `Tutorial` or `Advanced-Usage` to teach people how to populate these ymls.

abejgonzalez: My bad, I copy-pasted all the lines but forgot that Chipyard is a bit different.

PR #334, opened 2019-11-12
Let user specify existing PREFIX/RISCV for toolchain installs

jerryz123:
I like to manage my `$RISCV` independently.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,14 +31,14 @@ jobs:
                     .circleci/create-hash.sh
             - restore_cache:
                 keys:
-                    - riscv-tools-installed-v2-{{ checksum "../riscv-tools.hash" }}
+                    - riscv-tools-installed-v3-{{ checksum "../riscv-tools.hash" }}
             - run:
                 name: Building riscv-tools toolchain
                 command: |
                     .circleci/build-toolchains.sh riscv-tools
                 no_output_timeout: 120m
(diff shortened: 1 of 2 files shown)


alonamid:
In this case, if you've installed riscv-tools, and now decide you want to install esp-tools, you need to actively clean your RISC-V environment variable (otherwise it will silently now install esp-tools). Is there a way to indicate this to the unsuspecting user?

jerryz123:
> In this case, if you've installed riscv-tools, and now decide you want to install esp-tools, you need to actively clean your RISC-V environment variable (otherwise it will silently now install esp-tools). Is there a way to indicate this to the unsuspecting user?

Why would I need to actively clean my RISCV directory? The script should just overwrite existing files in there when it builds esp-tools.

a0u:
What @alonamid probably meant was, if one intends to build an alternative toolchain but has already sourced `env.sh`, the user must remember to unset `RISCV` to avoid overwriting the original installation.

I am fine with supporting arbitrary installation paths, but an explicit command line option (`-d <path>`) would be preferable instead.  Regardless of the approach, the user-specified input should also be converted into an absolute path, if not already one, to avoid a subsequent configure error.

sagark:
+1 to @a0u

jerryz123:
I added a `--prefix` option. I also incremented the CI cached tools index, so the build-tools tests reran.

PR #336, opened 2019-11-12
Faster FireMarshal Initialization

NathanTP:
This PR bumps marshal and moves some submodules around. Specifically:

- Marshal no longer initializes submodules by default. Instead, if you want linux you'll need to manually run ./init-submodules.sh
- The sha3 FireMarshal workloads are moving to the sha3 generator itself (see firesim/firemarshal#99 and ucb-bar/sha3#17)
- FireSim is bumped to account for the FireMarshal changes (see firesim/firesim#425)

This should wait until the dependent PRs (listed above) actually go through, I'll then re-update the submodule pointers to the merge commits and we'll be good to go. This should be ready for testing/review now.

### DIFF
--- a/scripts/init-submodules-no-riscv-tools.sh
+++ b/scripts/init-submodules-no-riscv-tools.sh
@@ -53,6 +53,8 @@ git config --unset submodule.vlsi/hammer-synopsys-plugins.update
 git config --unset submodule.vlsi/hammer-mentor-plugins.update
 
 git config --unset submodule.generators/sha3.update
+# Shallow clone by default (recursive clone would get linux)
+git submodule update --init generators/sha3
 
 if [ $NO_FIRESIM = false ]; then
   echo "initializing firesim"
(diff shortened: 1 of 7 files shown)


scripts/init-submodules-no-riscv-tools.sh
```diff
@@ -32,6 +32,9 @@ for name in toolchains/*/*/ ; do
 done
 git config submodule.toolchains/qemu.update none
 
+# Don't automatically initialize generators with big submodules (e.g. linux source)
+git config submodule.generators/sha3.update none
```
abejgonzalez: This doesn't work as intended, it will not check out all the sha3 sources. Instead something like this should be better:

```
git config submodule.generators/sha3.update none
git submodule update --init --recursive
git config --unset submodule.generators/sha3.update
```

abejgonzalez: This is causing the CI to fail.

PR #337, opened 2019-11-14
Bump rocket-chip, chisel, firrtl for aspect support

colinschmidt:
This also includes an example aspect.

Currently opening this to see if CI is passing and then fix related problems.


colinschmidt:
I'm splitting out the rocket-chip/chisel bump part of this into a separate PR

PR #347, opened 2019-11-24
Revamp the config system for Top/Harness

jerryz123:
It is strange that we use the config/mixin system to parameterize everything except for the Top/Harness, for which we still have to use a system of maintaining copies of Tops and Harnesses with precisely chosen traits to do what we want. There is also a lack of concrete documentation regarding the interactions between keys, traits, and mixins, and the standard patterns for using them. I restuctured the config system around what I believe should be the correct patterns for using these, and I define them below.

- **Keys** specify some parameter which controls some custom widget. 
  - Keys should be **Option types**, and the default value of `None` should mean no change to the system. (The custom widget is not instantiated and connected)
  - Keys should be defined and documented in sub-projects, since they generally deal with some specific block, and not system-level integration.
  - Users developing a custom block should define a custom key for their block as an `Option` type, with the default value set to `None`, preferably
- **Traits** specify that the System has been parameterized to read some custom Key and optionally instantiate and connect that widget. 
  - Traits **should not** mandate the instantiation of that widget. In other words, we should try to encourage traits to be written with `CanHave` semantics. Many Rocketchip traits already have `CanHave` semantics, since they are no-ops if their corresponding keys are set to None. 
  - Traits should be defined and documented in subprojects, alongside their corresponding keys. This is because they primarily deal with instantiating and connecting some custom block, and not system-level integration.
  - Users developing a custom block should define their own `CanHave` trait and add it to whatever Top they are using (either example/Top or a copy of it).
- **Mixins** set the keys to a non-default value. 
  - Mixins which affect the Harness/Top, should set the `BuildTop` key to instantiate whatever custom block is needed in the Harness and wire the Top ports to whatever is desired. See the WithTSI or WithGPIO mixins for examples. Mixins should also set the parameters key for the custom widget (for example, `BlockDeviceKey`)
  - These should be defined and documented centrally in Chipyard, since these are what users will primarily interact with to customize their system.
  - Users developing a custom block should define a mixin setting the key for their block in ConfigMixins.

Some other notes from these changes: 

- There is only one `BuildTop` key. Adding ports to a Top is done by finding the top via `p(BuildTop, site)`, and calling the method of that top which connects the ports. 
  - This enables intuitive composability of Tops. E.g:  `WithGPIO ++ WithPWMTL` results in a Top with both GPIO and PWM pins, with no other changes necessary.
- By adding the `success` signal to the `BuildTop` function, we avoid the need to instantiate a separate TestHarness for DTM-bringup.
- The `ConfigMixins` now contain mixins that set both the `BuildTop` key (which controls the Top ports), as well as the widget keys (which control the instantiation and parameterization of the widgets). Thus one mixin is sufficient to both set custom ports, and connect those custom ports to a custom module.
- The `WithNoGPIO` mixin is an unfortunate byproduct of `sifive-blocks` not setting a default value for `PeripheryGPIOKey`. I currently have a PR open for this which should allow us to remove all those mixins from the default. https://github.com/sifive/sifive-blocks/pull/123
- Encouraging traits to be written with `CanHave` semantics simplifies convergence between FireSim/Chipyard targets
  - The `FireSimDUT` and Chipyard's `Top` can be merged into one `Top` with a superset of all `CanHave` traits. Then, the divergence between the two designs can be configured at the mixin level, rather than the trait level


generators/example/src/main/scala/ConfigMixins.scala
```diff
@@ -29,110 +30,78 @@ import ConfigValName._
...
 
-// DOC include start: WithGPIO
-/**
- * Class to add in GPIO
- */
 class WithGPIO extends Config((site, here, up) => {
   case PeripheryGPIOKey => List(
     GPIOParams(address = 0x10012000, width = 4, includeIOF = false))
+  case BuildTop => (clock: Clock, reset: Bool, p: Parameters, success: Bool) => {
+    val top = up(BuildTop, site)(clock, reset, p, success)
```
davidbiancolin: This is cute.

davidbiancolin:
[I know this WIP, these are mostly just thoughts on the current implementation.]

I think this is a step in the right direction but i think the problem with the current implementation is that leans heavily on there existing only a single top class with the complete set of all possible `canHave` traits (or you're forced to extend it). I'm not sure if this restriction is a prudent one.

It might be better to relax the type BuildTop elaborates (say, all the way to a Module or LazyModule), and then engineer a more flexible solution to generate the harness around it. The partial-function-based solution we use in FireSim (see BridgeBinders.scala) is one attempt at this, but it too has its limitations.

A second independent point is that relying on `up` to gradually build up the test harness might prove to be a little bit dangerous. What if multiple mixins try to connect to the same interfaces? It's perfectly legal and even a powerful feature, but only if you order your configs right -- this is another most common gotcha when using `Config`.

jerryz123:
> I think this is a step in the right direction but i think the problem with the current implementation is that leans heavily on there existing only a single top class with the complete set of all possible `canHave` traits (or you're forced to extend it). I'm not sure if this restriction is a prudent one.

I'm assuming a world where all the traits can be parameterized to be no-ops. In that case, I'm not sure what issues would arise from making the standard pattern "add your custom trait to the base Top". Ideally we would be able to define new Tops with custom Traits at runtime, but I'm not sure how to do that, and this seems like a better middle-ground for now.

> It might be better to relax the type BuildTop elaborates (say, all the way to a Module or LazyModule), and then engineer a more flexible solution to generate the harness around it. The partial-function-based solution we use in FireSim (see BridgeBinders.scala) is one attempt at this, but it too has its limitations.

The partial function stuff is interesting. My concern with that approach is that it would require defining another layer of abstraction wrapping all the methods defined in the various traits, if I understand correctly.

> A second independent point is that relying on `up` to gradually build up the test harness might prove to be a little bit dangerous. What if multiple mixins try to connect to the same interfaces? It's perfectly legal and even a powerful feature, but only if you order your configs right -- this is another most common gotcha when using `Config`.

Yeah, the examples for this case I can think of are the block device and success ports. I hope that in the documentation, we can explain clearly enough why things like `WithDTM ++ WithTSI` and `WithSimBlockDevice ++ WithBlockDeviceModel` are nonsensical. We could even potentially add some way to check in the `BuildTop` definition that the port has not already been connected, but I can't think of the best way to do that. Otherwise, it seems rare to have multiple conflicting mixins to attach to the same custom interface.

generators/example/src/main/scala/TestHarness.scala
```diff
@@ -28,9 +38,8 @@ class TestHarness(implicit val p: Parameters) extends Module {
   // force Chisel to rename module
   override def desiredName = "TestHarness"
 
-  val dut = p(BuildTop)(clock, reset.toBool, p)
+  val dut = p(BuildTop)(clock, reset.toBool, p, io.success)
```
abejgonzalez: Also a slight nit with the current approach: currently the `dut` instance is being renamed to `TestHarness` in some/all cases (I haven't tested extensively). We should add a `suggestName("dut")` here if that is still occurring.

jerryz123: Wait why is this happening? `desiredName` should be changing the name of the Harness, not the dut.

jwright6323: This has been happening for a while, I believe. I noticed it a few weeks ago and didn't really think much of it.

abejgonzalez: image

This is what I see currently... where the `TestHarness` is called `testHarness` (dictated by the `TestDriver.v`) and the instance of `Top` is called `TestHarness`. Frankly, I don't know why it is happening... but I haven't looked into it past what I posted here.

jwright6323: It's entirely possible this is a weird chisel bug

jwright6323: It's not coming from the `override def desiredName` line; I changed the name and `dut` is still named `TestHarness`.

jwright6323: Interestingly `suggestName` also does not work. This seems like a @jackkoenig question.

harrisonliew: Yeah, I'm pretty sure `override def desiredName` doesn't work for an externally-included Blackbox verilog. I once solved this by putting my Verilog inline (using `setInline` where the first argument was `s"${this.desiredName}.v"`).

jwright6323: @harrisonliew I'm not sure I follow- there's no blackbox here.

jerryz123:
I added more documentation on how custom traits work. This should be good to go now, pending sub-project PRs. I'd like to get this in before the upcoming Chipyard release.

generators/example/src/main/scala/ConfigMixins.scala
```diff
@@ -33,23 +34,26 @@ import ConfigValName._
...
- */
+// DOC include start: gpio mixin
 class WithGPIO extends Config((site, here, up) => {
   case PeripheryGPIOKey => Seq(
     GPIOParams(address = 0x10012000, width = 4, includeIOF = false))
+  case BuildTop => (clock: Clock, reset: Bool, p: Parameters, success: Bool) => {
+    val top = up(BuildTop, site)(clock, reset, p, success)
+    for (gpio <- top.gpio) {
+      for (pin <- gpio.pins) {
+        pin.i.ival := false.B
```
jwright6323: Is this tying all of the GPIO inputs to 0 inside the DUT?

colinschmidt: This seems weird. I would not expect this restriction based on the name of the mixin.

generators/example/src/main/scala/ConfigMixins.scala
```diff
@@ -59,92 +63,58 @@ class WithUART extends Config((site, here, up) => {
...
- * Class to specify a top level BOOM and/or Rocket system with PWM
- */
-// DOC include start: WithPWMTop
-class WithPWMTop extends Config((site, here, up) => {
-  case BuildTop => (clock: Clock, reset: Bool, p: Parameters) =>
-    Module(LazyModule(new TopWithPWMTL()(p)).module)
+class WithDTM extends Config((site, here, up) => {
+  case BuildTop => (clock: Clock, reset: Bool, p: Parameters, success: Bool) => {
+    val top = up(BuildTop, site)(clock, reset, p, success)
+    top.reset := reset.asBool | top.debug.get.ndreset
```
jwright6323: Is this circumventing reset synchronizers?

generators/example/src/main/scala/ConfigMixins.scala
colinschmidt: Will this cause problems is someone includes multiple of the same mixin?
It seems to work as long as each buildtop method is idempotent, but I'm not sure thats the case. I.e. if any buildtop methods construct hardware rather than just making connections you might have problems.

generators/example/src/main/scala/RocketConfigs.scala
```diff
@@ -9,149 +9,191 @@ import freechips.rocketchip.config.{Config}
...
 
 // DOC include start: JtagRocket
 class jtagRocketConfig extends Config(
-  new WithDTMTop ++                                        // use top with dtm
-  new freechips.rocketchip.subsystem.WithJtagDTM ++        // add jtag+DTM module to coreplex
+  new WithDTM ++                                           // use top with dtm
+  new WithNoGPIO ++
   new WithBootROM ++
   new WithUART ++
+  new freechips.rocketchip.subsystem.WithJtagDTM ++        // add jtag+DTM module to coreplex
```
colinschmidt: Should we consider rolling this into the other mixin?
We could have `WithDTM` and `WithJTAGDTM`?

generators/firechip/src/main/scala/TargetConfigs.scala
```diff
@@ -47,13 +47,17 @@ class WithUARTKey extends Config((site, here, up) => {
      nRxEntries = 256))
 })
 
+class WithSerial extends Config((site, here, up) => {
```
colinschmidt: Shouldn't there be other parameters for TSI (at least bitwdith?) rolled into this key?

variables.mk
```diff
@@ -135,7 +135,8 @@ sim_common_files           ?= $(build_dir)/sim_files.common.f
 # java arguments used in sbt
 #########################################################################################
 JAVA_HEAP_SIZE ?= 8G
-JAVA_ARGS ?= -Xmx$(JAVA_HEAP_SIZE) -Xss8M -XX:MaxPermSize=256M
+JAVA_ARGS ?= -Xmx$(JAVA_HEAP_SIZE) -Xss8M -XX:MaxPermSize=256M -Dsbt.ivy.home=/scratch/jerryz/.ivy2 -Dsbt.global.base=/scratch/jerryz/.sbt
```
colinschmidt: I don't think we really want to commit your scratch dir.

generators/example/src/main/scala/GCD.scala
```diff
@@ -0,0 +1,207 @@
...
+  val gcd = Output(UInt(w.W))
+  val busy = Output(Bool())
+}
+
+trait GCDTopIO extends Bundle {
+  val gcd_busy = Output(Bool())
+}
+
+// DOC include start: GCD blackbox
+class GCDMMIOBlackBox(w: Int) extends BlackBox(Map("WIDTH" -> IntParam(w))) with HasBlackBoxResource
```
jwright6323: Something like this would make the code below a little DRYer:

```suggestion
trait IsGCDImplementation extends BaseModule { val io = IO(new GCDIO(w)) }

class GCDMMIOBlackBox(w: Int) extends BlackBox(Map("WIDTH" -> IntParam(w))) with HasBlackBoxResource with IsGCDImplementation
```

then delete the other `io` definitions.

generators/example/src/main/scala/GCD.scala
```diff
@@ -0,0 +1,207 @@
...
+{
+  val io = IO(new GCDIO(w))
+
+  addResource("/vsrc/GCDMMIOBlackBox.v")
+}
+// DOC include end: GCD blackbox
+
+// DOC include start: GCD chisel
+class GCDMMIOChiselModule(w: Int) extends Module
```
jwright6323:
```suggestion
class GCDMMIOChiselModule(w: Int) extends Module with IsGCDImplementation
```

... etc.

generators/example/src/main/scala/GCD.scala
```diff
@@ -0,0 +1,207 @@
...
+  val reset: Reset
+
+
+  // How many clock cycles in a PWM cycle?
+  val x = Reg(UInt(params.width.W))
+  val y = Wire(new DecoupledIO(UInt(params.width.W)))
+  val gcd = Wire(new DecoupledIO(UInt(params.width.W)))
+  val status = Wire(UInt(2.W))
+
+  if (params.useBlackBox) {
```
jwright6323: Something like 

```
val impl = Module(if (params.useBlackBox) { new GCDMMIOBlackBox(params.width) } else { new GCDMMIOChiselModule(params.width) })
...etc
```

docs/Customization/Custom-Chisel.rst
```diff
@@ -0,0 +1,38 @@
...
+To add a submodule to the Chipyard framework, make sure that your project is organized as follows.
+
+.. code-block:: none
+
+    yourproject/
+        build.sbt
+        src/main/scala/
+            YourFile.scala
+
+Put this in a git repository and make it accessible.
```
abejgonzalez: I don't think you want to force people to make it a git repository. Instead, I would move the part about putting it in a git repo towards the bottom and make it known that it is optional but recommended.

docs/Customization/Keys-Traits-Configs.rst
```diff
@@ -0,0 +1,95 @@
...
+Keys specify some parameter which controls some custom widget. Keys should typically be implemented as **Option types**, with a default value of ``None`` that means no change in the system. In other words, the default behavior when the user does not explicitly set the key should be a no-op.
+
+Keys should be defined and documented in sub-projects, since they generally deal with some specific block, and not system-level integration. (We make an exception for the example GCD widget).
+
+.. literalinclude:: ../../generators/example/src/main/scala/GCD.scala
+    :language: scala
+    :start-after: DOC include start: GCD key
+    :end-before: DOC include end: GCD key
+
+The object within a key is typically a ``case class XXXParams``, which defines a set of parameters which some block accepts. For example, the GCD widget's ``GCDParams`` parameterizes its address, operand widths, whether the widget should be connected by Tilelink or AXI4, and whether the widget should use the blackbox-verilog implementation, or the Chisel implementation
```
abejgonzalez:
```suggestion
The object within a key is typically a ``case class XXXParams``, which defines a set of parameters which some block accepts. For example, the GCD widget's ``GCDParams`` parameterizes its address, operand widths, whether the widget should be connected by Tilelink or AXI4, and whether the widget should use the blackbox-verilog implementation, or the Chisel implementation.
```

docs/Customization/Keys-Traits-Configs.rst
```diff
@@ -0,0 +1,95 @@
...
+    :language: scala
+    :start-after: DOC include start: GCD params
+    :end-before: DOC include end: GCD params
+
+Accessing the value stored in the key is easy in Chisel, as long as the ``implicit p: Parameters`` object is being passed through to the relevant module. For example, ``p(GCDKey).get.address`` returns the address field of ``GCDParams``. Note this only works if ``GCDKey`` was not set to ``None``, so your Chisel should check for that case!
+
+Traits
+------
+
+Typically, most custom blocks will need to modify the behavior of some pre-existing block. For example, the GCD widget needs the Top module to instantiate and connect the widget via Tilelink, generate a top-level ``gcd_busy`` port, and connect that to the module as well. Traits let us do this without modifying the existing code for the ``Top``, and enables compartmentalization of code for different custom blcoks.
```
abejgonzalez:
```suggestion
Typically, most custom blocks will need to modify the behavior of some pre-existing block. For example, the GCD widget needs the ``Top`` module to instantiate and connect the widget via Tilelink, generate a top-level ``gcd_busy`` port, and connect that to the module as well. Traits let us do this without modifying the existing code for the ``Top``, and enables compartmentalization of code for different custom blcoks.
```

docs/Customization/Keys-Traits-Configs.rst
```diff
@@ -0,0 +1,95 @@
...
+
+Top-level traits specify that the Top has been parameterized to read some custom Key and optionally instantiate and connect a widget defined by that Key. Traits **should not** mandate the instantiation of custom logic. In other words, traits should be written with ``CanHave`` semantics, where the default behavior when the Key is unset is a no-op.
```
abejgonzalez:
```suggestion
Top-level traits specify that the ``Top`` has been parameterized to read some custom Key and optionally instantiate and connect a widget defined by that Key. Traits **should not** mandate the instantiation of custom logic. In other words, traits should be written with ``CanHave`` semantics, where the default behavior when the Key is unset is a no-op.
```

docs/Customization/Keys-Traits-Configs.rst
```diff
@@ -0,0 +1,95 @@
...
+
+Top-level traits should be defined and documented in subprojects, alongside their corresponding Keys. The traits should then be added to the Top being used by Chipyard.
```
abejgonzalez:
```suggestion
Top-level traits should be defined and documented in subprojects, alongside their corresponding Keys. The traits should then be added to the ``Top`` being used by Chipyard.
```

docs/Customization/Keys-Traits-Configs.rst
```diff
@@ -0,0 +1,95 @@
...
+
+Below we see the traits for the GCD example. The Lazy trait connects the GCD module to the Diplomacy graph, while the Implementation trait causes the Top to instantiate an additional port and concretely connect it to the GCD module.
```
abejgonzalez:
```suggestion
Below we see the traits for the GCD example. The Lazy trait connects the GCD module to the Diplomacy graph, while the Implementation trait causes the ``Top`` to instantiate an additional port and concretely connect it to the GCD module.
```

docs/Customization/Keys-Traits-Configs.rst
```diff
@@ -0,0 +1,95 @@
...
+.. literalinclude:: ../../generators/example/src/main/scala/RocketConfigs.scala
+    :language: scala
+    :start-after: DOC include start: GCDTLRocketConfig
+    :end-before: DOC include end: GCDTLRocketConfig
+
+
+BuildTop
+--------
+
+The ``BuildTop`` key is special, because sometimes, we need to instantiate TestHarness modules to interface with a custom widget. The ``BuildTop`` key provides a function which can call some method of the Top to instantiate these TestHarness modules. Since the ``BuildTop`` key is called from the TestHarness, these modules will appear in the TestHarness. The config system also lets the ``BuildTop`` key look recursively into previous definitions of itself. This enables composability of the Top configurations.
```
abejgonzalez: As you can see from my previous suggestions, I think `Top` should be in "code" format. I also think that `TestHarness` should be the same. Basically, I say that anything that is present in the code should be "code format".

abejgonzalez: I would double check that this is done everywhere.

docs/Customization/Keys-Traits-Configs.rst
```diff
@@ -0,0 +1,95 @@
...
+    :start-after: DOC include start: tsi mixin
+    :end-before: DOC include end: tsi mixin
+
+
+.. literalinclude:: ../../generators/example/src/main/scala/ConfigMixins.scala
+    :language: scala
+    :start-after: DOC include start: gpio mixin
+    :end-before: DOC include end: gpio mixin
+
+The leftmost mixin is ``WithGPIO``, so at elaboration time, the ``BuildTop`` key is set according to that mixin. This mixin looks up the previously set value of ``BuildTop`` by ``up(BuildTop, site)`` and calls that function to get the Top  that would return. In this configuration, the previous ``BuildTop`` is set by ``WithTSI``. Thus after complete evaluation, the TSI ports are connected by the ``BuildTop`` function defined in ``WithTSI``, and the GPIO ports are connected by the ``BuildTop`` function defined in ``WithGPIO``.
```
abejgonzalez:
```suggestion
The leftmost mixin is ``WithGPIO``, so at elaboration time, the ``BuildTop`` key is set according to that mixin. This mixin looks up the previously set value of ``BuildTop`` by ``up(BuildTop, site)`` and calls that function to get the Top that would return. In this configuration, the previous ``BuildTop`` is set by ``WithTSI``. Thus after a complete evaluation, the TSI ports are connected by the ``BuildTop`` function defined in ``WithTSI``, and the GPIO ports are connected by the ``BuildTop`` function defined in ``WithGPIO``.
```

abejgonzalez: This section might be a bit confusing if people know the order of mixin evaluation. First the `WithTSI` would be evaluated then the `WithGPIO`. So maybe it would be better to explain it that order.

I would also change the wording of the "We need to both instantiate the TSI serial adapter, and tie off the GPIO pins" sentence to mention that the `WithGPIO` instantiates X GPIO ports and ties them off.

docs/Customization/MMIO-Peripherals.rst
```diff
@@ -0,0 +1,133 @@
+.. _mmio-accelerators:
+
+MMIO Peripherals
+==================
+
+The easiest way to create a MMIO peripheral is to use the ``TLRegisterRouter`` or ``AXI4RegisterRouter`` widgets, which abstracts away the details of handling the interconnect protocols and provides a convenient interface for specifying memory-mapped registers. Since Chipyard and Rocketchip SoCs primarily use Tilelink as the on-chip interconnect protocol, this section will primarily focus on designing Tilelink-based peripherals. However, see ``generators/example/src/main/scala/GCD.scala`` for how an example AXI4 based peripheral is defined and connected to the Tilelink graph through converters.
```
abejgonzalez: Rocketchip -> Rocket Chip

docs/Customization/MMIO-Peripherals.rst
```diff
@@ -0,0 +1,133 @@
...
+
+
+Connecting by TileLink
+----------------------
+
+Once you have these classes, you can construct the final peripheral by extending the ``TLRegisterRouter`` and passing the proper arguments.
+The first set of arguments determines where the register router will be placed in the global address map and what information will be put in its device tree entry.
+The second set of arguments is the IO bundle constructor, which we create by extending ``TLRegBundle`` with our bundle trait.
+The final set of arguments is the module constructor, which we create by extends ``TLRegModule`` with our module trait.
+Notice how we can create an analagous AXI4 version of our peripheral.
```
abejgonzalez:
```suggestion
Notice how we can create an analogous AXI4 version of our peripheral.
```

generators/example/src/main/scala/ConfigMixins.scala
jerryz123: Yeah you are right non-idempotent mixins will cause strange behavior with this. Is that a issue we should be concerned about?

generators/example/src/main/scala/ConfigMixins.scala
jerryz123: This code was copied from the older `WithGPIOTop` mixin. I'm not sure what the right thing to do here is, but I would be happy to implement it

docs/Customization/MMIO-Peripherals.rst
```diff
@@ -0,0 +1,133 @@
...
+    :language: scala
+    :start-after: DOC include start: Top
+    :end-before: DOC include end: Top
+
+Just as we need separate traits for ``LazyModule`` and module implementation, we need two classes to build the system.
+The ``Top`` class contains the set of traits which parameterize and define the ``Top``. Typically these traits will optionally add IOs or peripherals to the ``Top``.
+The ``Top`` class includes the pre-elaboration code and also a ``lazy val`` to produce the module implementation (hence ``LazyModule``).
+The ``TopModule`` class is the actual RTL that gets synthesized.
+
+And finally, we create a configuration class in ``generators/example/src/main/scala/Configs.scala`` that uses this mixin.
```
abejgonzalez: Which mixin are you referring to?

docs/Customization/MMIO-Peripherals.rst
```diff
@@ -0,0 +1,133 @@
...
+-------
+
+Now we can test that the GCD is working. The test program is in ``tests/gcd.c``.
+
+.. literalinclude:: ../../tests/gcd.c
+    :language: c
+
+This just writes out to the registers we defined earlier.
+The base of the module's MMIO region is at 0x2000 by default.
+This will be printed out in the address map portion when you generated the verilog code.
```
abejgonzalez: I think this would be a good place to mention that there are `json` files that have the address maps.

generators/example/src/main/scala/ConfigMixins.scala
jerryz123: This is called from outside the DUT (in the TestHarness), so it should be tying them to 0 outside the DUT.

docs/Customization/MMIO-Peripherals.rst
```diff
@@ -0,0 +1,133 @@
...
+
+Compiling this program with ``make`` produces a ``gcd.riscv`` executable.
+
+Now with all of that done, we can go ahead and run our simulation.
+
+.. code-block:: shell
+
+    cd sims/verilator
+    make CONFIG=GCDTLRocketConfig
+    ./simulator-example-GCDTLRocketConfig ../../tests/gcd.riscv
```
abejgonzalez: Can you also show the `make CONFIG=... BINARY=... run-binary` version here?

generators/example/src/main/scala/RocketConfigs.scala
jerryz123: Then you'll have duplication of code for connecting the DTM signals in the testharness. I think these two mixins do sufficiently different things that its best to leave them separate.
`WithDTM` instantiates the debug module with the DMI protocol, while `WithJtagDTM` adds JTAG to the Debug module's set of supported protocols.

generators/firechip/src/main/scala/TargetConfigs.scala
jerryz123: Can we leave that to future work? It doesn't look like testchipip exports a clean way to configure that yet.

variables.mk
jerryz123: :man_facepalming:

generators/example/src/main/scala/ConfigMixins.scala
jwright6323: I think we want a separate PR that adds IO cell blackboxes to have "Correct" functionality. That's been a work in progress for a while, but we can reprioritize it.

generators/example/src/main/scala/ConfigMixins.scala
jerryz123: Could you clarify? This was copied from the old DTMTop module, and I'm not familiar with some of its mechanics.

generators/example/src/main/scala/RocketConfigs.scala
jerryz123: I suppose the inline comment can be corrected

docs/Customization/RoCC-Accelerators.rst
```diff
@@ -0,0 +1,69 @@
...
+
+.. code-block:: scala
+
+    class WithCustomAccelerator extends Config((site, here, up) => {
+      case BuildRoCC => Seq((p: Parameters) => LazyModule(
+        new CustomAccelerator(OpcodeSet.custom0 | OpcodeSet.custom1)(p)))
+    })
+
+    class CustomAcceleratorConfig extends Config(
+      new WithCustomAccelerator ++ new RocketConfig)
```
abejgonzalez:
```suggestion
      new WithCustomAccelerator ++
      new RocketConfig)
```

generators/example/src/main/scala/BoomConfigs.scala
```diff
@@ -9,54 +9,72 @@ import freechips.rocketchip.config.{Config}
 // ---------------------
 
 class SmallBoomConfig extends Config(
-  new WithTop ++                                            // use normal top
+  new WithTSI ++                                            // use normal top
```
abejgonzalez: I would change the comment to be something like "use testchipip offchip link"

generators/example/src/main/scala/ConfigMixins.scala
abejgonzalez: While I agree with John on this, I don't think we need to block on this? Also, it might be good to add a comment that this is connecting from outside the DUT.

jwright6323: > separate PR 

means it's not blocking

abejgonzalez: While I agree, eventually this will be replaced by the IO cell blackbox work John mentioned... so maybe just have a comment saying this needs to be done or else FIRRTL will error?

generators/example/src/main/scala/ConfigMixins.scala
jwright6323: Unless something's changed (which is possible), this is connecting the test-harness-level clock and reset to the DUT. That should be fine for simulation, but we might want to note that in a real chip you'd probably have an async top-level reset and a `ResetCatchAndSync` somewhere inside the chip. I was asking the question because I don't actually know if that `ResetCatchAndSync` is in the `RocketSubsystem` already or not (maybe @colinschmidt can chime in).

generators/example/src/main/scala/TestHarness.scala
abejgonzalez: Hmm it looks like this is still not resolved with the current PR. Have you tried addressing this? If not, I can try to ask some Chisel experts.

generators/example/src/main/scala/TestHarness.scala
```diff
@@ -28,45 +38,8 @@ class TestHarness(implicit val p: Parameters) extends Module {
   // force Chisel to rename module
   override def desiredName = "TestHarness"
```
abejgonzalez: You can get rid of this now that there is only 1 `TestHarness`

generators/example/src/main/scala/TestHarness.scala
jerryz123: This is outside my area of expertise, and I thought this was a pre-existing issue as well? Maybe we can resolve this in another PR

abejgonzalez:
Overall seems fine with me. John covered most of my cleanliness comments on the GCD stuff.

generators/example/src/main/scala/TestHarness.scala
abejgonzalez: Ok. Ill go ahead and make an issue.

docs/Customization/Custom-Chisel.rst
jerryz123: I think if we reorder the submodule part to the bottom, people following step-by-step will first create their custom files in a normal directory. Creating a git submodule out of a normal directory that already exists is nontrivial.

Instead I will add stronger language here explaining that while the submodule approach is optional, it is heavily recommended.

generators/example/src/main/scala/TestHarness.scala
abejgonzalez: Ok found out two solutions. Figured I would put them here since it would be quick to add in this PR:

1. Call `suggestName` on the LazyModule
```
val lazytop = LazyModule(new Top()(p))
lazytop.suggestName("top")
val top = Module(lazytop.module)
```

2. Override the `suggestedName` in the `Top` module
```
override lazy val module = new TopModule(this)
override lazy val suggestedName = "sugN"
```

I personally think that the 1st option is a bit better since that is the proper way to override the default name (even though it is a bit more verbose).

generators/example/src/main/scala/ConfigMixins.scala
jerryz123: It looks like Rocket Chip's TestHarness does 
```
  dut.reset := reset | dut.debug.map { debug => AsyncResetReg(debug.ndreset) }.getOrElse(false.B)
```

I guess we should be using the `AsyncResetReg` here?

jwright6323: Yeah, we should fix that, but it wasn't the only problem I was referring to. The other thing was that you likely won't have a `clock` pin; you probably have a PLL, etc. that would generate a clock you need to synchronize `reset` to.

jerryz123: So in Chipyard currently the hierarchy is `TestHarness -> Top`, while for real chips the hierarchy would likely look like `TestHarness -> ChipTop -> Top`. In BEAGLE the chip-specific clocking and reset needs are handled within `ChipTop`.

With the current BuildTop rules, it looks like the way to do this would be to call `p(BuildTop)` within `ChipTop`, creating something like:

```
class ChipTop(implicit val p: Parameters) extends RawModule {
  val sys_clk = Wire(Clock())
  val sys_rst = Wire(Bool())
  val sys_success = Wire(Bool())
  val sys = p(BuildTop)(sys_clock, sys_reset, p, sys_success)
  ...
}
```

The issue is currently we don't have this extra `ChipTop` layer within Chipyard. It seems like that would be a good thing to add in the future.

generators/example/src/main/scala/GCD.scala
jerryz123: The lack of parameterized traits complicates this, I wasn't able to get this to work with
```
trait HasGCDIO extends BaseModule {
  val w: Int
  val io = IO(new GCDIO(w))
}
class GCDMMIOBlackBox(val w: Int) extends BlackBox(Map("WIDTH" -> IntParam(w))) with HasBlackBoxResource
  with HasGCDIO
{
  addResource("/vsrc/GCDMMIOBlackBox.v")
}
```
I see the error:
```
[deprecated] class example.GCDTL$$anonfun$1$$anon$1 (1 calls): Unable to automatically infer cloneType on class example.GCDTL$$anonfun$1$$anon$1: constructor has parameters ($outer, x$2$1) that are not both immutable and accessible. Either make all parameters immutable and accessible (vals) so cloneType can be inferred, or define a custom cloneType method.
[error] (run-main-0) java.lang.ClassCastException: chisel3.internal.firrtl.Ref cannot be cast to chisel3.internal.firrtl.ModuleIO
[error] java.lang.ClassCastException: chisel3.internal.firrtl.Ref cannot be cast to chisel3.internal.firrtl.ModuleIO
[error]         at chisel3.experimental.BaseModule.$anonfun$getChiselPorts$2(Module.scala:271)
[error]         at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:238)
[error]         at scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:62)
[error]         at scala.collection.mutable.ResizableArray.foreach$(ResizableArray.scala:55)
[error]         at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:49)
[error]         at scala.collection.TraversableLike.map(TraversableLike.scala:238)
[error]         at scala.collection.TraversableLike.map$(TraversableLike.scala:231)
[error]         at scala.collection.AbstractTraversable.map(Traversable.scala:108)
[error]         at chisel3.experimental.BaseModule.getChiselPorts(Module.scala:270)
[error]         at chisel3.experimental.DataMirror$.modulePorts(Data.scala:138)
[error]         at chisel3.experimental.DataMirror$.fullModulePorts(Data.scala:147)
[error]         at chisel3.Data.bindingToString(Data.scala:357)
[error]         at chisel3.Bool.toString(Bits.scala:1025)
[error]         at java.lang.String.valueOf(String.java:2994)
```

docs/Customization/Custom-Chisel.rst
```diff
@@ -0,0 +1,38 @@
+.. _custom_chisel:
+
+Integrating into the Generator Build System
```
alonamid:
```suggestion
Integrating Chisel Files into the Generator Build System
```

docs/Customization/Custom-Chisel.rst
alonamid: If you're going into this level of detail, you might as well describe what needs to be in your custom build.sbt....

alonamid: (like, an example of a minimal custom build.sbt)

docs/Customization/MMIO-Peripherals.rst
jerryz123: Forgot a inline snippet. Fixed

generators/example/src/main/scala/TestHarness.scala
jerryz123: Fixed

docs/Customization/Keys-Traits-Configs.rst
```diff
@@ -0,0 +1,104 @@
...
+    :start-after: DOC include start: tsi mixin
+    :end-before: DOC include end: tsi mixin
+
+
+.. literalinclude:: ../../generators/example/src/main/scala/ConfigMixins.scala
+    :language: scala
+    :start-after: DOC include start: gpio mixin
+    :end-before: DOC include end: gpio mixin
+
+When ``WithGPIO ++ WithTSI`` is evaluated right to left, the call to ``up(BuildTop, site)`` in ``WithGPIO`` will reference the function defined in the ``BuildTop`` key of ``WithTSI``. Thus, at elaboration time, when the ``BuildTop`` function is called by the ``TestHarness``, first the ``BuildTop`` function in ``WithTSI`` will be evaluated. This connects the ``success`` signal of the ``TestHarness`` to the ``SerialAdapter`` enabled by ``WithTSI``. Then, the rest of the code in the ``BuildTop`` function of ``WithGPIO`` will execute, tieing off the top-level GPIO input pins. Thus the evaluation of the ``BuildTop`` functions in a completed config is "right-to-left", matching how the evaluation of the mixins at compile-time is also "right-to-left".
```
davidbiancolin: This _may_ be an appropriate time to explain, or at least allude to, what `site` is.

docs/Customization/DMA-Devices.rst
```diff
@@ -0,0 +1,35 @@
...
+.. literalinclude:: ../../generators/example/src/main/scala/Top.scala
+    :language: scala
+    :start-after: DOC include start: Top
+    :end-before: DOC include end: Top
+
+We use ``TLHelper.makeClientNode`` to create a TileLink client node for us.
+We then connect the client node to the memory system through the front bus (fbus).
+For more info on creating TileLink client nodes, take a look at :ref:`Client Node`.
+
+Once we've created our top-level module including the DMA widget, we can create a configuration for it as we did before.
```
alonamid: Do you mention anywhere else what is a "DMA widget"?

generators/example/src/main/scala/ConfigMixins.scala
colinschmidt: As long as non of our mixins are currently non-idempotent then I think just the warning is fine. Thanks.

generators/example/src/main/scala/ConfigMixins.scala
colinschmidt: I guess so. It would be annoying if someone was currently tracking dev chipyard to do a real tapeout but I don't know of anyone doing that so it can wait.

generators/example/src/main/scala/RocketConfigs.scala
colinschmidt: Ok thats fine. Just wanted to consider it for a bit.

generators/firechip/src/main/scala/TargetConfigs.scala
colinschmidt: Sure. We should file an issue on testchipip so we don't lose track of this feature request.

generators/example/src/main/scala/ConfigMixins.scala
colinschmidt: I don't think rocket-chip currently does catch and sync for the top-level reset. They may be moving in that direction based on recent PRs, but its certainly not in our tagged version yet.
I like it would be cool to have something that is a "reasonable" `ChipTop` in chipyard so we can make the knowledge of how to do this "safely" more widely known.
I think thats outside of the scope of this PR, in that this doesn't change the current state of chipyard w.r.t. clock&reset. But we should file an issue to track a "realistic clock and reset mixin".

generators/firechip/src/main/scala/TargetConfigs.scala
jerryz123: https://github.com/ucb-bar/testchipip/issues/63

generators/example/src/main/scala/ConfigMixins.scala
jerryz123: I'll file an issue for `ChipTop`, while I'll leave it to someone more familiar with the clock&reset issues to write that.

https://github.com/ucb-bar/chipyard/issues/385

docs/Customization/Incorporating-Verilog-Blocks.rst
```diff
@@ -102,54 +101,32 @@ diplomatic memory mapping on the system bus, we still have to
...
     :language: scala
     :start-after: DOC include start: GCD instance regmap
     :end-before: DOC include end: GCD instance regmap
 
-Advanced Features of RegField Entries
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-One significant difference from the PWM example is in the peripheral's
-memory map. ``RegField`` exposes polymorphic ``r`` and ``w`` methods
-that allow read- and write-only memory-mapped registers to be
```
alonamid: Why is this removed?

docs/Customization/Keys-Traits-Configs.rst
alonamid: +1

docs/Customization/Incorporating-Verilog-Blocks.rst
alonamid: Oh, I see it moved to the MIMO section

docs/Customization/DMA-Devices.rst
jerryz123: Added paragraph to beginning of section

docs/Customization/Keys-Traits-Configs.rst
jerryz123: I found it difficult to mention `site` without discussing the rest of the CDE system. I added a sub-section on CDEs to the Advanced-Concepts section, and linked to it from here.

@davidbiancolin @alonamid do you mind reviewing that page I just added? Its `docs/Advanced-Concepts/CDEs.rst`. I think that's the final set of changes that need reviewing.

docs/Advanced-Concepts/CDEs.rst
```diff
@@ -0,0 +1,108 @@
...
+
+    val params = Config(new WithX(true) ++ new WithY(true)) // "chain" together mixins
+    params(SomeKeyX) // evaluates to true
+    params(SomeKeyY) // evaluates to true
+    params(SomeKeyZ) // evaluates to false
+
+In this example, the evaluation of ``params(SomeKeyX)`` will terminate in the partial function defined in ``WithX(true)``, while the evaluation of ``params(SomeKeyY)`` will terminate in the partial function defined in ``WithY(true)``. Note that when no partial functions match, the evaluation will return the default value for that parameter.
+
+The real power of CDEs arises from the ``(site, here, up)`` parameters to the partial functions, which provide useful "views" into the global parameterization that the partial functions may access to determine a parameterization.
+
```
alonamid: "Additional information about the reasons behind CDE can be found in chapter 2 of https://www2.eecs.berkeley.edu/Pubs/TechRpts/2016/EECS-2016-89.pdf"

abejgonzalez:
Can you confirm that all the configs in Chipyard at least build with this approach?

jerryz123:
> Can you confirm that all the configs in Chipyard at least build with this approach?

Good call @abejgonzalez . Turns out the SHA3 Configs were all broken after the recent rocket chip bump. I fixed that and bumped the SHA3 submodule in this PR as well. All RocketConfigs elaborated verilog successfully. The BlockDevice and GCD configs (which utilize scala resources) also compiled in VCS correctly.

This is still waiting on several submodule PRs to get merged

- [ ] https://github.com/ucb-bar/testchipip/pull/58
- [ ] https://github.com/firesim/firesim/pull/434
- [ ] https://github.com/firesim/icenet/pull/17
- [ ] https://github.com/ucb-bar/sha3/pull/22

abejgonzalez:
> Can you confirm that all the configs in Chipyard at least build with this approach?

Did you also check the FireChip configs (make sure things build with FireSim)?

jerryz123:
> > Can you confirm that all the configs in Chipyard at least build with this approach?
> 
> Did you also check the FireChip configs (make sure things build with FireSim)?

Both FireSim and FireSimNoNIC targets work. Perhaps we should add these to the regressions.

generators/example/src/main/scala/ConfigMixins.scala
```diff
@@ -33,118 +35,91 @@ import ConfigValName._
...
+    }
+    top
+  }
 })
-// DOC include end: WithGPIO
+// DOC include end: gpio mixin
 
 /**
- * Class to add in UART
+ * Class to add in UAbRT
```
a0u: UAbRT?

zhemao: Seems like a typo here.

jerryz123: Fixed

jerryz123: Fixed

jerryz123:
Merging. CI fails because I did not merge the FireSim submodule PR, so @davidbiancolin can deal with the circular dependency madness.

PR #348, opened 2019-11-25
Add UART and UARTAdapter to all configs

abejgonzalez:
This PR adds a UART to all systems and adds a UARTAdapter in the TestHarness to talk to the UART.

Things to do:
- [x] Add UART to all configurations

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -119,7 +119,7 @@ lazy val rocketchip = freshProject("rocketchip", rocketChipDir)
   .dependsOn(chisel, hardfloat, rocketMacros, rocketConfig)
 
 lazy val testchipip = (project in file("generators/testchipip"))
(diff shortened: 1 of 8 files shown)


colinschmidt:
Does this affect simulation speed?

abejgonzalez:
Yes. But I don't think by much if you are not running code that outputs to the UART.

abejgonzalez:
@colinschmidt Do you want me to check something specifically?

abejgonzalez:
Note: This works with Linux boot in simulation with baudrate 115200 on the adapter.

abejgonzalez:
> LGTM. Is there some place in the docs where we should now mention what this config class does?

I just added a quick section on the adapter in the documentation. I can add more later if necessary.

PR #358, opened 2019-12-06
Bump all submodules for chisel 3.2.0 and rocket-chip august-2019

colinschmidt:
This is a subset of my previous PR with only the bumping changes. 
Things work locally so hopefully I don't many if any CI failures.


abejgonzalez:
@colinschmidt Ok BOOM is ready to go when you are.

PR #361, opened 2019-12-09
Control Core Example

abejgonzalez:
Example MultiCore configuration with a small "control" core attached on the side (i.e. a PMU). This was requested by @jwright6323 and @alonamid.


abejgonzalez:
Note: This doesn't choose what `hartId` the simple core has... I can change this if need be.

jwright6323:
I think we want it the way you wrote it - the control core should be the last hart ID.

abejgonzalez:
So that gets overridden by the `new boom.common.WithRenumberHarts ++` line. If we want it as the last `hartId` I can move it to after that line.

colinschmidt:
Looks good.

PR #362, opened 2019-12-11
[tracegen] Add tracegen support for the BOOM L1D

jerryz123:
Depends on #358 's BOOM bump

### DIFF
--- a/generators/tracegen/src/main/scala/Configs.scala
+++ b/generators/tracegen/src/main/scala/Configs.scala
@@ -10,7 +10,7 @@ import freechips.rocketchip.rocket.DCacheParams
 import freechips.rocketchip.tile.{MaxHartIdBits, XLen}
 import scala.math.{max, min}
 
-class WithTraceGen(params: Seq[DCacheParams], boom_params: Seq[DCacheParams] = Nil, nReqs: Int = 8192)
+class WithTraceGen(params: Seq[DCacheParams], nReqs: Int = 8192)
     extends Config((site, here, up) => {
   case TraceGenKey => params.map { dcp => TraceGenParams(
     dcache = Some(dcp),
@@ -32,7 +32,12 @@ class WithTraceGen(params: Seq[DCacheParams], boom_params: Seq[DCacheParams] = N
     memStart = site(ExtMem).get.master.base,
     numGens = params.size)
   }
-  case BoomTraceGenKey => boom_params.map { dcp => TraceGenParams(
+  case MaxHartIdBits => log2Ceil(params.size + up(BoomTraceGenKey, site).length) max 1
+})
+
+class WithBoomTraceGen(params: Seq[DCacheParams], nReqs: Int = 8192)
+    extends Config((site, here, up) => {
+  case BoomTraceGenKey => params.map { dcp => TraceGenParams(
     dcache = Some(dcp),
     wordBits = site(XLen),
     addrBits = 48,
@@ -52,9 +57,9 @@ class WithTraceGen(params: Seq[DCacheParams], boom_params: Seq[DCacheParams] = N
     memStart = site(ExtMem).get.master.base,
     numGens = params.size)
   }
-  case MaxHartIdBits => if (params.size + boom_params.size == 1) 1
-      else log2Ceil(params.size + boom_params.size)
+  case MaxHartIdBits => log2Ceil(params.size + up(TraceGenKey, site).length) max 1
 })
+
 
 class TraceGenConfig extends Config(
   new WithTraceGen(List.fill(2) { DCacheParams(nMSHRs = 0, nSets = 16, nWays = 2) }) ++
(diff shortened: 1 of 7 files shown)


generators/tracegen/src/main/scala/Configs.scala
```diff
@@ -10,7 +10,7 @@ import freechips.rocketchip.rocket.DCacheParams
 import freechips.rocketchip.tile.{MaxHartIdBits, XLen}
 import scala.math.{max, min}
 
-class WithTraceGen(params: Seq[DCacheParams], nReqs: Int = 8192)
```
zhemao: This really should be in a separate config mixin.

generators/tracegen/src/main/scala/Tile.scala
```diff
@@ -29,6 +34,180 @@ class TraceGenTile(val id: Int, val params: TraceGenParams, q: Parameters)
   override lazy val module = new TraceGenTileModuleImp(this)
 }
 
+class BoomTraceGenTile(val id: Int, val params: TraceGenParams, q: Parameters)
+  extends BaseTile(params, SynchronousCrossing(), HartsWontDeduplicate(params), q) {
+  val boom_params = p.alterMap(Map(TileKey -> BoomTileParams(
+    dcache=params.dcache,
+    core=BoomCoreParams(nPMPs=0, numLdqEntries=32, numStqEntries=32, useVM=false))))
+  val dcache = params.dcache.map {
```
zhemao: Why use map here if you're just going to call get and don't use the actual config? Just create the lazy module unconditionally.

generators/tracegen/src/main/scala/Tile.scala
```diff
@@ -29,6 +34,180 @@ class TraceGenTile(val id: Int, val params: TraceGenParams, q: Parameters)
...
+
+  val ptw = Module(new DummyPTW(1))
+
+  val lsu = Module(new LSU()(outer.boom_params, outer.dcache.module.edge))
+  lsu.io.core.tsc_reg := 0.U(1.W)
+  ptw.io.requestors.head <> lsu.io.ptw
+  outer.dcache.module.io.lsu <> lsu.io.dmem
+
+
+  val rob_sz = outer.boom_params(TileKey).core.asInstanceOf[BoomCoreParams].numRobEntries
```
zhemao: I think it would be better to put this glue logic into a separate module that just gets instantiated in the TileModuleImp.

generators/tracegen/src/main/scala/Configs.scala
jerryz123: It might be useful to tracegen a boom tile and a rocket tile together. Having 1 mixin for both types enables this.

zhemao: You can still use both with separate config mixins.

Though you will have to change the part where you instantiate them to check whether or not the parameter is defined or not.

jerryz123:
I split up the shim logic into its own module, and split up the mixins for instantiating the BOOM tracegen.

generators/tracegen/src/main/scala/Tile.scala
```diff
@@ -29,6 +34,190 @@ class TraceGenTile(val id: Int, val params: TraceGenParams, q: Parameters)
...
+  tracegen_uop.ctrl.is_load := isRead(io.tracegen.req.bits.cmd) && !isWrite(io.tracegen.req.bits.cmd)
+  tracegen_uop.ctrl.is_sta  := isWrite(io.tracegen.req.bits.cmd)
+  tracegen_uop.ctrl.is_std  := isWrite(io.tracegen.req.bits.cmd)
+
+  io.lsu.dis_uops(0).valid         := io.tracegen.req.fire()
+  io.lsu.dis_uops(0).bits          := tracegen_uop
+  
+  when (io.tracegen.req.fire()) {
+    rob_tail := WrapInc(rob_tail, rob_sz)
+    rob_bsy(rob_tail)   := true.B
```
zhemao: Aren't rob_head and rob_tail backwards here? Entries are placed into a circular buffer at the head and taken out at the tail.

generators/tracegen/src/main/scala/Tile.scala
```diff
@@ -29,6 +34,190 @@ class TraceGenTile(val id: Int, val params: TraceGenParams, q: Parameters)
...
+
+  def WrapInc(idx: UInt, max: Int): UInt = {
+    Mux(idx === (max-1).U, 0.U, idx + 1.U)
+  }
+
+
+  io.tracegen.req.ready := (!rob_bsy(rob_tail) &&
+    !rob_wait_till_empty &&
+    (ready_for_amo || !(isAMO(io.tracegen.req.bits.cmd) || io.tracegen.req.bits.cmd === M_XLR || io.tracegen.req.bits.cmd === M_XSC)) &&
+    (WrapInc(rob_tail, rob_sz) =/= rob_head) &&
```
zhemao: This always leaves one entry in the ROB empty. Take a look at Chisel's queue implementation to see how you can detect full queues properly.

generators/tracegen/src/main/scala/Tile.scala
jerryz123: This follow's BOOM's ROB naming scheme.

zhemao:
Take a look at this part of the Chisel codebase to see how to detect queue empty/full correctly.

https://github.com/freechipsproject/chisel3/blob/master/src/main/scala/chisel3/util/Decoupled.scala#L207

PR #364, opened 2019-12-12
Pull in IceNet checksum offload changes

zhemao:


generators/example/src/main/scala/Top.scala
```diff
@@ -101,3 +103,12 @@ class TopWithInitZero(implicit p: Parameters) extends Top
...
 // DOC include end: TopWithInitZero
+
+class TopWithIceNIC(implicit p: Parameters) extends Top
+    with HasPeripheryIceNIC {
+  override lazy val module = new TopWithIceNICModule(this)
+}
+
+class TopWithIceNICModule(outer: TopWithIceNIC)
+  extends TopModule(outer)
+  with HasPeripheryIceNICModuleImp
```
jerryz123: This should be refactored as `CanHavePeripheryIceNICModuleImp`, but I guess I can do that in the future.

zhemao: The BlockDevice doesn't use CanHave* either currently. We should just do all of them in one go later on.

zhemao:
@a0u Just tested it and the existing icenet-driver still works with these changes.

PR #369, opened 2019-12-13
[tools] Build multilib gnu-toolchain by default.

jerryz123:


scripts/build-toolchains.sh
```diff
@@ -110,7 +110,7 @@ else
         esac; ) || die 'obsolete make version; need GNU make 4.x or later'
 
     module_prepare riscv-gnu-toolchain qemu
-    module_build riscv-gnu-toolchain --prefix="${RISCV}"
+    module_build riscv-gnu-toolchain --prefix="${RISCV} --enable-multilib"
```
a0u: `--enable-multilib` should be outside of the quoted string, hence the CI failure.

jerryz123: Clearly I did not do any testing :)

a0u:
If this is not particularly urgent, I sort of prefer to fold this change into my `syscalls.c` refactoring on https://github.com/ucb-bar/chipyard/tree/bsp.  Since that requires switching libraries to the `medany` code model, it would be simpler to update and test all the toolchain changes at the same time.  multilib also complicates the build/install procedure for the custom libgloss, which I will have to figure out.

jerryz123:
No it is not urgent, I'll close this then.

PR #379, opened 2020-01-16
Update LICENSE

abejgonzalez:
Update the BSD license so that it is detected by GitHub.


LICENSE
```diff
@@ -1,27 +1,29 @@
+BSD 3-Clause License
+
 Copyright (c) 2017-2019, The Regents of the University of California (Regents).
```
a0u: We may as well update the copyright year to 2020.

PR #384, opened 2020-01-18
don't check prebuilt tools are on master

colinschmidt:
Backport from #362 
h/t @jerryz123


jerryz123:
Just remember to merge this back into dev when we release

PR #386, opened 2020-01-21
Patch parallel make

### LINKED ISSUE #377: Most recently changed print statements aren't included when compiling
This seems to be the same issue as #319, but as there was no information provided about replicating and no details as to why it was closed shortly after opening, I figured a new issue would be better.

I also tried creating a new generator and noticed that printf outputs weren't showing up when I expected, but as I added more printf calls, earlier calls that hadn't shown up before began to show up. I somewhat replicated this on master using the pwm example generator, though it seems like something is kind of nondeterministic.

I first changed the `PWMBase` class to print the PWM output:

```
  // If PWM is enabled, pwmout is high when counter < duty                      
  // If PWM is not enabled, it will always be low                               
  val pwmval = Wire(Bool())                                                     
  pwmval := io.enable && (counter < io.duty)                                    
  printf(p"pwmval: $pwmval \n")                                        
```

After running a clean build (`make clean; make CONFIG=PWMRocketConfig TOP=TopWithPWMTL -j32`), running the test with `./simulator-example-PWMRocketConfig +verbose ../../tests/pwm.riscv` did not show the expected printf output (including or not including `+verbose` did not make a difference either way). However, when I went through my steps one more time to write up this issue, it did show up, and I don't believe there was any difference.

Adding another printf beneath the first call (I just used `printf(p"another print\n")`) and recompiling resulted in the `pwmval:` output displaying, but the `another print` line not being shown.

Similarly, adding a third printf beneath the second line and recompiling resulted in the first two printfs appearing in the output. (In the redo, it did not show up, but adding a fourth printf caused the second and third message to output).

Commenting out or removing lines suffers from the same issue - the next compilation still includes the printed output from the line that should no longer have any effect. 

Is this an issue with the build system for chipyard? When I add and remove print statements in a non-chipyard Chisel project, the change is immediately reflected when running tests.

abejgonzalez:
Occasionally, parallel `make` will silently ignore source file changes. See #377 and https://stackoverflow.com/questions/19822435/multiple-targets-from-one-recipe-and-parallel-execution/. This attempts to address this issue by adding a `touch` to the targets depending on the `INTERMEDIATE` value.

This was tested by doing `make -j32` on the config./system listed in #377. 

Fixes #377

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -40,7 +40,7 @@ $(sim_files): $(call lookup_scala_srcs,$(base_dir)/generators/utilities/src/main
 #########################################################################################
 .INTERMEDIATE: generator_temp
 $(FIRRTL_FILE) $(ANNO_FILE): generator_temp
-	touch $@
+	@echo "" > /dev/null
 
 generator_temp: $(SCALA_SOURCES) $(sim_files)
 	mkdir -p $(build_dir)
@@ -62,7 +62,7 @@ HARNESS_TARGETS = $(HARNESS_FILE) $(HARNESS_SMEMS_CONF) $(HARNESS_ANNO) $(HARNES
 # NOTE: These *_temp intermediate targets will get removed in favor of make 4.3 grouped targets (&: operator)
 .INTERMEDIATE: firrtl_temp
 $(TOP_TARGETS) $(HARNESS_TARGETS): firrtl_temp
-	touch $@
+	@echo "" > /dev/null
 
 firrtl_temp: $(FIRRTL_FILE) $(ANNO_FILE)
 	cd $(base_dir) && $(SBT) "project tapeout" "runMain barstools.tapeout.transforms.GenerateTopAndHarness -o $(TOP_FILE) -tho $(HARNESS_FILE) -i $(FIRRTL_FILE) --syn-top $(TOP) --harness-top $(VLOG_MODEL) -faf $(ANNO_FILE) -tsaof $(TOP_ANNO) -tdf $(sim_top_blackboxes) -tsf $(TOP_FIR) -thaof $(HARNESS_ANNO) -hdf $(sim_harness_blackboxes) -thf $(HARNESS_FIR) $(REPL_SEQ_MEM) $(HARNESS_CONF_FLAGS) -td $(build_dir)" && touch $(sim_top_blackboxes) $(sim_harness_blackboxes)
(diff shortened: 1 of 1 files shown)


abejgonzalez:
Note: `make` 4.3 now supports the group target operator (`&:`) that is much more elegant. So when we enforce `make >= 4.3` we can update this. 

https://www.gnu.org/software/make/manual/html_node/Multiple-Targets.html

common.mk
```diff
@@ -57,8 +59,11 @@ TOP_TARGETS = $(TOP_FILE) $(TOP_SMEMS_CONF) $(TOP_ANNO) $(TOP_FIR) $(sim_top_bla
 HARNESS_TARGETS = $(HARNESS_FILE) $(HARNESS_SMEMS_CONF) $(HARNESS_ANNO) $(HARNESS_FIR) $(sim_harness_blackboxes)
 
 # DOC include start: FirrtlCompiler
+# NOTE: These *_temp intermediate targets will get removed in favor of make 4.3 grouped targets (&: operator)
 .INTERMEDIATE: firrtl_temp
 $(TOP_TARGETS) $(HARNESS_TARGETS): firrtl_temp
+	touch $@
```
jwright6323: I think it's confusing to use `.INTERMEDIATE` when there isn't a file being created (`firrtl_temp`). `.INTERMEDIATE` here is telling make that you'll generate a file called `firrtl_temp` that it can then delete. Are you sure this wasn't the problem? Let's discuss offline.

jwright6323: Just to summarize the offline discussion: this `touch $@` solves the issue because any commands in this rule do so. We'll replace it with something that has fewer undesirable side effects (mangling timestamps is not great) like an `echo`.

abejgonzalez: Offline result: We tested things a bit more. Specifically, we added a `touch firrtl_temp` equivalent and noticed that it didn't make a difference. However, any line added under `$(TOP_TARGETS) $(HARNESS_TARGETS): firrtl_temp` works (just needs a rule body). So I will change that to be something that doesn't affect the files (i.e. an echo or something similar. 

In general, this is a very confusing error to deal with. TL:DR `make` is busted wait for 4.3.

colinschmidt:
Since make 4.3 is so new we should probably wait a while, probably until its in most package managers. If we ask our users to build lots of software from scratch they will get annoyed.

abejgonzalez:
> Since make 4.3 is so new we should probably wait a while, probably until its in most package managers. If we ask our users to build lots of software from scratch they will get annoyed.

I agree. This PR should fix the issue in the meantime.

PR #387, opened 2020-01-21
System Requirements

alonamid:
This tries to make the system requirements a bit more clear, but providing two scripts for common Linux platforms (Centos/Ubuntu).


scripts/centos-req.sh
```diff
@@ -0,0 +1,34 @@
...
+sudo yum install -y centos-release-scl
+sudo yum install -y devtoolset-8-make
+
+# install DTC
+DTCversion=dtc-1.4.4
+wget https://git.kernel.org/pub/scm/utils/dtc/dtc.git/snapshot/$DTCversion.tar.gz
+tar -xvf $DTCversion.tar.gz
+cd $DTCversion
+make -j16
+make install
```
a0u: CentOS Extras provides `dtc` (1.4.6 with CentOS 7).
Otherwise, sudo is needed to install to the `/usr/local` prefix.
```suggestion
sudo make install
```

scripts/centos-req.sh
```diff
@@ -0,0 +1,34 @@
...
+cd ..
+rm -rf $DTCversion.tar.gz
+rm -rf $DTCversion
+
+# install verilator
+git clone http://git.veripool.org/git/verilator
+cd verilator/
+git checkout v4.002
+autoconf && ./configure && make -j16 && sudo make install
```
a0u: EPEL provides `verilator` (3.922 with CentOS 7).

scripts/ubuntu-req.sh
```diff
@@ -0,0 +1,36 @@
...
+# Install GNU make 4.x (needed to cross-compile glibc 2.28+)
+sudo apt-get install build-essential
+
+# install DTC
+DTCversion=dtc-1.4.4
+wget https://git.kernel.org/pub/scm/utils/dtc/dtc.git/snapshot/$DTCversion.tar.gz
+tar -xvf $DTCversion.tar.gz
+cd $DTCversion
+make -j16
+make install
```
a0u: `device-tree-compiler` is an official package (>= 1.4.0 as of Xenial 16.04 LTS).

scripts/ubuntu-req.sh
```diff
@@ -0,0 +1,36 @@
...
+cd ..
+rm -rf $DTCversion.tar.gz
+rm -rf $DTCversion
+
+# install verilator
+git clone http://git.veripool.org/git/verilator
+cd verilator/
+git checkout v4.002
+autoconf && ./configure && make -j16 && sudo make install
```
a0u: `verilator` is an official package (3.874 with 16.04 LTS and 3.916 with 18.04 LTS).

scripts/ubuntu-req.sh
```diff
@@ -0,0 +1,36 @@
...
+echo "deb https://dl.bintray.com/sbt/debian /" | sudo tee -a /etc/apt/sources.list.d/sbt.list
+curl -sL "https://keyserver.ubuntu.com/pks/lookup?op=get&search=0x2EE0EA64E40A89B84B2DF73499E82A75642AC823" | sudo apt-key add
+sudo apt-get update
+sudo apt-get install -y sbt
+sudo apt-get install -y texinfo gengetopt
+sudo apt-get install -y libxpat1-dev libusb-dev libncurses5-dev cmake
+# deps for poky
+sudo apt-get install -y python3.6 patch diffstat texi2html texinfo subversion chrpath git wget
+# deps for qemu
+sudo apt-get install libgtk-3-dev
```
a0u: Be consistent with `sudo apt-get install -y`.

docs/Chipyard-Basics/Initial-Repo-Setup.rst
```diff
@@ -10,6 +10,19 @@ Chipyard is developed and tested on Linux-based systems.
 
 .. Warning:: Working under Windows is not recommended.
 
+
+In Centos-based platforms, we recommend installing the following dependencies:
```
a0u: Nitpick:
```suggestion
In CentOS-based platforms, we recommend installing the following dependencies:
```

a0u:
Also, if I recall correctly, [verilator.mk](../blob/master/sims/verilator/verilator.mk) should download and build its own Verilator, so installing it system-wide shouldn't be needed.

abejgonzalez:
> Also, if I recall correctly, [verilator.mk](../blob/master/sims/verilator/verilator.mk) should download and build its own Verilator, so installing it system-wide shouldn't be needed.

Yes, that is true. Verilator is installed separately.

docs/Chipyard-Basics/Initial-Repo-Setup.rst
```diff
@@ -10,6 +10,19 @@ Chipyard is developed and tested on Linux-based systems.
...
+
+.. include:: /../scripts/centos-req.sh
+   :code: bash
+
+In Ubuntu/Debian-based platforms (Ubuntu), we recommend installing the following dependencies:
+
+.. include:: /../scripts/ubuntu-req.sh
+   :code: bash
+
+Note that when running on an Amazon Web Services EC2 FPGA-development instance (for FireSim), FireSim includes a machine setup script that will install all of the aforementioned dependencies (and some additional ones).
```
abejgonzalez: Can we change this to be a formatted readthedocs note: `.. Note::`

scripts/ubuntu-req.sh
abejgonzalez: As mentioned in the main comment chain, we should just remove verilator from both since verilator is installed separately in the makefiles.

abejgonzalez:
Also if you can, on future commits can you add the `[ci skip]` to the commit so that the tests don't run on it.

scripts/centos-req.sh
```diff
@@ -0,0 +1,18 @@
...
+sudo yum install -y gmp-devel mpfr-devel libmpc-devel zlib-devel vim git java java-devel
+curl https://bintray.com/sbt/rpm/rpm | sudo tee /etc/yum.repos.d/bintray-sbt-rpm.repo
+sudo yum install -y sbt texinfo gengetopt
+sudo yum install -y expat-devel libusb1-devel ncurses-devel cmake "perl(ExtUtils::MakeMaker)"
+# deps for poky
+sudo yum install -y python36 patch diffstat texi2html texinfo subversion chrpath git wget
+# deps for qemu
+sudo yum install -y gtk3-devel
+# deps for firemarshal
+sudo yum install -y python36-pip python36-devel rsync
```
a0u: Not all of the latest dependencies for FireMarshal are listed here.
The authoritative list lives at https://github.com/firesim/FireMarshal/blob/master/centos-requirements.txt (which needs a Debian/Ubuntu equivalent).

@NathanTP: Any preferences on how to manage this?  Some options:
* Remove them and let the user follow the FireMarshal README.
* Inline `centos-dependencies.txt` here.
* Read `centos-dependencies.txt` with a shell redirect (but the FireMarshal submodule is not guaranteed to be initialized).

scripts/centos-req.sh
```diff
@@ -0,0 +1,18 @@
...
+
+sudo yum groupinstall -y "Development tools"
+sudo yum install -y gmp-devel mpfr-devel libmpc-devel zlib-devel vim git java java-devel
+curl https://bintray.com/sbt/rpm/rpm | sudo tee /etc/yum.repos.d/bintray-sbt-rpm.repo
+sudo yum install -y sbt texinfo gengetopt
+sudo yum install -y expat-devel libusb1-devel ncurses-devel cmake "perl(ExtUtils::MakeMaker)"
+# deps for poky
+sudo yum install -y python36 patch diffstat texi2html texinfo subversion chrpath git wget
+# deps for qemu
+sudo yum install -y gtk3-devel
```
a0u: Just to propose a (non-essential) alternative: `gtk3-devel` is overkill for a headless system if not using the GUI.
The minimum dependencies for QEMU are:
```suggestion
sudo yum install -y glib2-devel libfdt-devel pixman-devel zlib-devel
```
Source: https://wiki.qemu.org/Hosts/Linux

scripts/ubuntu-req.sh
```diff
@@ -0,0 +1,19 @@
...
+echo "deb https://dl.bintray.com/sbt/debian /" | sudo tee -a /etc/apt/sources.list.d/sbt.list
+curl -sL "https://keyserver.ubuntu.com/pks/lookup?op=get&search=0x2EE0EA64E40A89B84B2DF73499E82A75642AC823" | sudo apt-key add
+sudo apt-get update
+sudo apt-get install -y sbt
+sudo apt-get install -y texinfo gengetopt
+sudo apt-get install -y libxpat1-dev libusb-dev libncurses5-dev cmake
+# deps for poky
+sudo apt-get install -y python3.6 patch diffstat texi2html texinfo subversion chrpath git wget
+# deps for qemu
+sudo apt-get install -y libgtk-3-dev
```
a0u: See above.
```suggestion
sudo apt-get install -y libglib2.0-dev libfdt-dev libpixman-1-dev zlib1g-dev
```

scripts/centos-req.sh
alonamid: In in favor of inlining (these are recommended system requirements. It shouldn't be a complicated script)

a0u: I would also prefer inlining, but someone has to remember to synchronize the lists once in a while.

alonamid: synchronize when someone opens an issue that the requirements don't hold anymore :)

a0u:
This was tested on the official CentOS 7 and Ubuntu 18.04 AMIs: at least the toolchain build, Verilator sim of the default DUT, and FireMarshal br-base image build.

scripts/centos-req.sh
NathanTP: Historically, we've relied on FireSim to handle this:
https://github.com/firesim/firesim/blob/d801f0ffa7047ef2516ab752f2e2443f322cdaa0/build-setup-nolog.sh#L147

I would prefer not to duplicate this list. It's already enough of a pain to keep everything in sync. If we're going to do this in chipyard, I could put a centos-req.sh in FireMarshal and call that from chipyard and/or firesim to minimize duplicate logic.

As for debian equivalents; I have no interest in writing or maintaining that list. I don't have an ubuntu environment (that I manage) and the setup is more complicated. If someone contributes such a script, I'll happily include it. Even then, we'd need to have an official ubuntu version that we support since various hacks and workarounds will depend on the precise version.

alonamid: I'm not try to do anything robust or perfect. This is a need that came from users (more than one). I believe this PR is better than the current state. If it will get out of sync, we will fix it. I'm not assuming people will relay on this, but if they have a "bare" system, this is better than getting obscure message during the build process. They might still end up getting an obscure build message during the build, but less often.

alonamid: (this script does not get called automatically anywhere. It's just references in the docs)

scripts/centos-req.sh
```diff
@@ -0,0 +1,18 @@
...
+# deps for firemarshal
+sudo yum install -y python36-pip python36-devel rsync libguestfs-tools makeinfo yac bison flex expat glib-devel ctags
```
a0u: Not your fault, but `yac` should be `yacc` - this is a typo in `centos-requirements.txt` that I just noticed.
```suggestion
sudo yum install -y python36-pip python36-devel rsync libguestfs-tools makeinfo expat ctags
```
Some packages (yacc, bison, flex) are already installed with `Development tools`.

PR #388, opened 2020-01-22
Clone firesim non-recursively

a0u:
This is a draft PR to demonstrate the top-level fix for the excessive riscv-linux cloning issue.

Once firesim is bumped with the midas flattening, we don't need to clone any firesim submodules except for FireMarshal.
Deprecate the `--no-firesim` switch (which did not work correctly anyway).

### DIFF
--- a/scripts/init-submodules-no-riscv-tools.sh
+++ b/scripts/init-submodules-no-riscv-tools.sh
@@ -6,25 +6,6 @@ set -o pipefail
 
 RDIR=$(git rev-parse --show-toplevel)
 
-_usage() {
-  echo "usage: ${0} [--no-firesim]" >&2
-  exit 1
-}
-
-NO_FIRESIM=false
-while getopts 'h-:' opt ; do
-  case ${opt} in
-  -)
-    case ${OPTARG} in
-    no-firesim) NO_FIRESIM=true ;;
-    *) echo "invalid option: --${OPTARG}" >&2 ; _usage ;;
-    esac ;;
-  h) _usage ;;
-  *) echo "invalid option: -${opt}" >&2 ; _usage ;;
-  esac
-done
-shift $((OPTIND - 1))
-
 # Ignore toolchain submodules
 cd "$RDIR"
 for name in toolchains/*-tools/*/ ; do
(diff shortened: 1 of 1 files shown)


scripts/init-submodules-no-riscv-tools.sh
```diff
@@ -59,13 +40,6 @@ git config --unset submodule.generators/sha3.update
...
 git submodule update --init generators/sha3
 
-if [ $NO_FIRESIM = false ]; then
-  echo "initializing firesim"
-  # Renable firesim and init only the required submodules to provide
-  # all required scala deps, without doing a full build-setup
-  git config --unset submodule.sims/firesim.update
-  git submodule update --init sims/firesim
-  git -C sims/firesim submodule update --init sim/midas
-  git -C sims/firesim submodule update --init --recursive sw/firesim-software
```
NathanTP: Until I move firemarshal to chipyard, we'll need to shallow clone firesim-software as well (it's symlinked into chipyard/sw/firemarshal). If this is targeting the Friday release, you'll need to add the shallow clone. If this is scheduled for after the release, you may as well wait for me to move Marshal.

a0u: Does anything at the Chipyard top level mandate FireMarshal?  If someone wants to use FireMarshal, s/he would need to manually run its init-submodules.sh script, so it isn't an onerous expectation to have to initialize FireMarshal explicitly beforehand.

a0u: (Or rely on the firesim-setup.sh script to do so.)

scripts/init-submodules-no-riscv-tools.sh
```diff
@@ -59,13 +40,7 @@ git config --unset submodule.generators/sha3.update
...
-  git config submodule.sims/firesim.update none
-fi
+git config --unset submodule.sims/firesim.update
+# Non-recursive clone for minimal dependencies (midas)
+git submodule update --init sims/firesim
+git -C sims/firesim submodule update --init sw/firesim-software
```
a0u: Per offline discussion with @NathanTP, the upcoming release will still clone FireMarshal by default to avoid breaking user expectations.

When FireMarshal migrates to Chipyard in the subsequent release, it would make sense to initialize FireMarshal along with the toolchain build, perhaps renaming `build-toolchains.sh` to `setup-sw.sh` and this script to `setup-hw.sh`.

scripts/init-submodules-no-riscv-tools.sh
```diff
@@ -59,13 +40,7 @@ git config --unset submodule.generators/sha3.update
...
 
-if [ $NO_FIRESIM = false ]; then
-  echo "initializing firesim"
-  # Renable firesim and init only the required submodules to provide
-  # all required scala deps, without doing a full build-setup
-  git config --unset submodule.sims/firesim.update
-  git submodule update --init sims/firesim
-  git -C sims/firesim submodule update --init sim/midas
-  git -C sims/firesim submodule update --init --recursive sw/firesim-software
-  git config submodule.sims/firesim.update none
```
davidbiancolin: I think my preference would be re-disable the the firesim submodule if that's alright with you Albert.

sagark: +1

a0u: Done.

a0u:
This also ensures that dramsim2 is still initialized for CI, which subsumes #401.

PR #390, opened 2020-01-22
IOBinders and BuildTop unification between FireChip and Chipyard

jerryz123:
Large set of changes towards generator unification

- CI job naming cleaned up
- `tracegen` now uses the Chipyard harness for tests and configs. The tracegen `SUB_PROJECT` is removed
- Firechip CI now installs `libdwarf` and `libelf` to fix broken `dev` branch CI (thanks @sagark)
- `example` package now renamed to `chipyard`
  - `chipyard.example` contains `GCD, NodeTypes, InitZero` examples
  - `chipyard.iobinders` contains IOBinders stuff
- `IOBinders` config key which maps traits to partial functions.
  - `chipyard` designs use this to specify which `connectSimX` methods to call
  - `firechip` designs use this to specify which bridges to instantiate
- Unification between FireSIm configs/tops and Chipyard configs/tops
  - `BuildTop` used in both Chipyard and Firechip to specify custom top
  - `FireSimDUT` is now just `chipyard.Top`
- TraceIO stuff (for FirePerf) now moved into `testchipip`
- Documentation updated to reflect new naming and IOBinders
- All configs use L2TLBs to match Firesim


.circleci/config.yml
```diff
@@ -188,44 +188,44 @@ jobs:
                 key: extra-tests-{{ .Branch }}-{{ .Revision }}
                 paths:
                     - "/home/riscvuser/project/tests"
-    prepare-example:
+    prepare-chipyard-rocket:
         executor: main-env
```
jerryz123: This was cleaning up discrepencies in naming in this file

docs/Advanced-Concepts/Chip-Communication.rst
```diff
@@ -109,16 +109,13 @@ reminder, to run a software RTL simulation, run:
 
 FireSim FPGA-accelerated simulations use TSI by default as well.
 
-If you would like to build and simulate a Chipyard configuration with a DTM configured for DMI communication, then you must create a
-top-level system with the DTM (``TopWithDTM``), a test-harness to connect to the DTM (``TestHarnessWithDTM``), as well as a config to use that top-level system.
+If you would like to build and simulate a Chipyard configuration with a DTM configured for DMI communication, then you must tie-off the TSI interface, and instantiate the `SimDTM`. Note that we use `WithTiedOffSerial ++ WithSimDTM` instead of `WithTiedOffDebug ++ WithSimSerial`.
```
alonamid: Is there a reason that `WithSimDTM` and `WithTiedOffDebug` are not symmetric? (i.e. `WithSimDTM` & `WithTiedOffDTM`, or `WithSimDebug` & `WithTiedOffDebug`)

jerryz123: Good point, I think renaming it to `WithTiedOffDTM` makes sense.

docs/Chipyard-Basics/Configs-Parameters-Mixins.rst
```diff
@@ -74,52 +74,37 @@ Cake Pattern
...
-In the ``MySoC`` example class, the "outer" ``MySoC`` instantiates the "inner"
-``MySoCModuleImp`` as a lazy module implementation. This delays immediate elaboration
+In the ``Top`` example class, the "outer" ``Top`` instantiates the "inner"
+``TopModule`` as a lazy module implementation. This delays immediate elaboration
 of the module until all logical connections are determined and all configuration information is exchanged.
-The ``RocketSubsystem`` outer base class, as well as the
-``HasPeripheryX`` outer traits contain code to perform high-level logical
-connections. For example, the ``HasPeripherySerial`` outer trait contains code
-to lazily instantiate the ``SerialAdapter``, and connect the ``SerialAdapter``'s
+The ``Syatem`` outer base class, as well as the
```
alonamid:
```suggestion
The ``System`` outer base class, as well as the
```

docs/Customization/Keys-Traits-Configs.rst
```diff
@@ -42,64 +42,37 @@ Top-level traits should be defined and documented in subprojects, alongside thei
...
 We can use this mixin when composing our configs.
 
-.. literalinclude:: ../../generators/example/src/main/scala/RocketConfigs.scala
+.. literalinclude:: ../../generators/chipyard/src/main/scala/RocketConfigs.scala
     :language: scala
     :start-after: DOC include start: GCDTLRocketConfig
     :end-before: DOC include end: GCDTLRocketConfig
 
-
-BuildTop
```
alonamid: `BuildTop` doesn't exist anymore?

jerryz123: The `BuildTop` composition via recursive functions (which this section describes) does not exist anymore.

generators/chipyard/src/main/scala/ConfigMixins.scala
```diff
@@ -35,117 +35,40 @@ import ConfigValName._
...
- */
 class WithUART extends Config((site, here, up) => {
   case PeripheryUARTKey => Seq(
     UARTParams(address = 0x54000000L, nTxEntries = 256, nRxEntries = 256))
 })
 
-/**
- * Mixin to remove any GPIOs
- */
 class WithNoGPIO extends Config((site, here, up) => {
```
alonamid: Can we explain somewhere in the docs the reason for the `WithNoGPIO` (as opposed to everything else which is an additive). I know the reason, I just think it'll be nice to explain it in the docs as well

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -0,0 +1,167 @@
...
+import sifive.blocks.devices.gpio._
+import sifive.blocks.devices.uart._
+
+import testchipip._
+import icenet._
+import tracegen.{HasTraceGenTilesModuleImp}
+
+import scala.reflect.{ClassTag, classTag}
+
+// System for instantiating binders based
```
alonamid: Should probably have a version of this in the readable docs as well?

alonamid:
The FireSim part of the docs (which currently describes the limitations of the split configs (https://chipyard.readthedocs.io/en/dev/Simulation/FPGA-Accelerated-Simulation.html#current-limitations) , should be replaced with something that explains the relationship between IOBinders and BridgeBinders, and how to use the configs in the FireSim context

generators/firechip/src/main/scala/TargetConfigs.scala
```diff
@@ -112,235 +101,69 @@ class L2SingleBank512K extends freechips.rocketchip.subsystem.WithInclusiveCache
...
-// octa core config
-class FireSimRocketChipOctaCoreConfig extends Config(
-  new WithNDuplicatedRocketCores(8) ++
-  new FireSimRocketChipSingleCoreConfig)
+//*****************************************************************
+// Rocket configs, base off chipyard's RocketConfig
+//*****************************************************************
+class FireSimRocketConfig extends Config(
+  new WithDefaultFireSimBridges ++
+  new WithDefaultMemModel ++
```
alonamid: Why is this not part of `WithDefaultFireSimBridges`? What is the alternative? (i.e., what happens if this confix mix-in isn't added?)

generators/firechip/src/main/scala/TargetConfigs.scala
```diff
@@ -112,235 +101,69 @@ class L2SingleBank512K extends freechips.rocketchip.subsystem.WithInclusiveCache
...
-
-// octa core config
-class FireSimRocketChipOctaCoreConfig extends Config(
-  new WithNDuplicatedRocketCores(8) ++
-  new FireSimRocketChipSingleCoreConfig)
+//*****************************************************************
+// Rocket configs, base off chipyard's RocketConfig
+//*****************************************************************
+class FireSimRocketConfig extends Config(
+  new WithDefaultFireSimBridges ++
```
alonamid: This is the stuff we later want to break-off into a separate "platform/test-env/harness" config?

generators/firechip/src/main/scala/TargetConfigs.scala
jerryz123: `WithDefaultFireSimBridges` only contians stuff that manipulates the `IOBinders` key.

generators/firechip/src/main/scala/TargetConfigs.scala
jerryz123: I believe so.

generators/tracegen/src/main/scala/Configs.scala
```diff
@@ -60,19 +60,6 @@ class WithBoomTraceGen(params: Seq[DCacheParams], nReqs: Int = 8192)
...
-
-class NonBlockingTraceGenConfig extends Config(
-  new WithTraceGen(List.fill(2) { DCacheParams(nMSHRs = 2, nSets = 16, nWays = 2) }) ++
-  new BaseConfig)
-
-class BoomTraceGenConfig extends Config(
-  new WithBoomTraceGen(List.fill(2) { DCacheParams(nMSHRs = 8, nSets = 16, nWays = 2) }) ++
-  new BaseConfig)
-
 class WithL2TraceGen(params: Seq[DCacheParams], nReqs: Int = 8192)
```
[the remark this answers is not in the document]
jerryz123: The addition of the Configs into Chiypard was to enable `make CONFIG=TracegenConfig`. I want to leave the rest of the tracegen stuff there, to minimize the contents of `chipyard`

docs/Customization/Memory-Hierarchy.rst
```diff
@@ -62,22 +73,26 @@ Note that these configurations fully remove the L2 cache and mbus.
...
-        new freechips.rocketchip.subsystem.WithNSmallCores(1) ++
-        new freechips.rocketchip.system.BaseConfig)
-
-    class ScratchpadRocketConfig extends Config(
-        new freechips.rocketchip.subsystem.WithNoMemPort ++
-        new freechips.rocketchip.subsystem.WithNMemoryChannels(0) ++
-        new freechips.rocketchip.subsystem.WithNBanks(0) ++
-        new freechips.rocketchip.subsystem.WithScratchpadsOnly ++
-        new SmallRocketConfigNoL2)
+    class ScratchpadSmallRocketConfig extends Config(
```
davidbiancolin: Obligatory code-in-docs comment: This makes my eyes bleed.

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -0,0 +1,167 @@
...
+//
+// You could just as well create a custom harness module that instantiates
+// bridges explicitly, or add methods to
+// your target traits that instantiate the bridge there (i.e., akin to
+// SimAXI4Mem). Since cake traits live in Rocket Chip it was easiest to match
+// on the types rather than change trait code.
+
+
+
+// A map of partial functions that match on the type the DUT (_not_ it's
```
davidbiancolin: Update these comments. Feel free to remove the stuff that contrasts this against matching IO if you like.

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -0,0 +1,167 @@
...
+          case top: T => fn(clock, reset, success, top)
+          case _ => Nil
+        }
+      })
+  )
+})
+
+// This macro composes with previous matches on some Top mixin. This is useful for
+// annotation-like binders, since those can typically be composed
+class RegisterBinder[T](fn: => (Clock, Bool, Bool, T) => Seq[Any])(implicit tag: ClassTag[T]) extends Config((site, here, up) => {
```
davidbiancolin: I feel like this might need a better name. Really, what this does is compose with an existing IO binder.

.circleci/do-rtl-build.sh
```diff
@@ -32,7 +32,19 @@ run "cp -r ~/.sbt  $REMOTE_WORK_DIR"
 
 TOOLS_DIR=$REMOTE_RISCV_DIR
 LD_LIB_DIR=$REMOTE_RISCV_DIR/lib
-if [ $1 = "hwacha" ] || [ $1 = "gemmini" ]; then
+
```
abejgonzalez: Don't you want to build the tests after the config is built? Or am I missing something here?

.circleci/run-firesim-tests.sh
```diff
@@ -12,10 +12,19 @@ export FIRESIM_ENV_SOURCED=1
 
 SIMULATION_ARGS="${mapping[$1]}"
 
+cd $LOCAL_CHIPYARD_DIR/sims/firesim
```
abejgonzalez: Why is this repeated?

.circleci/do-rtl-build.sh
jerryz123: Why is that necessary?
If the tests fail to build, I would rather know sooner, rather than have to wait for the simulator to build.

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -0,0 +1,167 @@
...
+  case IOBinders => up(IOBinders, site) + (tag.runtimeClass.toString ->
+      ((clock: Clock, reset: Bool, success: Bool, t: Any) => {
+        t match {
+          case top: T => fn(clock, reset, success, top) ++
```
davidbiancolin: Unless i'm mistaken the the function associated with this Binder (lets call it _A_) will be invoked first, followed by the older one (_B_). This means that hardware generated by _A_ maybe overridden by hardware generated in _B_, which i'm not sure is the intuitive result.

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -0,0 +1,167 @@
...
+            up(IOBinders, site).getOrElse(tag.runtimeClass.toString, (c: Clock, r: Bool, s: Bool, t: Any) => Nil)(clock, reset, success, top)
```
davidbiancolin: We could make the Map just return this special Nil function as it's default.

docs/Advanced-Concepts/Chip-Communication.rst
abejgonzalez: Technically DTM is separate from DMI and JTAG. Do we want these mixins to state what's inside the system or what the interface is? Maybe we should name it something with DMI and JTAG like `WithSimDMI` `WithSimJTAG` if we are considering doing the latter.

.circleci/run-firesim-tests.sh
jerryz123: Because the libraries these generate are not cached in do-firesim-build.

I suppose the more correct solution would be to build these in the `build_tools` step, but that requires cloning all the firesim stuff in build_tools, which I judged to be more undesirable.

Blame @sagark for not trusting packaged libraries.

docs/Advanced-Concepts/Chip-Communication.rst
jerryz123: That doesn't work because `Debug.connectDebug` will connect either SimDTM or SimJTAG, depending on the interface of the debug module.
SimDTM and SimJTAG are the names of the actual widgets, so I don't think renaming this to SimDMI /SimJTAG is a good idea.

Given the behavior of `Debug.connectDebug`, I think `WithSimDTM` is the right name here.

generators/chipyard/src/main/scala/IOBinders.scala
jerryz123: What is the intuitive result when composing 2 Binders that attach to the same IO? My thought was that the `RegisterBinder` macro would result in composition, while the `RegisterIOBinder` macro would override.

docs/Advanced-Concepts/Chip-Communication.rst
abejgonzalez: Sorry. All `WithSimDTM` does is connect a `Debug` module in the `TestHarness` to the JTAG/DMI signals. I think this distinction should be mentioned in the docs. I don't think it is immediately clear that adding `WithSimDTM` connects both JTAG or DMI.

jerryz123: Ok, perhaps I'll name it `WithSimDebug` and `WithTiedOffDebug` then. That seems to be more accurate.

docs/Chipyard-Basics/Configs-Parameters-Mixins.rst
```diff
@@ -74,52 +74,37 @@ Cake Pattern
 A cake pattern is a Scala programming pattern, which enable "mixing" of multiple traits or interface definitions (sometimes referred to as dependency injection).
 It is used in the Rocket Chip SoC library and Chipyard framework in merging multiple system components and IO interfaces into a large system component.
 
-This example shows a Rocket Chip based SoC that merges multiple system components (BootROM, UART, etc) into a single top-level design.
+This example shows the Chipyard default top that composes multiple traits together into a fully-featured SoC with many optional components.
```
abejgonzalez: I think in the documentation we are mixing up "traits" with "mixins" again. Probably worth a second pass through the docs to make sure this is consistent.

docs/Chipyard-Basics/Configs-Parameters-Mixins.rst
```diff
@@ -74,52 +74,37 @@ Cake Pattern
...
-``MySoCModuleImp`` as a lazy module implementation. This delays immediate elaboration
+In the ``Top`` example class, the "outer" ``Top`` instantiates the "inner"
+``TopModule`` as a lazy module implementation. This delays immediate elaboration
 of the module until all logical connections are determined and all configuration information is exchanged.
-The ``RocketSubsystem`` outer base class, as well as the
-``HasPeripheryX`` outer traits contain code to perform high-level logical
-connections. For example, the ``HasPeripherySerial`` outer trait contains code
-to lazily instantiate the ``SerialAdapter``, and connect the ``SerialAdapter``'s
+The ``Syatem`` outer base class, as well as the
+``CanHavePeripheryX`` outer traits contain code to perform high-level logical
```
abejgonzalez: Nit: To make this clearer to read, maybe have the X separated out somehow i.e. `CanHavePeriphery<X>`

docs/Chipyard-Basics/Configs-Parameters-Mixins.rst
jerryz123: Is this not the correct terminology here?

docs/Customization/Memory-Hierarchy.rst
```diff
@@ -42,18 +60,11 @@ mixins for those too.
...
-        new WithUART ++
-        new freechips.rocketchip.subsystem.WithNoMMIOPort ++
-        new freechips.rocketchip.subsystem.WithNoSlavePort ++
-        new WithL1ICacheSets(128) ++                              // change rocket I$
-        new WithL1ICacheWays(2) ++                                // change rocket I$
-        new WithL1DCacheSets(128) ++                              // change rocket D$
-        new WithL1DCacheWays(2) ++                                // change rocket D$
-        new freechips.rocketchip.subsystem.WithNSmallCores(1) ++
-        new freechips.rocketchip.system.BaseConfig)
+         new freechips.rocketchip.subsystem.WithL1ICacheSets(128) ++  // change rocket I$
```
abejgonzalez: While the mixins shown are valuable... I don't think this config will work right out of the gate... maybe don't show this in config form... just show the mixins.

docs/Customization/Memory-Hierarchy.rst
```diff
@@ -158,15 +159,9 @@ number of DRAM channels is restricted to powers of two.
...
-        new WithTSI ++
-        new WithNoGPIO ++
-        new WithBootROM ++
-        new WithUART ++
-        new freechips.rocketchip.subsystem.WithNoMMIOPort ++
-        new freechips.rocketchip.subsystem.WithNoSlavePort ++
-        new WithNMemoryChannels(2) ++                          // multi-channel outer mem
-        new freechips.rocketchip.subsystem.WithNBigCores(1) ++
-        new freechips.rocketchip.system.BaseConfig)
+        new freechips.rocketchip.subsystem.WithNMemoryChannels(2) ++
```
abejgonzalez: Same as my above comment

docs/Customization/Memory-Hierarchy.rst
jerryz123: Why doesn't this config work?

generators/chipyard/src/main/scala/IOBinders.scala
jerryz123: How about `ComposeIOBinder`

generators/chipyard/src/main/scala/IOBinders.scala
davidbiancolin: I'd expect that anything you put in an LHS config instance should take priority over anything in an RHS config instance. Which is to say last connect semantics should favor connections made in an LHS config, which does not appear to be the case here (?).

jerryz123: Oh I see, yeah I'll switch the ordering then.

generators/chipyard/src/main/scala/IOBinders.scala
davidbiancolin: I think that's better. That's the verb you use in the comment anyways.

docs/Advanced-Concepts/Chip-Communication.rst
colinschmidt: I agree with `WithSimDebug`.

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -0,0 +1,167 @@
...
+import freechips.rocketchip.util._
+
+import sifive.blocks.devices.gpio._
+import sifive.blocks.devices.uart._
+
+import testchipip._
+import icenet._
+import tracegen.{HasTraceGenTilesModuleImp}
+
+import scala.reflect.{ClassTag, classTag}
```
colinschmidt: Fun naming

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -0,0 +1,167 @@
...
+
+// Since we also want to compose this structure like the existing config system,
+// use the scala string representation of the matched trait as a key
+
+case object IOBinders extends Field[Map[String, (Clock, Bool, Bool, Any) => Seq[Any]]](Map())
+
+
+// This macro overrides previous matches on some Top mixin. This is useful for
+// binders which modify IO, since those typically cannot be composed
+class RegisterIOBinder[T](fn: => (Clock, Bool, Bool, T) => Seq[Any])(implicit tag: ClassTag[T]) extends Config((site, here, up) => {
```
colinschmidt: So based on the below discussion this is for "non-idempotent IOBinders"?

colinschmidt: Maybe calling this `OverrideIOBinder` vs `ComposeIOBinder` below.

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -0,0 +1,167 @@
...
+  case IOBinders => up(IOBinders, site) + (tag.runtimeClass.toString ->
```
colinschmidt: Is this tag based on the class of the `IOBinder` itself, or the trait the user is matching on?

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -0,0 +1,167 @@
...
+    top.reset := r | top.debug.map { debug => AsyncResetReg(debug.ndreset) }.getOrElse(false.B)
+    Debug.connectDebug(top.debug, top.psd, c, r, dtm_success)(top.p)
+    when (dtm_success) { s := true.B }
+    Nil
+  }
+})
+
+
+class WithTraceGenSuccessBinder extends RegisterIOBinder({
+  (c, r, s, top: HasTraceGenTilesModuleImp) => s := top.success; Nil
```
colinschmidt: We need to be careful with last connect semantics here.
If the other success functions from `WithSimDTM` or `WithSimSerial` are before this then this direct assign (no when) will overwrite both of them.
I would make them all `when`s and also make sure there is no direct assign in `Top`.

generators/chipyard/src/main/scala/TestHarness.scala
```diff
@@ -0,0 +1,30 @@
...
+
+case object BuildTop extends Field[Parameters => Any]((p: Parameters) => Module(LazyModule(new Top()(p)).suggestName("top").module))
+
+class TestHarness(implicit val p: Parameters) extends Module {
+  val io = IO(new Bundle {
+    val success = Output(Bool())
+  })
+
+  val dut = p(BuildTop)(p)
+  io.success := false.B
```
colinschmidt: This appears to be the right thing w.r.t. last connect semantics.

docs/Chipyard-Basics/Configs-Parameters-Mixins.rst
abejgonzalez: Sorry. That is the correct terminology there. I was referring to all the docs in my comment. Sometimes we refer to "traits" as "mixins" and vice versa in the docs. So we should try to 1 (as @colinschmidt recommended) name them something else i.e. "CodeMixin" "ConfigMixin", or 2 just adjust the docs into traits and mixins as we intended originally.

generators/chipyard/src/main/scala/IOBinders.scala
jerryz123: The trait the user is matching on, which is determined from the function passed into this macro.

jerryz123:
> LGTM.
> Did you verify that you can overlay multiple binders correctly? I think we should also make some CI test that includes the overriding behavior.

The FireChip configs override the default behavior in `WithDefaultFireSimBridges`, and two of those configs are tested in our CI.

> I'd like to see both overriding binders (WithSimSerial ++ WithNoSimSerial ++ WithSimSerial)
and composing binders (not sure of which these are).

I couldn't find a practical example (at least within the stuff we already have) of composing binders being useful. Do you have a suggestion for a simple example of composing binders?

generators/chipyard/src/main/scala/IOBinders.scala
jerryz123: What do you mean no direct assign? As in wrap all assigns to `Top` IOs in `when`s?

colinschmidt: At least success which we know could be problematic.

generators/chipyard/src/main/scala/ConfigMixins.scala
jerryz123: There seems to be no good place to include this, since there aren't docs on GPIO anywhere. In any case this should get fixed eventually, so its not high priority.

docs/Customization/Memory-Hierarchy.rst
jerryz123: I decided to show just the mixins for brevity.

docs/Customization/Memory-Hierarchy.rst
jerryz123: I moved some of the configs in this section into RocketConfigs, and reduced the rest to just show the mixin

docs/Chipyard-Basics/Configs-Parameters-Mixins.rst
jerryz123: Seems like work for another PR

docs/Chipyard-Basics/Configs-Parameters-Mixins.rst
jerryz123: Done

generators/chipyard/src/main/scala/IOBinders.scala
jerryz123: Done

generators/chipyard/src/main/scala/IOBinders.scala
jerryz123: Done

docs/Simulation/FPGA-Accelerated-Simulation.rst
```diff
@@ -44,47 +44,19 @@ familiar with FireSim, please return to the `FireSim Docs
...
+    :language: scala
+    :start-after: DOC include start: firesimconfig
+    :end-before: DOC include end: firesimconfig
+
+
+Only 3 additional config-mixins are needed.
+
+* ``WithFireSimConfigTweaks`` modifies your design to better fit the FireSim usage model. For example, FireSim designs typically include a UART. Technically, adding this in is optional, but *strongly* recommended.
+* ``WithDefaultMemModel`` sets the external memory model in the FireSim simulation. See the FireSim documentation for details.
+* ``WithDefaultFireSimBridges`` sets the ``IOBinders`` key to use FireSim's Bridge system, which can drive target IOs with software bridge models running on the simulation host. See the FireSIm documnetation for details.
```
abejgonzalez:
```suggestion
* ``WithDefaultFireSimBridges`` sets the ``IOBinders`` key to use FireSim's Bridge system, which can drive target IOs with software bridge models running on the simulation host. See the FireSim documentation for details.
```

PR #401, opened 2020-01-24
Make sure midas dramsim2 gets initialized in CI

zhemao:


zhemao:
Don't need this because DRAMSim2 will be initialized in init-submodules-no-riscv-tools.sh from now on.

PR #412, opened 2020-01-27
[firechip] Enable trace by default in BOOM-based targets

davidbiancolin:
Since BOOM does not drive the trace port by default, TracerV was capturing an empty trace. 

**Impact**: rtl change

**Release Notes**
Enables trace capture by default in BOOM-based FireChip targets.

### DIFF
--- a/generators/firechip/src/main/scala/TargetConfigs.scala
+++ b/generators/firechip/src/main/scala/TargetConfigs.scala
@@ -80,6 +80,10 @@ class WithBoomL2TLBs(entries: Int) extends Config((site, here, up) => {
   ))
 })
 
+class WithBoomEnableTrace extends Config((site, here, up) => {
+  case BoomTilesKey => up(BoomTilesKey) map (tile => tile.copy(trace = true))
+})
+
 // Disables clock-gating; doesn't play nice with our FAME-1 pass
 class WithoutClockGating extends Config((site, here, up) => {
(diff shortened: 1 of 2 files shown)


jerryz123:
I'm a little hesitant to make this default behavior, since BOOM generates a ton of extra hardware to generate the trace. 

If we want to make this default, I would like to fixup the traceport's implementation in BOOM first.

davidbiancolin:
How much is a ton of extra hardware?

davidbiancolin:
If we choose not to make this default we should update the docs somewhere. Or even put a comment in the runtime_config.ini. @sagark thoughts?

davidbiancolin:
Also @alonamid.

jerryz123:
~2 kilobytes of flops. I noticed that enabling trace on a config with synth-asserts and aggressive ILA would frequently cause congestion issues.

I can fix the implementation so that it stores this extra state in SyncReadMems.

jerryz123:
@davidbiancolin This PR https://github.com/riscv-boom/riscv-boom/pull/435 should have a better implementation of the trace port. Can we merge that and bump boom to that?

davidbiancolin:
> @davidbiancolin This PR [riscv-boom/riscv-boom#435](https://github.com/riscv-boom/riscv-boom/pull/435) should have a better implementation of the trace port. Can we merge that and bump boom to that?

Yeah, let's do it.

PR #415, opened 2020-01-29
Marhsal in chipyard

NathanTP:
**Related issue**: 

**Impact**: software change

**Release Notes**
Moved FireMarshal submodule into chipyard properly (was a symlink to firesim before). This also adds FireMarshal to your $PATH in env.sh.


NathanTP:
see also:
firesim/firesim#482
firesim/firemarshal#127

scripts/build-toolchains.sh
```diff
@@ -141,5 +141,5 @@ cd "$RDIR"
 } > env-$TOOLCHAIN.sh
 
 # create general env.sh
-ln -sf env-$TOOLCHAIN.sh env.sh
+echo "source \$( realpath \$(dirname "\${BASH_SOURCE[0]}") )/env-$TOOLCHAIN.sh" >> env.sh
```
a0u: I'm not particularly enthused about introducing Bashisms in `env.sh` like `${BASH_SOURCE[0]}`, but it's probably not going to affect the vast majority of users.  As long as the toolchain `env-*.sh` is pure `sh(1)` syntax, I can work around this.

Issue #309, opened 2019-10-15
After running a test, the console output can't be seen unless `reset` is run


colinschmidt:
> Hmm. For some reason, after running a test, the console output can't be seen unless I `reset`. Im looking into it for now.

This is only a issue with the newer VCS version, if that helps

_Originally posted by @jerryz123 in https://github.com/ucb-bar/chipyard/pull/308#issuecomment-542326124_

jerryz123:
This seems like a Millennium environment issue, rather than a chipyard issue, but we can track it here

seldridge:
FWIW, this is pretty long-standing in Rocket Chip. I've noticed it and @ccelio has also experienced it.

Thanks for opening a tracking issue about it!

PR #417, opened 2020-01-30
[sim] Pipe /dev/null to simulators to fix VCS messing up stdout

jerryz123:
Thanks to @a0u for the fix.
This was messing up `tee`, causing spurious test failures with VCS.
This also fixes having to `reset` after running VCS simulations to get back `stdout`

**Note**: Verilator isn't affected, but do the same thing anyways to have uniform behavior across simulators.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -102,19 +102,19 @@ verilog: $(sim_vsrcs)
 #########################################################################################
 .PHONY: run-binary run-binary-fast run-binary-debug run-fast
 run-binary: $(sim)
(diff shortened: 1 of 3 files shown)


abejgonzalez:
Fixes #309

a0u:
For the record, the problem here was that VCS was setting stdin to non-blocking mode.  If stdin and stdout of the shell pipeline are both connected to the same tty, then the last command in the pipeline would also have its stdout be non-blocking.  Most programs (e.g., `tee`) aren't able to deal with `write(2)` calls returning EAGAIN.

jerryz123:
@a0u Good call, VCS actually does the right thing here, 

```
Warning: The UCLI could not be loaded. Interactive simulation is unavailable.
```

PR #427, opened 2020-02-10
Chipyard changes for TraceRV + Stack Unwinding + Flame Graph support from FirePerf

sagark:
This is just to bump firemarshal. See https://github.com/firesim/firesim/pull/496 and https://github.com/firesim/FireMarshal/pull/129


sagark:
@a0u @NathanTP @alonamid this is ready for review, but will not be merged until the firemarshal PR goes through

PR #433, opened 2020-02-14
[setup] log init submodules

### LINKED ISSUE #426: Copy FireSim's logging for initial setup scripts
E.g. https://github.com/firesim/firesim/blob/master/build-setup.sh

This makes it easier to debug user issues during setup

abejgonzalez:
**Related issue**: Fixes #426 

**Impact**: other

**Release Notes**
Log the setup scripts to help users debug (`init-submodules` and `build-toolchains`). Follows the recommendation given in #426.

### DIFF
--- a/scripts/init-submodules-no-riscv-tools.sh
+++ b/scripts/init-submodules-no-riscv-tools.sh
@@ -1,63 +1,8 @@
 #!/usr/bin/env bash
 
-# exit script if any command fails
+# wrapper to log output from init-submodules script
+
 set -e
 set -o pipefail
 
-RDIR=$(git rev-parse --show-toplevel)
-
-# Ignore toolchain submodules
-cd "$RDIR"
-for name in toolchains/*-tools/*/ ; do
-    git config submodule."${name%/}".update none
-done
-git config submodule.toolchains/libgloss.update none
-git config submodule.toolchains/qemu.update none
-
-# Don't automatically initialize generators with big submodules (e.g. linux source)
-git config submodule.generators/sha3.update none
-
-# Disable updates to the FireSim submodule until explicitly requested
-git config submodule.sims/firesim.update none
-# Disable updates to the hammer tool plugins repos
-git config submodule.vlsi/hammer-cadence-plugins.update none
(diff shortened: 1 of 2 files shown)


jerryz123:
The fireboom/firesim failures are being addressed in #390

a0u:
See:
https://github.com/ucb-bar/chipyard/blob/7bcedfa7ad654f7bbdc61a417dcc39c28ab676f0/scripts/build-util.sh#L50
https://github.com/ucb-bar/chipyard/blob/7bcedfa7ad654f7bbdc61a417dcc39c28ab676f0/scripts/build-util.sh#L83

abejgonzalez:
I think then we can just log the `init-submodules` step since that doesn't have a log. I didn't notice that the `build-utils` made a log. Thoughts?

colinschmidt:
I think we just log the init submodules, and point users to the other log files in that log or the readme/docs/end of build-tools error message.

scripts/init-submodules-no-riscv-tools.sh
```diff
@@ -1,63 +1,8 @@
...
-
-# Only shallow clone needed for basic SW tests
-git submodule update --init software/firemarshal
-
-# Configure firemarshal to know where our firesim installation is
-if [ ! -f $RDIR/software/firemarshal/marshal-config.yaml ]; then
-  echo "firesim-dir: '../../sims/firesim/'" > $RDIR/software/firemarshal/marshal-config.yaml
-fi
-echo "PATH=\$( realpath \$(dirname "\${BASH_SOURCE[0]}") )/software/firemarshal:\$PATH" >> $RDIR/env.sh
+bash scripts/init-submodules-no-riscv-tools-nolog.sh $@ 2>&1 | tee init-submodules-no-riscv-tools.log
```
a0u:
```suggestion
./scripts/init-submodules-no-riscv-tools-nolog.sh "$@" 2>&1 | tee init-submodules-no-riscv-tools.log
```

Quoting is needed so the individual parameters are passed intact, without interpretation or expansion.

PR #434, opened 2020-02-15
Make env.sh zsh-compatible. 

davidbiancolin:
This fixes the env.sh generation for weirdos like me.

**Impact**: software change 

**Release Notes**
Make references `BASH_SOURCE` `zsh` compatible.


davidbiancolin:
Wrong base.

PR #443, opened 2020-02-20
Add .cc and .v files to Makefile dependencies

jerryz123:
**Related issue**: 
Changes to C code and Verilog black boxes will trigger rebuilds

**Type of change**: bug fix

**Impact**: software change

**Release Notes**

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -6,7 +6,7 @@ SHELL=/bin/bash
 #########################################################################################
 # variables to get all *.scala files
 #########################################################################################
-lookup_scala_srcs = $(shell find -L $(1)/ -name target -prune -o -iname "*.scala" -print 2> /dev/null)
+lookup_scala_srcs = $(shell find -L $(1)/ -name target -prune -o \( -iname "*.scala" -o -iname "*.cc" -o -iname "*.v" \) -print 2> /dev/null)
 
 SOURCE_DIRS=$(addprefix $(base_dir)/,generators sims/firesim/sim)
 SCALA_SOURCES=$(call lookup_scala_srcs,$(SOURCE_DIRS))


jerryz123:
@alonamid this might present an issue for gemmini, since you guys package your software within generators/gemmini. Software changes will force hardware rebuilds.

colinschmidt:
I think this seems reasonable.

sagark:
> @alonamid this might present an issue for gemmini, since you guys package your software within generators/gemmini. Software changes will force hardware rebuilds.

isn't this the packaging format we're suggesting everyone use? e.g. sha3 does this too

jerryz123:
We could restrict the c src search to files within src/main/resources, but idk if every project follows that structure

sagark:
what about just searching src/

common.mk
```diff
@@ -6,7 +6,7 @@ SHELL=/bin/bash
 #########################################################################################
 # variables to get all *.scala files
 #########################################################################################
-lookup_scala_srcs = $(shell find -L $(1)/ -name target -prune -o -iname "*.scala" -print 2> /dev/null)
+lookup_scala_srcs = $(shell find -L $(1)/ -name target -prune -o \( -iname "*.scala" -o -iname "*.cc" -o -iname "*.v" \) -print 2> /dev/null)
```
abejgonzalez: IMO this makes this cluttered (one variable `lookup_scala_srcs` does everything). I think it would be better to split it up into multiple variables. I currently do that for the other work that I'm doing.

jerryz123: I don't see why that's necessary, since there appears to be no place where you care about one set of files, but not the other.
I will rename this function to `lookup_sbt_deps`, which is more accurate, as this function returns files which should trigger SBT re-runs when modified.

jerryz123:
> what about just searching src/

Not all scala srcs are available in src/  (see sifive-cache)

jerryz123:
#448 enables a more elegant way to do this. Closing.

PR #449, opened 2020-02-25
Use blackboxed SimDRAM instead of SimAXIMem

zhemao:
**Related issue**: 
https://github.com/ucb-bar/testchipip/pull/66

**Impact**: rtl change

**Release Notes**

* Replace SimAXIMem with SimDRAM blackbox in chipyard and tracegen configurations.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -156,10 +156,14 @@ $(output_dir)/tracegen.result: $(output_dir)/tracegen.out $(AXE)
 
 tracegen: $(output_dir)/tracegen.result
 
+.PHONY: tracegen
+
+#######################################
+# Rules for building DRAMSim2 library #
+#######################################
+
 dramsim_dir = $(base_dir)/tools/DRAMSim2
 dramsim_lib = $(dramsim_dir)/libdramsim.a
 
 $(dramsim_lib):
 	$(MAKE) -C $(dramsim_dir) $(notdir $@)
-
-.PHONY: tracegen
--- a/common.mk
+++ b/common.mk
@@ -129,7 +129,7 @@ $(output_dir)/%.run: $(output_dir)/% $(sim)
 	(set -o pipefail && $(sim) $(PERMISSIVE_ON) +max-cycles=$(timeout_cycles) $(SIM_FLAGS) $(PERMISSIVE_OFF) $< </dev/null | tee $<.log) && touch $@
 
 $(output_dir)/%.out: $(output_dir)/% $(sim)
-	(set -o pipefail && $(sim) $(PERMISSIVE_ON) +max-cycles=$(timeout_cycles) $(VERBOSE_FLAGS) $(PERMISSIVE_OFF) $< </dev/null 2> >(spike-dasm > $@) | tee $<.log)
+	(set -o pipefail && $(sim) $(PERMISSIVE_ON) +dramsim +max-cycles=$(timeout_cycles) $(VERBOSE_FLAGS) $(PERMISSIVE_OFF) $< </dev/null 2> >(spike-dasm > $@) | tee $<.log)
 
 #########################################################################################
 # include build/project specific makefrags made from the generator
@@ -155,5 +155,11 @@ $(output_dir)/tracegen.result: $(output_dir)/tracegen.out $(AXE)
 	$(base_dir)/scripts/check-tracegen.sh $< > $@
 
 tracegen: $(output_dir)/tracegen.result
+
+dramsim_dir = $(base_dir)/tools/DRAMSim2
+dramsim_lib = $(dramsim_dir)/libdramsim.a
+
+$(dramsim_lib):
+	$(MAKE) -C $(dramsim_dir) $(notdir $@)
 
 .PHONY: tracegen
--- a/generators/utilities/src/main/scala/Simulator.scala
+++ b/generators/utilities/src/main/scala/Simulator.scala
@@ -83,6 +83,11 @@ object GenerateSimFiles extends App with HasGenerateSimConfig {
   }
   def resources(sim: Simulator): Seq[String] = Seq(
     "/testchipip/csrc/SimSerial.cc",
+    "/testchipip/csrc/SimDRAM.cc",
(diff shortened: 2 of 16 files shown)


jerryz123:
Marking DO NOT MERGE while we figure out the best way to do this

generators/utilities/src/main/scala/Simulator.scala
```diff
@@ -83,6 +83,11 @@ object GenerateSimFiles extends App with HasGenerateSimConfig {
   }
   def resources(sim: Simulator): Seq[String] = Seq(
```
jerryz123: Why do we edit this instead of just using `addResource`?

zhemao: In emulator.cc, you need certain global variables defined or it won't compile. If there is no SimDRAM and we don't add these files here, we'd get a compile error in verilator because `dramsim` isn't there.

a0u: It might also work to define `dramsim` as a weak symbol somewhere, so `addResource` can be used for these files.

a0u:
We could make `ini_dir` initialized to the build directory path through a macro defined in `CFLAGS` during VCS/verilator compilation.  Symlinking still seems to be the simpler option.

zhemao:
@a0u Running tests through `make` invokes the simulator in the generated-src directory, which is why I'm relying on BlackBox resource generation to add dramsim2_ini. Where do you think we should add the line to create the simlink, in the Makefile?

davidbiancolin:
It's been a long time since i looked at the DramSim2 code base -- could we just not bake in a configuration in our wrapper. Practically nobody modifies the ini files.

zhemao:
That's probably a better solution than copying or symlinking the same file literally everywhere.

a0u:
The working directory for simulations is `sims/verilator` and `sims/vcs`, not `generated-src`.  In any case, `AddResource` flattens the directory structure so the `dramsim2_ini` directory isn't present anyway.

a0u:
For VCS, I get a compile error:
```
Error-[ICTA] Incompatible complex type                                                                              
.../chipyard/sims/vcs/generated-src/chipyard.TestHarness.RocketConfig/SimDRAM.v, 95       
  Incompatible complex type assignment                                                                              
  Type of source expression is incompatible with type of target expression. 
  Mismatching types cannot be used in assignments, initializations and                                              
  instantiations. The type of the target is 'chandle', while the type of the                                        
  source is 'int'.                                                                                                  
  Source Expression: 0                                    
  Instance stack trace:                                                                                             
      SimDRAM#(32, 64, 4, 268435456, 64, , )
...
```

zhemao:
@a0u LOL, this is a classic case of "fix something in verilator thus breaking VCS". I think I've got it fixed now. You'll see the fix in the testchipip PR.

common.mk
```diff
@@ -129,7 +129,7 @@ $(output_dir)/%.run: $(output_dir)/% $(sim)
 	(set -o pipefail && $(sim) $(PERMISSIVE_ON) +max-cycles=$(timeout_cycles) $(SIM_FLAGS) $(PERMISSIVE_OFF) $< </dev/null | tee $<.log) && touch $@
 
 $(output_dir)/%.out: $(output_dir)/% $(sim)
-	(set -o pipefail && $(sim) $(PERMISSIVE_ON) +max-cycles=$(timeout_cycles) $(VERBOSE_FLAGS) $(PERMISSIVE_OFF) $< </dev/null 2> >(spike-dasm > $@) | tee $<.log)
+	(set -o pipefail && $(sim) $(PERMISSIVE_ON) +dramsim +max-cycles=$(timeout_cycles) $(VERBOSE_FLAGS) $(PERMISSIVE_OFF) $< </dev/null 2> >(spike-dasm > $@) | tee $<.log)
```
a0u: Is `+dramsim` intended to be the default?

I'm curious why `$(SIM_FLAGS)` isn't used here like in the other simulation rules. @abejgonzalez

```suggestion
	(set -o pipefail && $(sim) $(PERMISSIVE_ON) +max-cycles=$(timeout_cycles) $(SIM_FLAGS) $(VERBOSE_FLAGS) $(PERMISSIVE_OFF) $< </dev/null 2> >(spike-dasm > $@) | tee $<.log)
```

zhemao: Yeah, I put it as the default. I want to make sure it gets tested in CI.

abejgonzalez: I think that might be a mistake.... though not very impactful in the normal case. I would recommend adding it.

common.mk
```diff
@@ -156,4 +156,10 @@ $(output_dir)/tracegen.result: $(output_dir)/tracegen.out $(AXE)
 
 tracegen: $(output_dir)/tracegen.result
 
+dramsim_dir = $(base_dir)/tools/DRAMSim2
```
abejgonzalez: Can we add some header comment to this section and move the `.PHONY` to be closer to the `tracegen` rule?

Something like this:
```
.PHONY: tracegen
tracegen: ...

#############################
# Target for DRAMSIM2 ....
#############################
```

sims/vcs/dramsim2_ini
```diff
@@ -0,0 +1 @@
+../../generators/testchipip/src/main/resources/dramsim2_ini
```
abejgonzalez: Is there anyway for this file not to exist in the sim. directories?

abejgonzalez: Maybe refer to it from within the `generated-src` and have it copied in there?

a0u: This was discussed in earlier comments.  The issue is that `ini_dir` is currently hardcoded to a relative path in the dramsim2 wrapper, so although the INI file is already copied (uselessly at the moment) as a resource, setting `ini_dir` to the `generated-src` directory requires a bit of contortion.  Symlinking is the simplest approach for now.

abejgonzalez: Sorry I jumped straight into the code and didn't see the main comments.

PR #451, opened 2020-02-26
Rename Config Mixins to Fragments

abejgonzalez:
**Related issue**: 

**Impact**: other

**Release Notes**
This renames all instances of config mixins to fragments to better distinguish between traits/mixins that apply to the `Top` and config fragments that make configs.

### DIFF
--- a/docs/Generators/IceNet.rst
+++ b/docs/Generators/IceNet.rst
@@ -8,7 +8,7 @@ A diagram of IceNet's microarchitecture is shown below.
 
 .. image:: ../_static/images/nic-design.png
 
-There are four basic parts of the NIC: the :ref:`Controller`, which takes requests 
+There are four basic parts of the NIC: the :ref:`Controller`, which takes requests
 from and sends responses to the CPU; the :ref:`Send Path`, which reads data from
 memory and sends it out to the network; the :ref:`Receive Path`, which receives
 data from the network and writes it to memory; and, optionally,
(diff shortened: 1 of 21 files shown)


alonamid:
I'd rather we use the full term "config fragment" every time we refer to it, rather than just "fragment". I think it maker it clearer, and keeps it within the context of configs

docs/Generators/IceNet.rst
```diff
@@ -80,8 +80,8 @@ and ``HasPeripheryIceNICModuleImp`` to the module implementation. If you
 are confused about the distinction between lazy module and module
 implementation, refer to :ref:`Cake Pattern`.
 
-Then add the ``WithIceNIC`` config mixin to your configuration. This will
-define ``NICKey``, which IceNIC uses to determine its parameters. The mixin
+Then add the ``WithIceNIC`` config fragment to your configuration. This will
+define ``NICKey``, which IceNIC uses to determine its parameters. The fragment
```
abejgonzalez: Address this.

generators/chipyard/src/main/scala/GCD.scala
```diff
@@ -200,8 +200,8 @@ trait CanHavePeripheryGCDModuleImp extends LazyModuleImp {
 // DOC include end: GCD imp trait
 
 
-// DOC include start: GCD mixin
+// DOC include start: GCD fragment
```
abejgonzalez: Address this.

generators/chipyard/src/main/scala/ConfigFragments.scala
```diff
@@ -31,21 +31,21 @@ object ConfigValName {
...
+// Common Parameter Fragments
+// --------------------------
 
 class WithBootROM extends Config((site, here, up) => {
   case BootROMParams => BootROMParams(
     contentFileName = s"./bootrom/bootrom.rv${site(XLen)}.img")
 })
 
-// DOC include start: gpio mixin
+// DOC include start: gpio fragment
```
abejgonzalez: Address this.

docs/Generators/IceNet.rst
alonamid:
```suggestion
define ``NICKey``, which IceNIC uses to determine its parameters. The config fragment
```

PR #452, opened 2020-02-26
Tutorial development

jerryz123:
- Moves `WithRenumberHarts` to Chipyard
- Adds BlackBoxSha3 option
- Reorganizes `chiypard/src` directory into `example`, `config` subdirectories
- Adds `TutorialConfigs.scala`, meant to be used in live tutorials

The plan is to have a `setup-tutorial.sh` script delete the sha3 submodule, and comment out the `Sha3RocketConfig` in `RocketConfigs.scala`, as well as the `sha3` lines in the `build.sbt`, to provide participants the experience of integrating the accelerator from scratch.


scripts/tutorial-patches/RocketConfigs.scala.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/generators/chipyard/src/main/scala/config/RocketConfigs.scala b/generators/chipyard/src/main/scala/config/RocketConfigs.scala
+index f0548b2..b0d0135 100644
+--- a/generators/chipyard/src/main/scala/config/RocketConfigs.scala
++++ b/generators/chipyard/src/main/scala/config/RocketConfigs.scala
+@@ -293,7 +293,7 @@ class Sha3RocketConfig extends Config(
+   new chipyard.config.WithBootROM ++
+   new chipyard.config.WithUART ++
+   new chipyard.config.WithL2TLBs(1024) ++
+-  new sha3.WithSha3Accel ++                                // add SHA3 rocc accelerator
++//  new sha3.WithSha3Accel ++                                // add SHA3 rocc accelerator
```
sagark: can we just make this a different config with the commented out bit, rather than modifying this?

sagark: like Sha3RocketTutorialConfig,
which has this same line commented out

jerryz123: Note that we need a checked in config with the fragment enabled for regressions, see the new `chipyard-sha3` tests which verify the Sha3 thing working

jerryz123: Wait what the TutorialConfigs exist in TutorialConfigs.scala. My point above was that we still need a config with the accel built in to enable CI on the accel

sagark:
do we still want to ship this, or close it?

jerryz123:
> do we still want to ship this, or close it?

I'm in favor of shipping. The changes are non-intrusive, and moving WithRenumberHarts is a useful into chipyard is a useful addition.

build.sbt
```diff
@@ -123,7 +123,9 @@ lazy val testchipip = (project in file("generators/testchipip"))
   .settings(commonSettings)
 
 lazy val chipyard = conditionalDependsOn(project in file("generators/chipyard"))
-  .dependsOn(boom, hwacha, sifive_blocks, sifive_cache, utilities, sha3, gemmini, icenet, tracegen)
+  .dependsOn(boom, hwacha, sifive_blocks, sifive_cache, utilities,
+    sha3,
```
alonamid: I feel like there should be a comment here about this, because someone is definitely going to add another dependency at some point and merge the lines (and then fail CI and not understand why....)

jwright6323: For those of us not in the loop, why does this fail?

jerryz123: To setup the repo for tutorial mode (delete SHA3 to walk attendees through integrating it), we need to patch away lines referencing sha3.

If sha3 is not on its own line, the patch will fail.

alonamid:
Why is the barstools bump part of this PR?
Also, I still think the sha3 separate line in the build.sbt should have an explanation comment (as is clear for the comments here)

build.sbt
sagark: +1 on sticking a comment on this line

build.sbt
```diff
@@ -195,7 +197,9 @@ lazy val midas      = ProjectRef(firesimDir, "midas")
 lazy val firesimLib = ProjectRef(firesimDir, "firesimLib")
 
 lazy val firechip = (project in file("generators/firechip"))
-  .dependsOn(boom, hwacha, chipyard, icenet, testchipip, sifive_blocks, sifive_cache, sha3, utilities, midasTargetUtils, midas, firesimLib % "test->test;compile->compile")
+  .dependsOn(boom, hwacha, chipyard, icenet, testchipip, sifive_blocks, sifive_cache,
+    sha3,
```
sagark: same

PR #457, opened 2020-03-02
Dsptools examples 

ryan-lund:
Adds two examples of accelerators created using dsptools that can be attached to a Rocket Core as MMIOs. The blocks are:
1. Passthrough (sanity check)
2. Generic FIR Filter


generators/example/src/main/scala/ConfigMixins.scala
```diff
@@ -244,3 +244,14 @@ class WithLoopbackNIC extends Config((site, here, up) => {
...
 })
+
+/**
+ * Mixin to add DSPTools components
+ */
+class WithUIntPassthrough extends Config((site, here, up) => {
+  case PassthroughKey => Some(PassthroughParams(depth = 8))
+})
+
+class WithUIntTestFIR extends Config((site, here, up) => {
```
grebe: The example should probably use `FixedPoint` for the FIR filter.

generators/example/src/main/scala/dsptools/GenericFIR.scala
```diff
@@ -0,0 +1,197 @@
...
+  }
+
+  // connect input to first cell
+  directCells.head.in.bits.data := io.in.bits.data
+  directCells.head.in.bits.carry := Ring[T].zero
+  directCells.head.in.valid := io.in.valid
+  io.in.ready := directCells.head.in.ready
+
+  // connect adjacent cells
+  for ((current, next) <- directCells.zip(directCells.tail)) {
```
grebe: This is doing something a little tricky, and the comment should probably reflect that. `directCells.tail` is one element shorter than `directCells`, and `zip` doesn't look at the last element of `directCells`. Maybe we should call `dropRight`? I'm not sure what's least confusing. It definitely works as written, but I think it would be good to make the example easy to understand. What do you think?

generators/example/src/main/scala/dsptools/Passthrough.scala
```diff
@@ -0,0 +1,172 @@
...
+  * @tparam T Type parameter for passthrough data type
+  */
+class TLPassthroughBlock[T<:Data:Ring]
+(
+  val proto: T
+)(implicit p: Parameters) extends
+PassthroughBlock[TLClientPortParameters, TLManagerPortParameters, TLEdgeOut, TLEdgeIn, TLBundle, T](proto)
+with TLDspBlock
+
+/**
```
grebe: Let's remove this commented out stuff. I think it actually should work to do this way, but as currently written I think it makes a better example

generators/example/src/main/scala/dsptools/Passthrough.scala
```diff
@@ -0,0 +1,172 @@
...
+  * PassthroughChain is the "right way" to do this, but the dspblocks library seems to be broken.
+  * In the interim, this should work.
+  * @param depth depth of queues
+  * @param ev$1
+  * @param ev$2
+  * @param ev$3
+  * @param p
+  * @tparam T Type parameter for passthrough, i.e. FixedPoint or DspReal
+  */
+class TLPassthroughThing[T<:Data:Ring](params: PassthroughParams, proto: T)(implicit p: Parameters)
```
grebe: Instead of `Thing`, let's call it something else. Maybe `Chain`?

generators/example/src/main/scala/dsptools/GenericFIR.scala
```diff
@@ -0,0 +1,197 @@
...
+(
+  val genIn: T,
+  val genOut: T,
+  coeffs: Seq[T]
+)(implicit p: Parameters) extends
+GenericFIRBlock[TLClientPortParameters, TLManagerPortParameters, TLEdgeOut, TLEdgeIn, TLBundle, T](
+    genIn, genOut, coeffs
+) with TLDspBlock
+
+class TLGenericFIRThing[T<:Data:Ring] (genIn: T, genOut: T, coeffs: Seq[T], params: GenericFIRParams)(implicit p: Parameters)
```
grebe: Let's call this something other than `Thing`, perhaps `Chain`? `MM`?

alonamid:
Some initial comments:
- I would recommend re-basing this ASAP over the current dev branch (the configs system has changed since PR #390 )
- Needs documentation (in within chipyard/docs)
- The passthrough name is somewhat confusing. Would recommend renaming it (maybe dsp-passthrough?)

grebe:
Maybe `StreamingPassthrough`?

generators/example/src/main/scala/dsptools/Passthrough.scala
ryan-lund: Should I in this case remove the initial PassthroughChain implementation?

generators/example/src/main/scala/dsptools/GenericFIR.scala
ryan-lund: Renamed in latest update (TLGenericFIRChain)

generators/example/src/main/scala/dsptools/Passthrough.scala
ryan-lund: Same as below

ryan-lund:
> Maybe `StreamingPassthrough`?

As of now I've changed Passthrough to this.

docs/Customization/Dsptools-Blocks.rst
```diff
@@ -0,0 +1,105 @@
+.. _dsptools-blocks:
+
+Dsptools Blocks
+===============
+ 
+Another way to create a MMIO peripheral is to use the Dsptools library for Chisel. In this method, a chain is created by placing a custom module inside a ``DspBlock`` and sandwiching that block between a ``ReadQueue`` and ``WriteQueue``. Those queues then act as memory mapped interfaces to the Rocket Chip SoCs. This section will again primarily focuso n designing Tilelink-based peripherals. However, through the resources provided in Dsptools, one could also define an AXI4-based peripheral by following similar steps.
```
alonamid: When you say "chain", do you mean a streaming interface?

alonamid:
```suggestion
Another way to create a MMIO peripheral is to use the Dsptools library for Chisel. In this method, a chain is created by placing a custom module inside a ``DspBlock`` and sandwiching that block between a ``ReadQueue`` and ``WriteQueue``. Those queues then act as memory mapped interfaces to the Rocket Chip SoCs. This section will again primarily focus on designing Tilelink-based peripherals. However, through the resources provided in Dsptools, one could also define an AXI4-based peripheral by following similar steps.
```

docs/Customization/Dsptools-Blocks.rst
```diff
@@ -0,0 +1,105 @@
+.. _dsptools-blocks:
+
+Dsptools Blocks
```
alonamid: There needs to be some introduction here. 
What is the main use-case here? (DSP chains....)
Why does this main use-case need a different flow than the standard MMIO method?

docs/Customization/Dsptools-Blocks.rst
```diff
@@ -0,0 +1,105 @@
...
+===============
+ 
+Another way to create a MMIO peripheral is to use the Dsptools library for Chisel. In this method, a chain is created by placing a custom module inside a ``DspBlock`` and sandwiching that block between a ``ReadQueue`` and ``WriteQueue``. Those queues then act as memory mapped interfaces to the Rocket Chip SoCs. This section will again primarily focuso n designing Tilelink-based peripherals. However, through the resources provided in Dsptools, one could also define an AXI4-based peripheral by following similar steps.
+
+For this example, we will show you how to connect a simple FIR filter created using Dsptools as an MMIO peripheral. The full code can be found in ``generators/example/src/main/scala/dsptools/GenericFIR.scala``.
```
alonamid: Is there a documentation page somewhere that talks about how to write an FIR filter using Dsptools?

docs/Customization/Dsptools-Blocks.rst
```diff
@@ -0,0 +1,105 @@
...
+
+We module ``GenericFIR`` links together a variable number of ``GenericFIRDirectCell`` submodules which work togther to perform the filtering. It is important to note that both modules are type generic, which means that they can be instantiated for any datatype that implements ``Ring`` operations per the specifications on ``T``.
```
alonamid: Can you clarify the first sentence here?

docs/Customization/Dsptools-Blocks.rst
```diff
@@ -0,0 +1,105 @@
+.. _dsptools-blocks:
```
alonamid: @grebe , I think there should be some introduction to DSP tools. I know there is some in the dsp-tools README, but I don't think it's sufficient.

This page talks about Ring operations which could use some additional introduction

alonamid:
I would also like to see the documentation page generalized to a general DSP blocks. The FIR walkthrough should still be there, but as an example, rather than a target design. The target audience of this page should be someone writing an arbitrary DSP block (correlator, averaging window, FFT, etc.), and they should be able to understand what is specific to an FIR filter in this example, and what is generally applicable to their DSP block.

generators/chipyard/src/main/scala/example/dsptools/Passthrough.scala
```diff
@@ -0,0 +1,156 @@
...
+
+class StreamingPassthroughIO[T<:Data:Ring](proto: T) extends Bundle {
+    val in = Flipped(Decoupled(StreamingPassthroughBundle(proto)))
+    val out = Decoupled(StreamingPassthroughBundle(proto))
+}
+object StreamingPassthroughIO {
+    def apply[T<:Data:Ring](proto: T): StreamingPassthroughIO[T] = new StreamingPassthroughIO(proto)
+}
+
+class StreamingPassthrough[T<:Data:Ring](proto: T) extends Module {
```
alonamid, line 44: Update the file name as well to StreamingPassthrough?

tests/Makefile
```diff
@@ -5,7 +5,7 @@ LDFLAGS= -static
 
 include libgloss.mk
 
-PROGRAMS = pwm blkdev accum charcount nic-loopback big-blkdev pingd
+PROGRAMS = pwm blkdev accum charcount nic-loopback big-blkdev pingd passthrough fir
```
alonamid, line 8: update the "passthrough" name here as well?

docs/Customization/Dsptools-Blocks.rst
```diff
@@ -0,0 +1,105 @@
...
+
+Dsptools Blocks
+===============
+ 
+Another way to create a MMIO peripheral is to use the Dsptools library for Chisel. In this method, a memory interface is created by creating a "chain". This chain consists of a custom module placed inside a ``DspBlock``, which is then sandwiched between a ``ReadQueue`` and ``WriteQueue``. Those queues then act as memory mapped interfaces to the Rocket Chip SoCs. This section will again primarily focus on designing Tilelink-based peripherals. However, through the resources provided in Dsptools, one could also define an AXI4-based peripheral by following similar steps.
```
alonamid, line 6: Can you clarify the difference between adding an MMIO peripheral using this method vs. the "usual" method. In other words: is this a "special case" of the general method, or is this an alternative?

docs/Customization/Dsptools-Blocks.rst
```diff
@@ -0,0 +1,105 @@
...
+
+For this example, we will show you how to connect a simple FIR filter created using Dsptools as an MMIO peripheral. The full code can be found in ``generators/example/src/main/scala/dsptools/GenericFIR.scala``. That being said, one could substitute any module with a ready valid interface in the place of the FIR and achieve the same results. As long as the read and valid signals of the module are attached to those of a corresponding ``DSPBlock`` wrapper, and that wrapper is placed in a chain with a ``ReadQueue`` and a ``WriteQueue``, following the general outline establised by these steps will allow you to interact with that block as a memory mapped IO
```
alonamid, line 8:
```suggestion
For this example, we will show you how to connect a simple FIR filter created using Dsptools as an MMIO peripheral. The full code can be found in ``generators/example/src/main/scala/dsptools/GenericFIR.scala``. That being said, one could substitute any module with a ready valid interface in the place of the FIR and achieve the same results. As long as the ready and valid signals of the module are attached to those of a corresponding ``DSPBlock`` wrapper, and that wrapper is placed in a chain with a ``ReadQueue`` and a ``WriteQueue``, following the general outline established by these steps will allow you to interact with that block as a memory mapped IO
```

docs/Customization/Dsptools-Blocks.rst
```diff
@@ -0,0 +1,105 @@
...
+
+The module ``GenericFIR`` is the overall wrapper of our FIR module. This module links together a variable number of ``GenericFIRDirectCell`` submodules, each of which performs the computations for one coefficient in a FIR direct form architecture. It is important to note that both modules are type generic, which means that they can be instantiated for any datatype that implements ``Ring`` operations per the specifications on ``T``.
```
alonamid, line 10: Could you add a diagram to clarify this?

docs/Customization/Dsptools-Blocks.rst
alonamid: Also, what would be other good examples of things other than FIR to implement using this method?

docs/Customization/Dsptools-Blocks.rst
```diff
@@ -0,0 +1,105 @@
...
+Top Level Traits
+----------------
+As in the previous MMIO example, we use a cake pattern to hook up our module to our SoC.
+
+.. literalinclude:: ../../generators/example/src/main/scala/dsptools/GenericFIR.scala
+    :language: scala
+    :start-after: DOC include start: CanHavePeripheryUIntTestFIR chisel
+    :end-before: DOC include end: CanHavePeripheryUIntTestFIR chisel
+
+Note that this is the point at which we decide the datatype for our FIR. It is also possible with some reworking to push the datatype selection out to the top level.
```
alonamid, line 62: If you're not going to explain what "reworking" you're talking about, I would just remove that comment

docs/Customization/Dsptools-Blocks.rst
```diff
@@ -0,0 +1,105 @@
...
+
+.. literalinclude:: ../../generators/example/src/main/scala/RocketConfigs.scala
+    :language: scala
+    :start-after: DOC include start: FIRRocketConfig
+    :end-before: DOC include end: FIRRocketConfig
+
+Testing
+-------
+
+We can now test that the FIR is working. The test program is found in ``tests/gcd.c``.
```
alonamid, line 91:
```suggestion
We can now test that the FIR is working. The test program is found in ``tests/fir.c``.
```

docs/Customization/index.rst
```diff
@@ -11,7 +11,9 @@ These guides will walk you through customization of your system-on-chip:
 
 - Adding custom MMIO widgets to the Chipyard memory system by Tilelink or AXI4, with custom Top-level IOs
 
-- Standard practices for using keys, traits, and configs to parameterize your design
+- Adding custom Dsptools based blocks as MMIO widgets.
```
alonamid, line 14: I would consider putting this in a different place, since it is in fact a "special case" of the previous section

generators/chipyard/src/main/scala/config/RocketConfigs.scala
```diff
@@ -384,3 +384,15 @@ class RingSystemBusRocketConfig extends Config(
   new freechips.rocketchip.subsystem.WithNBigCores(1) ++
   new freechips.rocketchip.system.BaseConfig)
 // DOC include end: RingSystemBusRocket
+
+class UIntStreamingPassthroughRocketConfig extends Config(
+  new chipyard.example.WithUIntStreamingPassthrough ++     // use top with tilelink-controlled passthrough
+  new RocketConfig
```
alonamid, line 390: Our current approach is to use explicit expanded configs rather than this inherited one (see every other config in this file)

generators/chipyard/src/main/scala/config/RocketConfigs.scala
```diff
@@ -384,3 +384,15 @@ class RingSystemBusRocketConfig extends Config(
...
+)
+
+// DOC include start: FIRRocketConfig
+class UIntTestFIRRocketConfig extends Config (
+  new chipyard.example.WithUIntTestFIR ++ // use top with tilelink-controlled FIR
+  new RocketConfig
```
alonamid, line 396: Our current approach is to use explicit expanded configs rather than this inherited one (see every other config in this file)

alonamid:
A few additional comments:
- @grebe , at some point, can you PR this into a branch within the chipyard repo (not dev), and then open a new PR to dev so CI will run? 
- For consistency sake, make sure `passthrough` is replaced with `streaming-passthrough` everywhere (including file names, C files, documentation, etc.)
- Maybe talk about the streaming-passthrough example in the documentation as well?
- From a documentation and organization perspective, is this a "special case" of an MMIO peripheral, or an alternative? We've been describing RoCC and MMIO as a pseudo "dichotomy" for custom blocks, and I think it's worth clarifying where do these dsptools examples fit in this world. I believe you're trying to put an emphasis on streaming chains, but I'm not sure this idea is conveyed well enough. If this is a special case of MMIO, I would prefer to put perhaps put it as a subsection of the general MMIO explanation, or clarify that this is essentially another example of the same method.
- Add a CI test for dsptools

lpppeipei:
> I would also like to see the documentation page generalized to a general DSP blocks. The FIR walkthrough should still be there, but as an example, rather than a target design. The target audience of this page should be someone writing an arbitrary DSP block (correlator, averaging window, FFT, etc.), and they should be able to understand what is specific to an FIR filter in this example, and what is generally applicable to their DSP block.

About  " The target audience of this page should be someone writing an arbitrary DSP block (correlator, averaging window, FFT, etc.)",  now  I  would  like  to  write a FFT   DSP  Block , is there some information  for  me?  Thanks  a  lot  !

Issue #453, opened 2020-02-27
Document how to update the repo


jwright6323:
We don't really have a section on updating to a new release or dev branch commit. It probably goes here or in a subsection after it: https://chipyard.readthedocs.io/en/latest/Chipyard-Basics/Initial-Repo-Setup.html

IMO it's not super clear if you're expect to rerun `./scripts/init-submodules-no-riscv-tools.sh` or do a `git submodule update --recursive`. I would instinctively do the latter, but this time I remembered that doesn't work, so I cd'ed into `scripts` and ran `./init-submodules-no-riscv-tools.sh`, which doesn't work either because you can't be inside of `scripts`. So, we should probably also have that script emit an error message if you're in the wrong spot (or fix it).

I'll PR a doc and script fix, but I'll forget about it if I don't file an issue.

jwright6323:
Fixed by #459, not sure why it didn't close automatically.

PR #459, opened 2020-03-04
Add a git version check to the init scripts and make them work outside of the repo root. 

jwright6323:
**Related issue**: #453 

**Type of change**: bug fix

**Impact**: other

**Release Notes**
Add a git version check to the init scripts and make them work outside of the repo root.

### DIFF
--- a/scripts/init-submodules-no-riscv-tools-nolog.sh
+++ b/scripts/init-submodules-no-riscv-tools-nolog.sh
@@ -5,7 +5,12 @@ set -e
 set -o pipefail
 
 # Check that git version is at least 1.7.8
-MYGIT=$(git --version | awk '{print $3;}')
+MYGIT=$(git --version)
+MYGIT=${MYGIT#'git version '} # Strip prefix
+case ${MYGIT} in
+[1-9]*) ;;
+*) echo 'warning: unknown git version' ;;
+esac
 MINGIT="1.7.8"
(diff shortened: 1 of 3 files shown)


jwright6323:
Labeled "DO NOT MERGE" until we confirm that 1.7.8 is the git version we care about. I did confirm that the version checking logic works by making a dummy `git` executable that echos different version number strings.

jwright6323:
I confirmed that both updating and a fresh clone work with git 1.7.8.

scripts/init-submodules-no-riscv-tools-nolog.sh
```diff
@@ -4,6 +4,14 @@
 set -e
 set -o pipefail
 
+# Check that git version is at least 1.7.8
+MYGIT=$(git --version | awk '{print $3;}')
```
a0u: If somehow `MYGIT` isn't a version string but begins with a non-numeric character, then it would pass the check.  This probably isn't worth considering because if the `git --version` output ever changed, then it would imply a sufficiently new git.

Although overkill, we could make the version parsing more robust like so:
```suggestion
MYGIT=$(git --version)
MYGIT=${MYGIT#'git version '} # Strip prefix
case ${MYGIT} in
[1-9]*) ;;
*) echo 'warning: unknown git version' ;;
esac
```

jwright6323: Cool, thanks for the improvement- applied (and tested locally just to be diligent).

PR #461, opened 2020-03-06
Add ring topology system bus

zhemao:
**Related issue**: 

**Impact**: rtl change 

**Release Notes**
* Add a TileLink network that uses a ring topology


generators/chipyard/src/main/scala/RocketConfigs.scala
```diff
@@ -365,3 +365,6 @@ class ScratchpadRocketConfig extends Config(
   new freechips.rocketchip.subsystem.WithNBigCores(1) ++
   new freechips.rocketchip.system.BaseConfig)
 // DOC include end: scratchpadrocket
+
+class RingSystemBusRocketConfig extends Config(
```
jerryz123: Can you refactor this in the style of other configs here? Fully expand it.

PR #464, opened 2020-03-08
respect --no-firesim flag in init-submodules script

zhemao:
Firesim's build-setup script expects the init-submodules script in chipyard to respect a "--no-firesim" flag and not initialize the sims/firesim submodule. The current script does not do this, which means a second copy of firesim will be checked out if firesim is used as top. This change adds support for the flag back in. 

**Related issue**: 

**Type of change**: bug fix

**Impact**: other

**Release Notes**

### DIFF
--- a/scripts/init-submodules-no-riscv-tools-nolog.sh
+++ b/scripts/init-submodules-no-riscv-tools-nolog.sh
@@ -19,6 +19,18 @@ fi
 
 RDIR=$(git rev-parse --show-toplevel)
 
+FIRESIM=true
+
+while test $# -gt 0
+do
+    case "$1" in
+        --no-firesim)
+            FIRESIM=false
+            ;;
+    esac
+    shift
+done
+
 # Ignore toolchain submodules
 cd "$RDIR"
 for name in toolchains/*-tools/*/ ; do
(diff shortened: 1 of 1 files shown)


davidbiancolin:
can we just remove this flag entirely? this script should implicitly be no firesim

zhemao:
You're right, the firesim-setup.sh script already initializes the firesim submodule, so not sure why it's even necessary to initialize firesim in init-submodules.

davidbiancolin:
it's necessary to pull in the scala sources. however that should need to
happen in all cases

zhemao:
What do you mean? The scala sources in firesim are only necessary if you want to build firesim.

davidbiancolin:
the generators depends on projects in firesim. it should be evident in the
build.sbt . if things still compile okay without firesim initialized then I
agree you should just remove any initialization in this script

zhemao:
Just checked the build.sbt. It's true that the regular libraries depends on stuff in firesim (specifically midas targetutils). But when firesim is used as top, the build.sbt uses the top-level firesim instead of its own submodule. So adding this "--no-firesim" flag is, in fact, the correct way.

abejgonzalez:
Shouldn't this be closed (as discussed in the Chipyard meeting?) @zhemao

scripts/init-submodules-no-riscv-tools-nolog.sh
```diff
@@ -56,15 +68,17 @@ git config --unset submodule.software/firemarshal.update
 # Non-recursive clone to exclude riscv-linux
 git submodule update --init generators/sha3
 
-git config --unset submodule.sims/firesim.update
-# Minimal non-recursive clone to initialize sbt dependencies
-git submodule update --init sims/firesim
-(
```
davidbiancolin: What you can do is delete everything in the ( )

PR #468, opened 2020-03-13
Target-Facing Support for Multiclock Simulation in FireSim

davidbiancolin:
This is the chipyard side of the multiclock PR in FireSim. I'm mostly interested in feedback on the chisel API for the clock bridge. How we choose to architect in support for multiple clock domains can easily change in the future.

**Related issue**: firesim/firesim#441 

**Release Notes**
Support simulation of targets with multiple clock domains in FireSim


generators/firechip/src/main/scala/BridgeBinders.scala
```diff
@@ -55,19 +56,17 @@ class WithFASEDBridge extends RegisterBridgeBinder({
...
+        FASEDBridge(t.clock, axi4Bundle, t.reset.toBool,
           CompleteConfig(p(firesim.configs.MemModelKey), nastiKey, Some(AXI4EdgeSummary(edge))))
       })
     }).toSeq
 })
 
-class WithTracerVBridge extends RegisterBridgeBinder({
-  case target: HasTraceIOImp => TracerVBridge(target.traceIO)(target.p)
-})
+class WithTracerVBridge extends Config((_,_,_) => { case InstantiateTracerVBridges => true })
```
jerryz123: I'm curious why this moves away from the IOBinders approach? 
I believe @abejgonzalez is developing the Dromajo widget as another BridgeBinder that attaches to `HasTraceIOImp`.

davidbiancolin: I made this change prior to your generator unification work, and i thought it would be nice to move the instantiation closer to the source.  I was thiking that ideally the bridge could be instantiated directly in the tile itself perhaps using one of the generation hooks that exists in rocketchip. Then i could just delete the target mixins altogether.

But yes, in light of recent changes, I think it makes sense to continue plumbing this out to the top. I'll switch back to that scheme.

abejgonzalez: Yes. Jerry is right. Right now the Dromajo Bridge nicely hooks into the `TraceIO`.

generators/firechip/src/main/scala/TargetConfigs.scala
```diff
@@ -344,3 +346,29 @@ class FireSimTraceGenL2Config extends Config(
...
+
+
+class WithRationalTiles(multiplier: Int, divisor: Int) extends Config((site, here, up) => {
+  case FireSimClockKey => FireSimClockParameters(Seq(RationalClock("TileDomain", multiplier, divisor)))
+  case RocketCrossingKey => up(RocketCrossingKey, site) map { r =>
+    r.copy(crossingType = RationalCrossing())
+  }
+  case BoomCrossingKey => up(BoomCrossingKey, site) map { r =>
+    r.copy(crossingType = RationalCrossing())
+  }
```
abejgonzalez: Hmm. I have no idea if the `ArianeTile` deals with crossings properly... but I think this should include this for now:

```
case ArianeCrossingKey => up(ArianeCrossingKey, site) map { a =>
    a.copy(crossingType = RationalCrossing())
  }
```

abejgonzalez: Also nit: change the map from `r` -> `b` for BOOM?

generators/firechip/src/main/scala/TargetMixins.scala
```diff
@@ -72,3 +78,15 @@ trait CanHaveMultiCycleRegfileImp {
   }
 }
 
+trait HasFireSimClockingImp extends HasAdditionalClocks {
+  val outer: HasTiles
+  val (tileClock, tileReset) = p(FireSimClockKey).additionalClocks.headOption match {
```
abejgonzalez: Can you give an example of assigning different clocks to each of the tiles? Is that supported with this right now? It looks like you have the tiles on one domain and the uncore on another.

generators/firechip/src/main/scala/TargetConfigs.scala
```diff
@@ -344,3 +346,29 @@ class FireSimTraceGenL2Config extends Config(
     outerLatencyCycles = 50) ++
   new WithTraceGenBridge ++
   new FireSimRocketChipConfig)
+
+
+class WithRationalTiles(multiplier: Int, divisor: Int) extends Config((site, here, up) => {
```
colinschmidt: Is this the right place for this? Or name?
I think there is a similar Config object in rocketchip

generators/firechip/src/main/scala/TargetMixins.scala
```diff
@@ -72,3 +78,15 @@ trait CanHaveMultiCycleRegfileImp {
...
+    case Some(RationalClock(_, numer, denom)) if numer != denom => (clocks(1), ResetCatchAndSync(clocks(1), reset.toBool))
```
colinschmidt: Does this check really do anything? I.e. are there generators that do this math and don't realize they are making a synchronized design?
Also are the magic numbers necessary?

generators/firechip/src/main/scala/TargetConfigs.scala
davidbiancolin: I'll look for it.

generators/firechip/src/main/scala/TargetMixins.scala
davidbiancolin: The numer != denom check? No, but it saves the catch and sync hardware in the synchronous case. I can remove it. 

I think how we plumb the clocks through the target is mostly a separate question to how we want to expose the generation of those clocks for firesim. I mostly just hacked together the prior to get something working quick.

In light of @jerryz123's recent push to harmonize the targets across firechip and chipyard, we should consider making our targets support multiclock by default. So for RTL bound for target-level RTL simulation or physical design, we'll need to define the other derived clocks somehow. 

To remove the vector-indexing hack here i think what we want in the interim is to generate some clock-map [Key/String, ClockInfo] in the parameters object.  We can provide some sugar to let the generator index the right element in the clock vector using the same key.

generators/firechip/src/main/scala/TargetConfigs.scala
jerryz123: I believe we can tear out all the TraceGen related stuff out of firechip. @zhemao do you still want it in here? TraceGenConfigs are now run out of Chipyard.

generators/firechip/src/test/scala/ScalaTestSuite.scala
```diff
@@ -125,51 +130,59 @@ abstract class FireSimTestSuite(
...
+//class RamModelRocketF1Tests extends FireSimTestSuite("FireSim", "FireSimDualRocketConfig", "BaseF1Config_MCRams")
+//class RamModelBoomF1Tests extends FireSimTestSuite("FireSim", "FireSimBoomConfig", "BaseF1Config_MCRams")
+
+// Multiclock tests
+class RocketMulticlockF1Tests extends FireSimTestSuite(
+  "FireSimMulticlockPOC",
+  "FireSimQuadRocketMulticlockConfig",
+  "WithSynthAsserts_BaseF1Config")
+
+// Jerry broke these -- damn it Jerry.
```
colinschmidt: Is this already tracked in a PR/Issue?

jerryz123: Oops
I don't think this tracegen-specific stuff should be tracked in Firechip anyways. Doing a tracegen simulation in firesim should just require changing the BuildTop key.

generators/firechip/src/main/scala/BridgeBinders.scala
```diff
@@ -52,19 +56,24 @@ class WithFASEDBridge extends OverrideIOBinder({
...
           CompleteConfig(p(firesim.configs.MemModelKey), nastiKey, Some(AXI4EdgeSummary(edge))))
       })
     }).toSeq
   }
 })
 
 class WithTracerVBridge extends OverrideIOBinder({
-  (c, r, s, target: CanHaveTraceIOModuleImp) => target.traceIO.map(t => TracerVBridge(t)(target.p)).toSeq
+  (c, r, s, target: CanHaveTraceIOModuleImp) => target.traceIO match {
+    case Some(t) => t.traces.map(tileTrace => TracerVBridge(tileTrace)(target.p))
```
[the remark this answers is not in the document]
davidbiancolin: It does, it's included in the TileTraceIO as it is synchronous to the specific tile that produced it.

generators/firechip/src/test/scala/ScalaTestSuite.scala
davidbiancolin: Yeah i'm in 100% agreement. No one runs the scala tests except me though. I'll fix it when i get the CI to run the scalatests for firechip.

PR #480, opened 2020-03-14
Add ChipTop to enable real chip configs with IO cells, etc.

### LINKED ISSUE #385: Add ChipTop module layer between TestHarness and Top
The `Top` in Chipyard is really a digital-only `Top`. Typically, our chips have had at least one other later of hierarchy above the `DigitalTop`, below the `TestHarness`. This "ChipTop" layer would also contain analog blocks.

It may be useful to create that layer in Chipyard, such that the default designs are more realistic representations of chips that might be taped out.

jwright6323:
This adds an additional layer (ChipTop) between the System module and the TestHarness. The IOBinder API is now changed to take only a single parameter (an Any) and return a 3 things: The IO port(s), the IO cell(s), and a function to call inside the test harness, which is analogous to the old IOBinder function, except that it takes a TestHarness object as an argument instead of (clock, reset, success). 

**Related issue**: Fixes #385 

**Impact**: rtl change

**Release Notes**
* A new Top-level module, ChipTop, has been created. ChipTop instantiates a "system" module specified by BuildSystem.
* BuildTop now builds a ChipTop dut module in the TestHarness by default
* A new BuildSystem key has been added, which by default builds DigitalTop (previously just called Top)
* The IOBinders API has changed. IOBinders are now called inside of ChipTop and return a tuple3 of (IO ports, IO cells, harness functions). The harness functions are now called inside the TestHarness (this is analogous to the previous IOBinder functions).
* IO cell models have been included in ChipTop. These can be replaced with real IO cells for tapeout, or used as-is for simulation.
* The default for the TOP make variable is now ChipTop (was Top)


generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -65,42 +67,55 @@ class ComposeIOBinder[T](fn: => (Clock, Bool, Bool, T) => Seq[Any])(implicit tag
...
 
+class WithGPIOTiedOff extends Config(new OverrideIOBinder({
+  (c, r, s, top: HasPeripheryGPIOModuleImp) => top.gpio.foreach(_.pins.foreach(_.i.ival := false.B)); Nil
+}) ++ new OverrideIOBinder({
+  (c, r, s, top: CanHaveChipTopGPIO) => top.gpio.foreach(_ := false.B); Nil
+}))
+
+class WithSimBlockDevice extends Config(new OverrideIOBinder({
+  (c, r, s, top: CanHavePeripheryBlockDeviceModuleImp) => SimBlockDevice.connect(c, r, top.bdev)(top.p); Nil
+}) ++ new OverrideIOBinder({
```
jerryz123: This paradigm smells fishy to me. Is there a way to do this without duplicating the traits between ChipTop and DigitalTop?

What if instead of calling IOBinders at the Harness level, we call them from within ChipTop? This avoids

jwright6323: I thought about something like that, but I don't think it can work. You'll still need to connect things at the TestHarness level, but you'll have no way to match on traits for those IOBinders. I also tried to make `ChipTop` type-parameterized (`[+T <: SystemModule[System]]`) but it ended up being kind of a pain. We could try that route again if desired, but I think it's still going to require the same `IOBinder` duplication unless we scrap support for raw `Top`, which I was trying to avoid.

jerryz123: What are the reasons to keep around support for raw `Top`?

jwright6323: I haven't looked into any FireSim bridge stuff yet, so I assume that would break at a minimum. But if there's no opposition I do think it would be cleaner to just get rid of raw `Top` support.

jerryz123: @davidbiancolin can FireSim bridges be instantiated anywhere in the design? Can we have the module hierarchy be FireSimTestHarness->ChipTop->DigitalTop, and instantiate the bridges within ChipTop?

abejgonzalez:
Note: I cancelled this PR's CI to free up more resources for the `dev->master` merge. We can run this a bit later.

generators/chipyard/src/main/scala/ChipTop.scala
```diff
@@ -0,0 +1,209 @@
...
+
+}
+
+trait CanHaveChipTopGPIO { this: BaseChipTop =>
+  implicit val p: Parameters
+  // TODO this becomes analog
+  val gpio: Seq[Bool] = system match {
+    case system: HasPeripheryGPIOModuleImp => {
+      system.gpio.map(_.pins).map(_.map(_.i.ival)).flatten.map { pin =>
+        // TODO sane naming
```
colinschmidt: Is this supposed to also be the point where the IOCell model is instantiated?

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -65,42 +67,55 @@ class ComposeIOBinder[T](fn: => (Clock, Bool, Bool, T) => Seq[Any])(implicit tag
...
+  (c, r, s, top: HasPeripheryUARTModuleImp) => UARTAdapter.connect(top.uart)(top.p); Nil
+}) ++ new OverrideIOBinder({
+  (c, r, s, top: CanHaveChipTopUART) => UARTAdapter.connect(top.uart)(top.p); Nil
+}))
+
 // DOC include start: WithSimAXIMem
 class WithSimAXIMem extends OverrideIOBinder({
   (c, r, s, top: CanHaveMasterAXI4MemPortModuleImp) => top.connectSimAXIMem(); Nil
 })
+// Note: No equivalent ChipTop binder, as you generally wouldn't expose AXI4 at the chip level in a real system
```
colinschmidt: Don't tell me what to do 😛

generators/chipyard/src/main/scala/ChipTop.scala
jwright6323: Yes

generators/chipyard/src/main/scala/IOBinders.scala
davidbiancolin: This kinda came up on Monday, but yes we can put the bridges anywhere.

generators/chipyard/src/main/scala/ChipTop.scala
```diff
@@ -0,0 +1,312 @@
...
+  } else {
+    val port = IO(Input(Bool()))
+    systemReset := port
+    (port, Seq())
+  }
+
+  iocells ++= systemClockIO
+  iocells ++= systemResetIO
+
+  def connectReset(r: Bool) { reset := r }
```
jerryz123: Why is this `reset := r`, instead of `systemReset := r`?

jwright6323: `systemReset` is the internal wire, which is already connected to the `reset` IO. This method is needed because `reset` is not a member of the base class.

jwright6323: We probably need this for clock, too...

jwright6323: Never mind- we only need it for reset so we can override the default reset connection when we mix in the `Debug` trait

generators/chipyard/src/main/scala/ChipTop.scala
```diff
@@ -0,0 +1,312 @@
...
+          port <> u
+          port
+        }
+      }
+    }
+    case _ => Seq()
+  }
+}
+
+trait CanHaveChipTopBlockDevice { this: BaseChipTop =>
```
jerryz123: Do we need to plumb this stuff all the way out past ChipTop, given that this isn't the right way to tape-out a chip that supports blkdev?

In general, for stuff like NIC, blockdevice, masteraxi4, etc, is it fine to not provide plumbing between DigitalTop and ChipTop passing through those IOs, since those are sort of simulation-only devices?

jerryz123:
Can we split this into two PRs?
- One PR to add the ChipTop hierarchy layer
- Another to add the IOCells abstraction.

For adding the ChipTop layer, we need to canonicalize the  proper way to attach SimDevices and FireSim bridges, and the right place to call the IOBinders functions. Then we can think about the right way to make everything work once there are IOCells involved in the IOs. This will also give us some time to test that everything works across Firesim with the ChipTop layer, before adding IOCell stuff.

For adding the ChipTop layer, I think the best way to do this is to canonicalize IOBinders as "DigitalTop IOBinders", and have a separate system for driving ChipTop IOs. The simple ChipTop we start with can just have the clock, reset, and success signals plumbed through, and we'll figure out the right way to plumb through more IOs and attach the SimIFs in the second PR.

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -65,74 +79,121 @@ class ComposeIOBinder[T](fn: => (Clock, Bool, Bool, T) => Seq[Any])(implicit tag
...
+    val harnessFn = (th: TestHarness) => { UARTAdapter.connect(ports)(system.p); Nil }
+    Seq((ports, ioCells2d.flatten, Some(harnessFn)))
+  }
+})
+
+// TODO: Add a note about synthesizability of SimBlockDevice
 class WithSimBlockDevice extends OverrideIOBinder({
-  (c, r, s, top: CanHavePeripheryBlockDeviceModuleImp) => top.connectSimBlockDevice(c, r); Nil
+  (system: CanHavePeripheryBlockDeviceModuleImp) => SimBlockDevice.connect(system.clock, system.reset.asBool, system.bdev)(system.p); Nil
 })
```
jerryz123: Can we revert to `top.connectSimBlockDevice` here?

generators/chipyard/src/main/scala/TestHarness.scala
```diff
@@ -26,5 +29,32 @@ class TestHarness(implicit val p: Parameters) extends Module {
...
+
+  // Aliases for clock, reset, and success
+  def c  = clock
+  def r  = reset.asBool
+  def ro = resetOverride
+  def s  = io.success
+}
+
+object TestHarnessUtils {
+
```
jerryz123: Why are these two functions defined here, instead of inlined within the IOBinders? Are they ever reused across multiple IOBinders?

variables.mk
```diff
@@ -36,8 +36,12 @@ ifeq ($(SUB_PROJECT),chipyard)
 	CONFIG_PACKAGE    ?= $(SBT_PROJECT)
 	GENERATOR_PACKAGE ?= $(SBT_PROJECT)
 	TB                ?= TestDriver
+ifneq (,$(findstring Chip,$(CONFIG)))
```
jerryz123: Is this still necessary?

jwright6323: Nope, will fix.

generators/chipyard/src/main/scala/TestHarness.scala
jwright6323: They were previously. We can revert it back, but I could see a situation where we do want to split it out in the future. No strong opinions; it's easy to move either way.

generators/chipyard/src/main/scala/IOBinders.scala
jwright6323: Sure.

jwright6323: This is done

variables.mk
jwright6323: Fixed.

generators/chipyard/src/main/scala/ChipTop.scala
jwright6323: This was removed in a later commit.

generators/chipyard/src/main/scala/TestHarness.scala
jwright6323: I went ahead and changed this back.

jerryz123:
This is pretty neat. I'm happy with how this works.

One minor nit, can we undo the bumps to the sub-modules? Specifically, I don't think icenet and testchipip need bumps with the new approach.

jwright6323:
Yeah, we can undo the bumps since this does not strictly require them. I do think it would be a good idea to still break that code out, though (like how rocket does with `Debug`: https://github.com/chipsalliance/rocket-chip/blob/69df9c78f77afeb097c621607e0ee78cb5dafd4c/src/main/scala/devices/debug/Periphery.scala#L219 )

jwright6323:
Ah, we need the `testchipip` bump so that we can use the object-ized `SerialAdapter` methods. (Otherwise we'd have to duplicate that code).

generators/chipyard/src/main/scala/TestHarness.scala
```diff
@@ -26,5 +24,15 @@ class TestHarness(implicit val p: Parameters) extends Module {
...
   io.success := false.B
-  p(IOBinders).values.map(fn => fn(clock, reset.asBool, io.success, dut))
+  // resetOverride can be overridden via a harnessFunction
+  val resetOverride = Wire(Bool())
+  resetOverride := reset
+  dut.harnessFunctions.foreach(_(this))
+
+  // Aliases for clock, reset, and success
+  def c  = clock
+  def r  = reset.asBool
```
davidbiancolin: Do functions ever need to disambiguate between reset and the resetOverride? AFAICT resetOverride _is_ the reset, with the added bonus it can be assigned to.  I feel like something like this will present
and implicit reset that that is more intuitive to the user: 

```val resetOverride = WireDefault(Bool())
withClockAndReset(clock, resetOverride) {
     dut.harnessFunctions.foreach(_(this))
   }
}
```
and then remove the reset alias "r" as it currently exists.

jerryz123:
> Also can we inline IceNet into TestChipIp?

You mean merge icenet and testchipip into one submodule? Why?

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -31,18 +36,25 @@ import scala.reflect.{ClassTag}
 // You can add your own binder by adding a new (key, fn) pair, typically by using
 // the OverrideIOBinder or ComposeIOBinder macros
 
+
```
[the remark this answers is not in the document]
jwright6323: the `type` definitions have to be in an object, and for whatever reason `package object iobinders` wouldn't compile. Thoughts there?

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -31,18 +36,25 @@ import scala.reflect.{ClassTag}
...
+object types {
+  type TestHarnessFunction = (TestHarness) => Seq[Any]
+  type IOBinderTuple = (Seq[Data], Seq[IOCell], Option[TestHarnessFunction])
```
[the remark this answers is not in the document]
jwright6323: Yep, I'll do a docs/comments pass before we merge

generators/chipyard/src/main/scala/TestHarness.scala
jwright6323: > > Also can we inline IceNet into TestChipIp?
> 
> You mean merge icenet and testchipip into one submodule? Why?

No opinions on the merge, but it does feel a bit out of scope for this PR.

davidbiancolin:
> > Also can we inline IceNet into TestChipIp?
> 
> You mean merge icenet and testchipip into one submodule? Why?

Because submodules suck, and now there exist no instances of people using them independently (AFAIK).

jwright6323:
> > Also can we inline IceNet into TestChipIp?
> 
> You mean merge icenet and testchipip into one submodule? Why?

No opinions on the merge, but it does feel a little out of scope for this PR.

davidbiancolin:
> > > Also can we inline IceNet into TestChipIp?
> > 
> > 
> > You mean merge icenet and testchipip into one submodule? Why?
> 
> No opinions on the merge, but it does feel a little out of scope for this PR.

haha definitely, it just came to mind when i saw you bumped it.

generators/chipyard/src/main/scala/TestHarness.scala
jwright6323: Yeah, I like that. Good suggestion.

jwright6323: We might need to keep the `r` alias though, to support this kind of thing (from the `WithSimDebug` binder):
```
th.ro := th.r | debugPortOpt.map { debug => AsyncResetReg(debug.ndreset).asBool }.getOrElse(false.B)
```

Open to suggestions about giving it a better name, though.

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -1,19 +1,24 @@
...
 import freechips.rocketchip.util._
 
 import sifive.blocks.devices.gpio._
 import sifive.blocks.devices.uart._
 
-import testchipip._
-import icenet._
+import barstools.iocell.chisel._
+
+import chipyard.{TestHarness}
```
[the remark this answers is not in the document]
jwright6323: I tried that and got annoyed writing out `chipyard.TestHarness` in every IO Binder. I'd be willing to change it, though.

generators/chipyard/src/main/scala/IOBinders.scala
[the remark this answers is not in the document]
jwright6323: Fixed.

generators/chipyard/src/main/scala/IOBinders.scala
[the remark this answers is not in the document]
jwright6323: I went ahead and did this change.

generators/chipyard/src/main/scala/TestHarness.scala
jwright6323: I went back and forth on this a bit. I think I've actually decided to scrap all of the aliases and make people use the full names- it adds a lot of clarity IMO. I also think we don't want the `withClockAndReset`, because generally I think the test harness widgets will want to be on the `TestHarness` reset, not `dutReset` (I renamed `resetOverride` to `dutReset`, which I think is clearer). If we find that in the future we need some advanced reset sequencing we can rethink this.

davidbiancolin: yeah that's much clearer.

scripts/tutorial-patches/build.sbt.patch
```diff
@@ -5,7 +5,7 @@ index a633066..3df8b74 100644
 @@ -124,7 +124,7 @@ lazy val testchipip = (project in file("generators/testchipip"))
  
  lazy val chipyard = conditionalDependsOn(project in file("generators/chipyard"))
-   .dependsOn(boom, hwacha, sifive_blocks, sifive_cache, utilities,
+   .dependsOn(boom, hwacha, sifive_blocks, sifive_cache, utilities, iocell,
 -    sha3, // On separate line to allow for cleaner tutorial-setup patches
 +//    sha3, // On separate line to allow for cleaner tutorial-setup patches
```
jerryz123: This should not be commented out?

generators/firechip/src/main/scala/BridgeBinders.scala
```diff
@@ -100,13 +100,31 @@ class WithFireSimMultiCycleRegfile extends ComposeIOBinder({
...
+
+class WithTiedOffDebug extends OverrideIOBinder({
+  (system: HasPeripheryDebugModuleImp) => {
+    Debug.tieoffDebug(system.debug, system.psd)
+    // tieoffDebug doesn't actually tie everything off :/
+    system.debug.foreach(_.clockeddmi.foreach({ cdmi => cdmi.dmi.req.bits := DontCare }))
+    Nil
+  }
+})
+
```
jerryz123: Why don't the IOCell modules work with Firechip?

scripts/tutorial-patches/build.sbt.patch
jwright6323: I think you're misreading the syntax highlighting- this is the patch. The only thing that's changing is the `iocells` part.

generators/firechip/src/main/scala/BridgeBinders.scala
jwright6323: I don't know exactly, but my guess is something to do with constant propagation. If there's an extra clock (e.g. the dmiClock, even if it's tied to 0) it fails during firrtl compilation (see my slack message for a stack trace).

scripts/tutorial-patches/build.sbt.patch
jerryz123: Oh yeah, you are right.

generators/firechip/src/main/scala/BridgeBinders.scala
jerryz123: Ah ok. If we have to have these separate "system-level" tie-offs, instead of the "chip-level" tie-offs, I think we should reflect that in the naming, for example `WithSystemGPIOTiedOff`. We can talk about this tomorrow.

jwright6323: Sure, I'm fine with that.

generators/chipyard/src/main/scala/ConfigFragments.scala
```diff
@@ -65,8 +65,8 @@ class WithL2TLBs(entries: Int) extends Config((site, here, up) => {
   ))
 })
 
-class WithTracegenTop extends Config((site, here, up) => {
-  case BuildTop => (p: Parameters) => Module(LazyModule(new tracegen.TraceGenSystem()(p)).suggestName("Top").module)
+class WithTracegenSystem extends Config((site, here, up) => {
+  case BuildSystem => (p: Parameters) => Module(LazyModule(new tracegen.TraceGenSystem()(p)).suggestName("Top").module)
```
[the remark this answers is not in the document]
jwright6323: @davidbiancolin Should we change it to match?

generators/chipyard/src/main/scala/ConfigFragments.scala
```diff
@@ -150,3 +150,12 @@ class WithControlCore extends Config((site, here, up) => {
...
   case MaxHartIdBits => log2Up(up(RocketTilesKey, site).size + up(BoomTilesKey, site).size + 1)
 })
+
+
+/**
+ * Config fragment to use ChipTopCaughtReset as the top module, which adds a reset synchronizer to
+ * the top-level reset, allowing it to be asynchronous with the clock.
+ */
+class WithChipTopCaughtReset extends Config((site, here, up) => {
+  case BuildTop => (p: Parameters) => Module(new ChipTopCaughtReset()(p).suggestName("top"))
```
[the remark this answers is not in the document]
jwright6323: Yeah, annotation is the right way to do it (https://github.com/ucb-bar/barstools/issues/77). Until we fix that I'll note in the docs that you have to change TOP.

generators/chipyard/src/main/scala/ConfigFragments.scala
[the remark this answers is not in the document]
jwright6323: This was resolved offline.

generators/chipyard/src/main/scala/ConfigFragments.scala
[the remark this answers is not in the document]
jwright6323: Noted

jwright6323:
@abejgonzalez assuming this passes CI it's ready for its green checkmark.

jwright6323:
@alonamid, @jerryz123 may also want to take a look at the docs/comments that have been added since the last review

docs/Advanced-Concepts/Top-Testharness.rst
```diff
@@ -2,17 +2,28 @@ Tops, Test-Harnesses, and the Test-Driver
...
 
-The top-level module of a Rocket Chip SoC is composed via cake-pattern.
-Specifically, "Tops" extend a ``System``, which extends a ``Subsystem``, which extends a ``BaseSubsystem``.
+ChipTop is the top-level module that instantiates the ``System`` submodule, usually an instance of the concrete class ``DigitalTop``.
+The vast majority of the design resides in the ``System``.
+Other components that exist inside the ``ChipTop`` layer are generally IO cells, clock receivers and multiplexers, reset synchronizers, and other analog IP that needs to exist outside of the ``System``.
+The ``IOBinders`` are responsible for instantiating the IO cells and defining the test harness collateral that connects to the top-level ports.
+Most of these types of devices can be instantiated using custom ``IOBinders``, so the provided ``ChipTop`` and ``ChipTopCaughtReset`` classes are sufficient.
+However, if needed, the ``BaseChipTop`` abstract class can be extended for building more custom ``ChipTop`` designs.
+
```
jerryz123, line 20: Unrelated to this PR... we should probably create a diagram for all of this at some point.

docs/Customization/IOBinders.rst
```diff
@@ -1,19 +1,19 @@
...
 
 
-This special key solves the problem of duplicating test-harnesses for each different ``Top`` type.
+This special key solves the problem of duplicating test-harnesses for each different ``System`` type.
 
-You could just as well create a custom harness module that attaches IOs explicitly. Instead, the IOBinders key provides a map from Scala traits to attachment behaviors.
+You could just as well create a custom harness module that attaches IOs explicitly. Instead, the ``IOBinders`` key provides a map from Scala traits to attachment behaviors.
 
-For example, the ``WithSimAXIMemTiedOff`` IOBinder specifies that any ``Top`` which matches ``CanHaveMasterAXI4MemPortModuleImp`` will have a ``SimAXIMem`` connected.
+For example, the ``WithSimAXIMemTiedOff`` IOBinder specifies that any ``System`` which matches ``CanHaveMasterAXI4MemPortModuleImp`` will have a ``SimAXIMem`` connected.
```
jerryz123, line 16: I think we need a bit more text here explaining how the IOBinders generate IO ports, IOCells, and harness-functions which generate stuff in the TestHarness.

docs/Advanced-Concepts/Top-Testharness.rst
jwright6323: Agree, filed #502

generators/chipyard/src/main/scala/ChipTop.scala
```diff
@@ -0,0 +1,102 @@
...
+  harnessFunctions += { (th: TestHarness) => {
+    // Connect clock; it's not done implicitly with RawModule
+    clock := th.clock
+    // Connect reset; it's not done implicitly with RawModule
+    // Note that we need to use dutReset, not harnessReset
+    areset := th.dutReset
+    Nil
+  } }
+
+}
```
jerryz123, line 96: Is there a way to use the Config system to specify whether to use ChipTop or ChipTopCaughtReset? This smells similar to the movement towards parameterized traits across all our projects. The duplicated code across these two traits seems especially suspicious.

generators/chipyard/src/main/scala/ConfigFragments.scala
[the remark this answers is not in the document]
jerryz123: I think its worth devising a way to avoid manual setting of TOP. Specifically, I think we should try to have just one default "ChipTop" class, similar to how we have just one default "DigitalTop" (System) class.

generators/chipyard/src/main/scala/ChipTop.scala
jwright6323: We could make a `ChipTopReset` Field and just change behavior inside of `HasChipTopSimpleClockAndReset`, but I did want an example of creating your own `ChipTop`. Probably not a big deal either way, but I agree having to change `TOP` is super annoying. Once @colinschmidt or whoever is able to migrate barstools to stage/phase, we can make it easy to have `TOP` be an annotation input rather than a command line flag (related to @abejgonzalez's comment above)

generators/chipyard/src/main/scala/ConfigFragments.scala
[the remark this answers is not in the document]
jwright6323: OK, see comment above. For this PR we can do that, but I think we do want to make `TOP` programmatically generated anyway, which makes this less painful.

generators/chipyard/src/main/scala/ChipTop.scala
jerryz123: That's a good point. We can just leave these as two separate traits then, for educational purposes.

jwright6323: OK, sounds good. I actually did make the change and pushed to the `iocell_notwotops` branch if you want to look at it.

docs/Customization/IOBinders.rst
jwright6323: Added some docs

docs/Customization/IOBinders.rst
```diff
@@ -1,19 +1,35 @@
...
    :end-before: DOC include end: IOBinders
 
 
-This special key solves the problem of duplicating test-harnesses for each different ``Top`` type.
+This special key solves the problem of duplicating test-harnesses for each different ``System`` type.
+You could just as well create a custom harness module that attaches IOs explicitly.
+Instead, the ``IOBinders`` key provides a map from Scala traits to attachment behaviors.
+Each ``IOBinder`` returns a tuple of three values: the list of ``ChipTop`` ports created by the ``IOBinder``, the list of all IO cell modules instantiated by the ``IOBinder``, and an optional function to be called inside the test harness.
+This function is responsible for instantiating logic inside the ``TestHarness`` to appropriately drive the ``ChipTop`` IO ports created by the ``IOBinder``.
+Conviently, because the ``IOBinder`` is generating the port, it may also use the port inside this function, which prevents the ``BaseChipTop`` code from ever needing to access the port ``val``.
```
jerryz123, line 17:
```suggestion
Conveniently, because the ``IOBinder`` is generating the port, it may also use the port inside this function, which prevents the ``BaseChipTop`` code from ever needing to access the port ``val``.
```

jwright6323: oops

docs/Advanced-Concepts/Top-Testharness.rst
```diff
@@ -2,17 +2,28 @@ Tops, Test-Harnesses, and the Test-Driver
...
 file in Rocket Chip.
 
 
-Top/DUT
+ChipTop/DUT
 -------------------------
 
-The top-level module of a Rocket Chip SoC is composed via cake-pattern.
-Specifically, "Tops" extend a ``System``, which extends a ``Subsystem``, which extends a ``BaseSubsystem``.
+ChipTop is the top-level module that instantiates the ``System`` submodule, usually an instance of the concrete class ``DigitalTop``.
```
abejgonzalez, line 14: Nit: ChipTop here should be in code formatting ```ChipTop```

jwright6323, line 14:
```suggestion
``ChipTop`` is the top-level module that instantiates the ``System`` submodule, usually an instance of the concrete class ``DigitalTop``.
```

docs/Customization/IOBinders.rst
```diff
@@ -1,19 +1,35 @@
...
+This function is responsible for instantiating logic inside the ``TestHarness`` to appropriately drive the ``ChipTop`` IO ports created by the ``IOBinder``.
+Conveniently, because the ``IOBinder`` is generating the port, it may also use the port inside this function, which prevents the ``BaseChipTop`` code from ever needing to access the port ``val``.
+This scheme prevents the need to have two separate binder functions for each ``System`` trait.
+When creating custom ``IOBinders`` it is important to use ``suggestName`` to name ports; otherwise Chisel will raise an exception trying to name the IOs.
+The example ``IOBinders`` demonstrate this.
 
-You could just as well create a custom harness module that attaches IOs explicitly. Instead, the IOBinders key provides a map from Scala traits to attachment behaviors.
+As an example, the ``WithGPIOTiedOff`` IOBinder creats IO cells for the GPIO module(s) instantiated in the ``System``, then punches out new ``Analog`` ports for each one.
```
abejgonzalez, line 22:
```suggestion
As an example, the ``WithGPIOTiedOff`` IOBinder creates IO cells for the GPIO module(s) instantiated in the ``System``, then punches out new ``Analog`` ports for each one.
```

docs/Customization/IOBinders.rst
```diff
@@ -1,19 +1,35 @@
...
    :end-before: DOC include end: IOBinders
 
 
-This special key solves the problem of duplicating test-harnesses for each different ``Top`` type.
+This special key solves the problem of duplicating test-harnesses for each different ``System`` type.
+You could just as well create a custom harness module that attaches IOs explicitly.
+Instead, the ``IOBinders`` key provides a map from Scala traits to attachment behaviors.
+Each ``IOBinder`` returns a tuple of three values: the list of ``ChipTop`` ports created by the ``IOBinder``, the list of all IO cell modules instantiated by the ``IOBinder``, and an optional function to be called inside the test harness.
+This function is responsible for instantiating logic inside the ``TestHarness`` to appropriately drive the ``ChipTop`` IO ports created by the ``IOBinder``.
+Conveniently, because the ``IOBinder`` is generating the port, it may also use the port inside this function, which prevents the ``BaseChipTop`` code from ever needing to access the port ``val``.
```
abejgonzalez, line 17: Is accessing the `val` in the IOBinders just so that you can keep `BaseChipTop` clean (without specific modifications)? If this is true maybe we should add something like this to the sentence.

```suggestion
Conveniently, because the ``IOBinder`` is generating the port, it may also use the port inside this function, which prevents the ``BaseChipTop`` code from ever needing to access the port ``val``, thus having the ``IOBinder`` house all port specific code.
```

generators/chipyard/src/main/scala/ChipTop.scala
```diff
@@ -0,0 +1,102 @@
...
+trait HasChipTopSimpleClockAndReset { this: BaseChipTop =>
+
+  val (clock, systemClockIO) = IOCell.generateIOFromSignal(systemClock, Some("iocell_clock"))
+  val (reset, systemResetIO) = IOCell.generateIOFromSignal(systemReset, Some("iocell_reset"))
+
+  iocells ++= systemClockIO
+  iocells ++= systemResetIO
+
+  // Add a TestHarnessFunction that connects clock and reset
+  harnessFunctions += { (th: TestHarness) => {
```
davidbiancolin, line 60: Hmm, i would have though the extra curly braces wouldn't be necessary here.

generators/chipyard/src/main/scala/ChipTop.scala
```diff
@@ -0,0 +1,102 @@
...
+  val systemClock = Wire(Input(Clock()))
+  // The system reset (synchronous to clock)
+  val systemReset = Wire(Input(Bool()))
+
+  // The system module specified by BuildSystem
+  val system = withClockAndReset(systemClock, systemReset) { p(BuildSystem)(p) }
+
+  // Call all of the IOBinders and provide them with a default clock and reset
+  withClockAndReset(systemClock, systemReset) {
+    val (_ports, _iocells, _harnessFunctions) = p(IOBinders).values.map(_(system)).flatten.unzip3
```
davidbiancolin, line 38: In the future,  you can replace this map().flatten with flatMap. I've done this in my bump.

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -65,74 +77,146 @@ class ComposeIOBinder[T](fn: => (Clock, Bool, Bool, T) => Seq[Any])(implicit tag
...
+      val lineSize = system.p(CacheBlockBytes)
       (io zip node.in).foreach { case (axi4, (_, edge)) =>
         val mem = Module(new SimDRAM(memSize, lineSize, edge.bundle))
         mem.io.axi <> axi4
-        mem.io.clock := clock
-        mem.io.reset := reset
+        mem.io.clock := system.clock
+        mem.io.reset := system.reset
       }
     }; Nil
```
davidbiancolin, line 196: Nit: Drop the semicolon and put the Nil on another line, unless you're doing a one-liner.

generators/chipyard/src/main/scala/ChipTop.scala
jwright6323: Ah, yep. Cool

generators/chipyard/src/main/scala/IOBinders.scala
jwright6323: Yeah, I think it was just a copy/paste artifact. I will fix it if I end up making another commit, otherwise if you want to fix it in your PR that would help.

generators/chipyard/src/main/scala/ChipTop.scala
jwright6323: let me try it actually. IIRC it failed.

jwright6323: Yeah, it complains:
```
[error] /tools/B/johnwright/chipyard/generators/chipyard/src/main/scala/ChipTop.scala:60:41: ';' expected but '=>' found.
[error]   harnessFunctions += (th: TestHarness) => {
[error]                                         ^
[error] /tools/B/johnwright/chipyard/generators/chipyard/src/main/scala/ChipTop.scala:87:41: ';' expected but '=>' found.
[error]   harnessFunctions += (th: TestHarness) => {
[error]                                         ^
[error] two errors found
```

PR #485, opened 2020-03-17
[docs/ci] cleanup docs and add ci to check it

### LINKED ISSUE #483: CI for Documentation
**Impact**: other

**Description**
Sanity check all documentation changes to make sure things don't break from release to release.

**What is a motivating example for changing the behavior?**
Chipyard 1.2 accidentally broke includes in the documentation.

abejgonzalez:
**Related issue**: Fixes #483 

**Impact**: other

**Release Notes**
This runs CI on the documentation to make sure that documentation changes still build properly (very strict checking and errors as soon as possible).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -145,6 +145,17 @@ jobs:
                 name: Check that the tutorial-setup patches apply
                 command: |
                     scripts/tutorial-setup.sh
+    documentation-check:
+        executor: main-env
+        steps:
+            - checkout
+            - run:
+                name: Check that documentation builds with no warnings/errors
+                command: |
+                    sudo apt-get update -y
+                    sudo apt-get install -y python3-pip
+                    sudo pip3 install -r docs/requirements.txt
+                    make -C docs html
 
     install-riscv-toolchain:
         executor: main-env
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -145,6 +145,14 @@ jobs:
...
                 command: |
                     scripts/tutorial-setup.sh
+    documentation-check:
+        executor: main-env
+        steps:
+            - checkout
+            - run:
+                name: Check that documentation builds with no warnings/errors
+                command: |
+                    make -C docs html
```
zhemao: Are you sure this actually gives a failing error code if there are only warnings?

abejgonzalez: Yes because that's how I fixed the other documentation errors (missing references to labels).

docs/Makefile
```diff
@@ -2,8 +2,8 @@
 #
 
 # You can set these variables from the command line.
-SPHINXOPTS    = -w warnings.txt
-SPHINXBUILD   = python -msphinx
+SPHINXOPTS    = -w warnings.txt -n -W
```
abejgonzalez: Here the `-n` makes sure that links are correct and the `-W` converts all warnings to errors. If the `-W` is not added then the errors that @zhemao found in the X PR would be ignored.

PR #489, opened 2020-03-19
[WIP][dromajo] Automate dromajo-enabled simulations

jerryz123:
This PR integrates Dromajo like dramsim, based off Abe's work, and seamlessly enables dromajo-enabled simulations. The changes are mildly invasive, but the additional flags should have no effect on non-dromajo simulations.

@zhemao can you take a look at the dramsim flag change here? I think the flag wasn't being used everywhere before. 

The plan is to merge dromajo/boom work into the sboom-release branch, and we will merge that branch into `dev` closer to a potential 1.3.0 release.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -119,3 +119,6 @@
 [submodule "tools/DRAMSim2"]
 	path = tools/DRAMSim2
 	url = https://github.com/firesim/DRAMSim2.git
+[submodule "tools/dromajo"]
+	path = tools/dromajo
+	url = https://github.com/abejgonzalez/dromajo.git
--- a/common.mk
+++ b/common.mk
@@ -41,13 +41,15 @@ $(sim_files): $(call lookup_scala_srcs,$(base_dir)/generators/utilities/src/main
 # create firrtl file rule and variables
 #########################################################################################
 .INTERMEDIATE: generator_temp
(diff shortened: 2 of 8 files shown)


abejgonzalez:
While I think this PR does all the right things to make this work... I disagree with adding more things to the Makefile that 99.999% of people will not use. This will just add to the clutter of the Makefile and add bloat down the line (I share the same opinion on the `tracegen` and `axe` stuff that is taking up space at the bottom of the `common.mk`).

Fundamentally, this is a problem with the simulation flags / build / etc depending on what is elaborated. I think a step in the right direction is to emit a `Makefrag` that is added to the `Makefile` that sets relevant simulation flags / sets up the rest of the build process. However, this also will be harder to understand since there will be an extra level of indirection (and will make you 1st have to elaborate the Verilog... then finish building the simulator binary... which is IMO kinda ugly).

Additionally, I don't know if `wake` is nice and allows us to deal with this easily (having downstream collateral (i.e. simulators) builds change on different collateral emitted). Maybe they have a concept of `lazy` including the `makefrag` so that you don't have to split up the build in 2? Since we are eventually planning on supporting that soon, I think we should hold off on adding more than what is necessary to the make system.

jerryz123:
Closing since it looks like this is too invasive

PR #496, opened 2020-03-20
[RFC] Separate project make targets/variables from Makefile

abejgonzalez:
**Related issue**: 

**Impact**: other

**Release Notes**
This is the first attempt to have sub-project Makefile changes separated from the main makefile. It allows `mk` fragments in subprojects to inject variables/targets into the various parts of the makefile.

Note: This is not targeting `dev` so this will be merged in at a later point.


common.mk
```diff
@@ -45,7 +51,8 @@ $(FIRRTL_FILE) $(ANNO_FILE): generator_temp
 	@echo "" > /dev/null
 
 # AG: must re-elaborate if ariane sources have changed... otherwise just run firrtl compile
-generator_temp: $(SCALA_SOURCES) $(ARIANE_VLOG_SOURCES) $(sim_files)
+generator_temp: $(SCALA_SOURCES) $(PROJECT_GENERATOR_SOURCES) $(sim_files)
+	echo $(PROJECT_GENERATOR_SOURCES)
```
abejgonzalez: Ignore these `echo`s. They are just for debugging purposes. I will remove them when this is ready to merge.

common.mk
```diff
@@ -45,7 +51,8 @@ $(FIRRTL_FILE) $(ANNO_FILE): generator_temp
 	@echo "" > /dev/null
 
 # AG: must re-elaborate if ariane sources have changed... otherwise just run firrtl compile
-generator_temp: $(SCALA_SOURCES) $(ARIANE_VLOG_SOURCES) $(sim_files)
+generator_temp: $(SCALA_SOURCES) $(PROJECT_GENERATOR_SOURCES) $(sim_files)
```
jerryz123: I think the `PROJECT_X` naming scheme for the additional flags can easily be confused with the `PROJECT` and `SUBPROJECT` make variables, which are unrelated.

What about `MISC_GENERATOR_SOURCES` or `EXTRA_GENERATOR_SOURCES`

common.mk
```diff
@@ -3,6 +3,13 @@
 #########################################################################################
 SHELL=/bin/bash
 
+#########################################################################################
+# specific make variables/rules from subprojects
+#########################################################################################
+include $(base_dir)/generators/ariane/ariane.mk
+include $(base_dir)/generators/tracegen/tracegen.mk
+include $(base_dir)/tools/dromajo/dromajo.mk
+
```
jerryz123: I get that making this change against the dromajo PR is done to show why this splitting is useful, but I'm still apprehensive about making so many changes in a chipyard feature branch. If its not too much work, and people like these changes, can we start with a PR into the dev branch that does this restructuring for just ariane and tracegen, and merge in dromajo support later?

abejgonzalez: Sure, we can split this up... I just want to show how this might be done nicely with something that needs it more (i.e. Dromajo). If people agree on this approach then we can migrate it to `dev`.

common.mk
abejgonzalez: Agreed. I think we can do `EXTRA`.

common.mk
davidbiancolin: so what happens when these projects settings conflict?

davidbiancolin: this is strictly better than what we had before though.

abejgonzalez: Currently, everything is additive (see the `+=` for all the flags in their respective `mk`s). However, I haven't put much thought into how we deal with conflicts (for ex. when `+drj_bin=` is defined twice).

abejgonzalez: I know that the simulator will probably error... but maybe having some sort of check in `common.mk` would be nice? Maybe warn if full words match between `mk`s for the SIM_FLAGS variable (i.e. so that `+arg +arg` is caught)?

common.mk
```diff
@@ -129,10 +139,11 @@ $(output_dir)/%: $(RISCV)/riscv64-unknown-elf/share/riscv-tests/isa/%
 	ln -sf $< $@
 
 $(output_dir)/%.run: $(output_dir)/% $(sim)
-	(set -o pipefail && $(sim) $(PERMISSIVE_ON) +max-cycles=$(timeout_cycles) $(SIM_FLAGS) $(PERMISSIVE_OFF) $< </dev/null | tee $<.log) && touch $@
+	echo $(PROJECT_SIM_FLAGS)
+	(set -o pipefail && $(sim) $(PERMISSIVE_ON) +max-cycles=$(timeout_cycles) $(SIM_FLAGS) $(PROJECT_SIM_FLAGS) $(PERMISSIVE_OFF) $< </dev/null | tee $<.log) && touch $@
```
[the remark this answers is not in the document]
jerryz123: I brought this up to @zhemao to fix in chipyard/dev. We can roll that into this PR, since we should unify usage of SIM_FLAGS

common.mk
```diff
@@ -129,10 +139,11 @@ $(output_dir)/%: $(RISCV)/riscv64-unknown-elf/share/riscv-tests/isa/%
...
 
 $(output_dir)/%.out: $(output_dir)/% $(sim)
-	(set -o pipefail && $(sim) $(PERMISSIVE_ON) +dramsim +max-cycles=$(timeout_cycles) $(VERBOSE_FLAGS) $(PERMISSIVE_OFF) $< </dev/null 2> >(spike-dasm > $@) | tee $<.log)
+	(set -o pipefail && $(sim) $(PERMISSIVE_ON) +dramsim +max-cycles=$(timeout_cycles) $(VERBOSE_FLAGS) $(SIM_FLAGS) $(PROJECT_SIM_FLAGS) $(PERMISSIVE_OFF) $< </dev/null 2> >(spike-dasm > $@) | tee $<.log)
```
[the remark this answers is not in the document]
abejgonzalez: I don't know what `dramsim` does but I think we should keep these two separate. 

In many cases, I have to add extra flags to the simulation with `SIM_FLAGS` so having these both move there will make it so that adding new variables also requires adding the `+dramsim +max-cycles` flags as well as `+abe-flag=1`.

A compromise that we can do is have the `SIM_FLAGS` use `override` and just add the `+dramsim +max-cycles` always? Thoughts?

generators/tracegen/tracegen.mk
```diff
@@ -0,0 +1,33 @@
...
+
+$(output_dir)/tracegen.result: $(output_dir)/tracegen.out $(AXE)
+	$(base_dir)/scripts/check-tracegen.sh $< > $@
+
+.PHONY: tracegen
+tracegen: $(output_dir)/tracegen.result
+
+##################################################################
+# THE FOLLOWING MUST BE += operators
+##################################################################
```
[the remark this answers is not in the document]
abejgonzalez: I agree with you now. If these flags change (like a name change) then this will just break unnecessarily.

tools/dromajo/dromajo.mk
```diff
@@ -0,0 +1,51 @@
...
+##################################################################
+
+# sourced used to run the generator
+PROJECT_GENERATOR_SOURCES +=
+
+# simargs needed (i.e. like +drj_test=hello)
+PROJECT_SIM_FLAGS += $(DROMAJO_FLAGS)
+
+# extra vcs compile flags
+PROJECT_VCS_FLAGS += -CC "-I$(DROMAJO_DIR)" $(DROMAJO_LIB)
```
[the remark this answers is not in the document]
abejgonzalez: Ok how does this look now?

sims/vcs/Makefile
```diff
@@ -50,7 +50,8 @@ VCS_CC_OPTS = \
 	-CC "-I$(dramsim_dir)" \
 	-CC "-std=c++11" \
 	$(dramsim_lib) \
-	$(RISCV)/lib/libfesvr.a
+	$(RISCV)/lib/libfesvr.a \
+	-CC "$(EXTRA_SIM_CC_FLAGS)"
```
colinschmidt: Does this work if there are more than 1 extra flags?

common.mk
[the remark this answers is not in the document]
colinschmidt: Yeah override is the solution for this.
`override SIM_FLAGS += +dramsim +max-cycles`

tools/dromajo/dromajo.mk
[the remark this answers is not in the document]
colinschmidt: Looks good aside from the question I had about multiple `EXTRA_SIM_CC_FLAGS`.

sims/vcs/Makefile
abejgonzalez: Yup. If there are extra flags in the `EXTRA_SIM_CC_FLAGS` this works.

common.mk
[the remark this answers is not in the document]
abejgonzalez: Yup Ill change it to that then

sims/vcs/Makefile
colinschmidt: Are you sure that vcs does the right thing w.r.t. `-CC "-Iblah -Iblah2"`

abejgonzalez: I just tried the `-CC "-I$(VCS_HOME)/include -I$(RISCV)/include -I$(dramsim_dir) -std=c++11"` on one line and that worked.

abejgonzalez:
Closing since this is split up into #499 and later into the `sboom-release` branch.

PR #499, opened 2020-03-25
[make] split up specific make vars/targets into frags

abejgonzalez:
**Related issue**: #496 

**Type of change**: new feature/other enhancement

**Impact**: other

**Release Notes**
This changes the make system to have project-specific changes be put into a special `mk` file instead of being in the Chipyard main `mk`s. This allows for sub-projects to inject requirements, sources, flags and more into the build system using the `EXTRA_*` variables. This also adds `dramsim` as a default flag to all simulations. Finally, this also changes Verilators LDFLAGS to use `.a` libraries instead of the long thing that Rocket Chip uses. 

To see how this affects things with a more in-depth example see #496


common.mk
```diff
@@ -3,6 +3,18 @@
 #########################################################################################
 SHELL=/bin/bash
 
+#########################################################################################
+# extra make variables/rules from subprojects
+#
+# EXTRA_GENERATOR_REQS - requirements needed for the main generator
```
jerryz123: This is currently unused, I guess this is for dromajo?

abejgonzalez: `GENERATOR_REQS` is for Ariane

generators/tracegen/tracegen.mk
```diff
@@ -0,0 +1,18 @@
...
+##############################################################
+
+AXE_DIR=$(base_dir)/tools/axe/src
+AXE=$(AXE_DIR)/axe
+
+$(AXE): $(wildcard $(AXE_DIR)/*.[ch]) $(AXE_DIR)/make.sh
+	cd $(AXE_DIR) && ./make.sh
+
+$(output_dir)/tracegen.out: $(sim)
+	mkdir -p $(output_dir) && $(sim) $(PERMISSIVE_ON) +dramsim +max-cycles=$(timeout_cycles) $(SIM_FLAGS) $(EXTRA_SIM_FLAGS) $(VERBOSE_FLAGS) $(PERMISSIVE_OFF) none </dev/null 2> $@
```
jerryz123: Why does this still use +dramsim and +max-cycles?

abejgonzalez: I missed this. I only find/replaced in the `common.mk` and `Makefiles`

jerryz123:
I like this change a lot, but I think someone else should sign-off on this

PR #501, opened 2020-04-01
Sboom dromajo bump

abejgonzalez:
**Related issue**: 

**Impact**: rtl change | software change | other

**Release Notes**
This adds the Dromajo bridge to the `sboom` branch and has it updated with the multi-clock changes.


abejgonzalez:
Running target and MIDAS level tests as we speak

jerryz123:
Can you move dromajo.mk into your dromajo fork? I don't really like the `tools/dromajo/dromajo/src` structure.

abejgonzalez:
> Can you move dromajo.mk into your dromajo fork? I don't really like the `tools/dromajo/dromajo/src` structure.

If we ever try to push things back to upstream Dromajo, I would make sense to not have the `mk` I the fork. I would prefer to keep main Dromajo as "clean" as possible. Why in particular do you not like the extra folder?

abejgonzalez:
Ok both MIDAS and target level tests seem to work. So this should be good to review @jerryz123

PR #503, opened 2020-04-01
Stage / Phase RC Bump 

davidbiancolin:
This will be squashed before preforming the FireSim submodule chicanery. This is ready for review and should be merged ASAP due to the number of submodules it touches.

Associated PRs:
- riscv-boom/riscv-boom#453
- firesim/firesim#532
- ucb-bar/testchipip#76
- ucb-bar/barstools#79
- ucb-bar/ariane-wrapper#3
- ucb-bar/hwacha#18
- firesim/icenet#23

Discussion below:

Here are the issues i'm running into that require some thought to port.

Challenges: 
1) Reset inference and IOCells.
  - DebugIO has two abstract resets. We need to decide how we wish to concretize these in our generator, before (or as part of) IOcell generation, as we cannot emit IOCell BBs with abstract resets for outputs.
  - Workaround: temporarily skip emission of IOCells for DebugIO 
  - Proposed solution: IOCell generation should croak on abstract resets. Add a case for AsyncReset, which uses the same underlying verilog but placates reset inference.

2) Test-suite generation:
  - Test-suite generation phases as they currently exist within rocket chip are not extensible 
  - Workaround: copy relevant rocket chip stage sources into chipyard and manually add our own phase for this purpose
  - Proposed solution A: Let the generator emit annotations calling for specific test suites instead of having a phase to do so. Collect these in a later phase. 
-  Proposed solution B: Make it possible to add additional test suites in the existing phase / make it possible to run our own phase which does this.

3) Generator command-line API
  - The new rocket-chip phase removes the "_" delimited configs in favor of comma-delimited ones
  - Workaround: We're copying stage sources into chipyard anyways, so in our stage let's just include the legacy mode.
- Proposed solution: -Upstream a --legacy-config -LC to rocketchip 

TODOs:
- [x] Rectify make variable hacks to maintain existing make API 
- [x] Fix longname


jwright6323:
- [ ] Just noting here that when we go to switch `IOBinders` to matching on the `LazyModule` instead of the `Module`, we should put the `system` `suggestName` inside `ChipTop` (#480)

generators/chipyard/src/main/scala/Subsystem.scala
```diff
@@ -52,25 +52,37 @@ trait HasChipyardTiles extends HasTiles
...
+      //    val t = LazyModule(new ArianeTile(a, crossing, PriorityMuxHartIdFromSeq(arianeTileParams), logicalTreeNode))
+      //    (t, t.rocketLogicalTree) // TODO FIX rocketLogicalTree is not a member of the superclass, both child classes define it separately
+      //  }
+      //}
       connectMasterPortsToSBus(tile, crossing)
       connectSlavePortsToCBus(tile, crossing)
 
-      def treeNode: RocketTileLogicalTreeNode = new RocketTileLogicalTreeNode(rocketLogicalTree.getOMInterruptTargets)
-      LogicalModuleTree.add(logicalTreeNode, rocketLogicalTree)
+      //def treeNode: RocketTileLogicalTreeNode = new RocketTileLogicalTreeNode(rocketLogicalTree.getOMInterruptTargets)
```
abejgonzalez, line 84: Why is this deleted?

davidbiancolin: We don't use the this OM so i think we should remove our support for it.

abejgonzalez:
@zhemao Why does IceNet need a `TestHarness` file? Is it still needed?

generators/chipyard/src/main/scala/Subsystem.scala
abejgonzalez: My impression when I last looked at it a while ago was that it was tied to the DTS. I'm not opposed to getting rid of support for it as long as it doesn't affect anything else.

variables.mk
```diff
@@ -86,7 +86,7 @@ CHIPYARD_FIRRTL_DIR = $(base_dir)/tools/firrtl
 #########################################################################################
 # names of various files needed to compile and run things
 #########################################################################################
-long_name = $(MODEL_PACKAGE).$(MODEL).$(CONFIG)
+long_name = $(CONFIG_PACKAGE).$(MODEL).$(CONFIG)
```
abejgonzalez, line 89: Random thought: Should we look to re-organize this into something like "MODEL.CONFIG_PACKAGE.CONFIG"? Just to have it be clear that the CONFIG comes from CONFIG_PACKAGE and the MODEL can be located somewhere else?

abejgonzalez: Also, should we start deprecating the support for the Hwacha generator @colinschmidt? Is there a use case for it to still exist?

generators/chipyard/src/main/scala/System.scala
```diff
@@ -38,8 +38,5 @@ class System(implicit p: Parameters) extends Subsystem
 class SystemModule[+L <: System](_outer: L) extends SubsystemModuleImp(_outer)
   with HasRTCModuleImp
   with HasExtInterruptsModuleImp
-  with CanHaveMasterAXI4MemPortModuleImp
```
abejgonzalez, line 41: Why is this deleted here? Am I missing something / where they are added?

variables.mk
davidbiancolin: I think that makes more sense.

generators/chipyard/src/main/scala/System.scala
davidbiancolin: There's no imp trait anymore; it's been replaced with an InModuleBody call

generators/chipyard/src/main/scala/Subsystem.scala
davidbiancolin: ill check that the generated DTS is the same

colinschmidt:
For the test suite generation I like the idea of emitting annotations, but not exactly sure how it would work out.
I also like `--legacy-config`

build.sbt
```diff
@@ -166,6 +166,8 @@ lazy val gemmini = (project in file("generators/gemmini"))
 lazy val tapeout = conditionalDependsOn(project in file("./tools/barstools/tapeout/"))
   .dependsOn(chisel_testers, chipyard)
   .settings(commonSettings)
+  // Biancolin Bump: How did things compile without this dependency previously?
+  .settings(libraryDependencies ++= Seq("io.github.daviddenton" %% "handlebars-scala-fork" % "2.3.0"))
```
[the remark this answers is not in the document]
jwright6323: @grebe might know

generators/chipyard/src/main/scala/ChipTop.scala
```diff
@@ -12,7 +12,7 @@ import chipyard.iobinders.{IOBinders, TestHarnessFunction, IOBinderTuple}
 
 import barstools.iocell.chisel._
 
-case object BuildSystem extends Field[Parameters => RawModule]((p: Parameters) => Module(LazyModule(new DigitalTop()(p)).suggestName("system").module))
+case object BuildSystem extends Field[Parameters => LazyModule]((p: Parameters) => LazyModule(new DigitalTop()(p)).suggestName("system"))
```
jwright6323, line 15: can we move the `suggestName` from here to `ChipTop` now?

generators/chipyard/src/main/scala/ChipTop.scala
```diff
@@ -31,11 +31,15 @@ abstract class BaseChipTop()(implicit val p: Parameters) extends RawModule with
   val systemReset = Wire(Input(Bool()))
 
   // The system module specified by BuildSystem
-  val system = withClockAndReset(systemClock, systemReset) { p(BuildSystem)(p) }
+  val lSystem = p(BuildSystem)(p)
```
jwright6323, line 34:
```suggestion
  val lSystem = p(BuildSystem)(p).suggestName("system")
```

generators/chipyard/src/main/scala/ChipTop.scala
jwright6323: Also, is there ever a case where someone would want to have a concrete `Module` instead of a `LazyModule` as the system? It's easy to go from `LazyModule` -> `RawModule` but the reverse  requires a bit of work. Maybe that's rare enough that it's fine.

variables.mk
jerryz123: I'm in favor of removing `SUB_PROJECT`.

generators/chipyard/src/main/scala/ChipTop.scala
```diff
@@ -31,11 +31,15 @@ abstract class BaseChipTop()(implicit val p: Parameters) extends RawModule with
...
+  val system = withClockAndReset(systemClock, systemReset) { Module(lSystem.module) }
 
   // Call all of the IOBinders and provide them with a default clock and reset
   withClockAndReset(systemClock, systemReset) {
-    val (_ports, _iocells, _harnessFunctions) = p(IOBinders).values.map(_(system)).flatten.unzip3
+    // Biancolin: I'm lazy and didn't want to port code, but this may be a feature...
```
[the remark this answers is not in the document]
jerryz123: :scream: despite how this makes me uncomfortable, the flexibility seems to be useful enough.

generators/chipyard/src/main/scala/ChipTop.scala
jerryz123: Why is the reverse hard? It should be pretty easy to create a empty `LazyModule` to wrap an arbitrary concrete `Module`

jwright6323: Eh, "Hard" is probably the wrong word. My point was that creating an entirely new class is more typing than 8 characters (`.module`).

generators/chipyard/src/main/scala/ChipTop.scala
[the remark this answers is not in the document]
davidbiancolin: I'm happy to add a check here which only calls it on the LazyModule if the PF is defined on both.

generators/chipyard/src/main/scala/ChipTop.scala
davidbiancolin: We could have both a buildTop and buildLazy top if we really wanted and delegate appropriately, or we could have the BuildTop function return `(Option[LazyModule], Module)`

jwright6323: `(Option[LazyModule], Module)` actually seems pretty good.

generators/chipyard/src/main/scala/ChipTop.scala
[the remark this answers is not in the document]
colinschmidt: Yeah I like that check. 👍

generators/chipyard/src/main/scala/Subsystem.scala
davidbiancolin: @abejgonzalez The DTS seems unaffected by this change

build.sbt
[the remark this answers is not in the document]
grebe: `conditionalDependsOn` discards the build.sbt info b/c sbt is dumb at handling subprojects (I think the problematic case is actually subprojects of subprojects. I forget. Rocketchip in my experience has been very difficult to get working without entirely dodging the problem). Chipyard's top level build.sbt needs to reproduce all the deps from subprojects that use `conditionalDependsOn`. This looks like something barstools needs

generators/chipyard/src/main/scala/ChipTop.scala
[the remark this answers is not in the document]
davidbiancolin: Ok so it's probably going to be like 10 - 15 lines of reflection to make this happen ... so i'll  write it in a later PR.

generators/chipyard/src/main/scala/ChipTop.scala
davidbiancolin: I think this might require more careful thought so i'll resolve it in a follow up PR.

generators/chipyard/src/main/scala/ChipTop.scala
davidbiancolin: I've applied this to a local change i'll push tomorrow.

generators/chipyard/src/main/scala/Subsystem.scala
hcook: They're (unfortunately) parallel implementations that share some common infrastructure. If anything, OM builds/relies on the `Device` snippets used by DTS, not vice versa.

variables.mk
abejgonzalez: Likewise at the beginning it made sense. Now less so.

generators/chipyard/src/main/scala/Subsystem.scala
abejgonzalez: Thanks for the clarification

variables.mk
abejgonzalez: @colinschmidt Thoughts on the Hwacha generator? If you don't use it then we can just use the Chipyard generator and get rid of the "SUB_PROJECT=hwacha"

colinschmidt: It needs to provide feature parity.

davidbiancolin: Unless i screwed up porting the Makefrag generation, it should now.

colinschmidt: Looks reasonable to me. I'll test after the merge and fix it if its broken.

generators/chipyard/src/main/scala/ChipTop.scala
jwright6323:
```suggestion
case object BuildSystem extends Field[Parameters => LazyModule]((p: Parameters) => LazyModule(new DigitalTop()(p)))
```

generators/chipyard/src/main/scala/ChipTop.scala
```diff
@@ -26,16 +61,21 @@ abstract class BaseChipTop()(implicit val p: Parameters) extends RawModule with
...
+  // These are given so that IOCell can use DataMirror and generate ports with
+  // the right flow (Input/Output)
   val systemClock = Wire(Input(Clock()))
-  // The system reset (synchronous to clock)
-  val systemReset = Wire(Input(Bool()))
+  val systemReset = Wire(Input(Reset()))
 
   // The system module specified by BuildSystem
-  val system = withClockAndReset(systemClock, systemReset) { p(BuildSystem)(p) }
+  val lSystem = p(BuildSystem)(p).suggestName("system")
```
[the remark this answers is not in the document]
jwright6323: @zhemao I think we wanted to get rid of the other one

generators/chipyard/src/main/scala/ChipTop.scala
```diff
@@ -45,13 +85,22 @@ abstract class BaseChipTop()(implicit val p: Parameters) extends RawModule with
 
 /**
  * A simple clock and reset implementation that punches out clock and reset ports with the same
- * names as the implicit clock and reset for standard Module classes. Reset is synchronous to
- * clock, which may not be a good idea to use for tapeouts.
+ * names as the implicit clock and reset for standard Module classes. Three basic reset schemes 
+ * a provided. See [[GlobalResetScheme]].
```
jwright6323, line 89:
```suggestion
 * are provided. See [[GlobalResetScheme]].
```

generators/chipyard/src/main/scala/ChipTop.scala
```diff
@@ -45,13 +85,22 @@ abstract class BaseChipTop()(implicit val p: Parameters) extends RawModule with
...
  */
 trait HasChipTopSimpleClockAndReset { this: BaseChipTop =>
 
   val (clock, systemClockIO) = IOCell.generateIOFromSignal(systemClock, Some("iocell_clock"))
-  val (reset, systemResetIO) = IOCell.generateIOFromSignal(systemReset, Some("iocell_reset"))
+  val (reset, systemResetIO) = p(GlobalResetSchemeKey) match {
+    case GlobalResetSynchronous  =>
+      IOCell.generateIOFromSignal(systemReset, Some("iocell_reset"))
+    case GlobalResetAsynchronousFull =>
+      IOCell.generateIOFromSignal(systemReset, Some("iocell_areset"), abstractResetAsAsync = true)
```
jwright6323, line 98:
```suggestion
      IOCell.generateIOFromSignal(systemReset, Some("iocell_reset"), abstractResetAsAsync = true)
```

Unless we're going to suggestName `reset` to `areset` we probably want to keep the IO cell naming convention of `iocell_ + <signal name>`

generators/chipyard/src/main/scala/ChipTop.scala
```diff
@@ -45,13 +85,22 @@ abstract class BaseChipTop()(implicit val p: Parameters) extends RawModule with
...
+    case GlobalResetAsynchronous =>
+      val asyncResetCore = Wire(Input(AsyncReset()))
+      systemReset := ResetCatchAndSync(systemClock, asyncResetCore.asBool)
+      IOCell.generateIOFromSignal(asyncResetCore, Some("iocell_areset"), abstractResetAsAsync = true)
```
jwright6323, line 102:
```suggestion
      IOCell.generateIOFromSignal(asyncResetCore, Some("iocell_reset"), abstractResetAsAsync = true)
```

See above

generators/chipyard/src/main/scala/ConfigFragments.scala
```diff
@@ -66,7 +66,7 @@ class WithL2TLBs(entries: Int) extends Config((site, here, up) => {
 })
 
 class WithTracegenSystem extends Config((site, here, up) => {
-  case BuildSystem => (p: Parameters) => Module(LazyModule(new tracegen.TraceGenSystem()(p)).suggestName("Top").module)
+  case BuildSystem => (p: Parameters) => LazyModule(new tracegen.TraceGenSystem()(p)).suggestName("Top")
```
jwright6323, line 69:
```suggestion
  case BuildSystem => (p: Parameters) => LazyModule(new tracegen.TraceGenSystem()(p))
```

generators/chipyard/src/main/scala/ChipTop.scala
davidbiancolin: sounds good.

abejgonzalez:
LGTM pending Ariane/BOOM comments.

PR #505, opened 2020-04-05
NVDLA Integration + Cleanup Ariane Preprocessing

abejgonzalez:
**Related issue**: 

**Impact**: rtl change | software change

**Release Notes**

*Chipyard*
This integrates `master` of NVDLA into Chipyard based off work done in (https://github.com/CSL-KU/nvidia-dla-blocks) and (https://github.com/sifive/block-nvdla-sifive). This builds in both VCS and Verilator. Tested with the NVDLA test put in `tests/`. This also changes Ariane to use the `insert-includes.py` script that replaces includes (instead of using Verilator).

*FireSim*
Added both `small`@75MHz and `large`@50MHz NVDLA to Rocket configs. Both `small` and `large` have completed AFI builds.

*Remaining Todos*
- [x] Add documentation
- [x] Run FPGA tests (run YOLO3, ResNet-50, `nvdla-sw` regressions)
- [x] Add to CI (basic test only)
- [x] Consider removing FireSim configs in favor of just keeping config fragments (i.e. `WithNVDLALarge` `WithNVDLASmall`) - Removed
- [x] Add recipe to FireSim for `small` and `large` (just added `small` unless FireSim devs want more)
- [x] Rename `opendla_small.h` to something more reasonable
- [x] Rename `HasPeripheryNVDLA` to `CanHave`
- [x] Change `.gitmodules` for `nvdla-workload` `nvdla-sw` to be `https`


sims/vcs/Makefile
```diff
@@ -55,18 +55,15 @@ VCS_CC_OPTS = \
 
 VCS_NONCC_OPTS = \
 	+lint=all,noVCDE,noONGS,noUI \
-	-error=PCWM-L \
```
abejgonzalez, line 58: This is a port width mismatch error that I am removing for NVDLA. Now it just warn instead of error. If people want it kept in, I can have it be conditionally removed if NVDLA is present (like with `assert` for Ariane).

abejgonzalez:
Thanks to @farzadfch for the help!

farzadfch:
Thanks for sharing the PR. A few points:
- For FPGA tests, we want to start from the basic tests in [here](https://github.com/nvdla/sw/tree/master/regression/flatbufs/kmd).
- The NVDLA driver is written for Linux 4.13.3. I hope that is ports smoothly to 5.3.
- The driver has been updated since I integrated NVDLA with FireSim. The DTS [here](https://github.com/ucb-bar/nvdla-wrapper/blob/db69836c9a310b40f97e9b8013aa4647f07f4a70/src/main/scala/devices/nvdla/NVDLA.scala#L27) is based on the older version. In the updated driver, the correct driver parameters are selected based on the DTS ([here](https://github.com/nvdla/sw/blob/79538ba1b52b040a4a4645f630e457fa01839e90/kmd/port/linux/nvdla_core_callbacks.c#L347) and [here](https://github.com/nvdla/sw/blob/79538ba1b52b040a4a4645f630e457fa01839e90/kmd/port/linux/nvdla_core_callbacks.c#L66)). So we should generate the correct DTS based on the configuration in the Chisel code.

tymcauley:
Agreed, thanks for sharing this! I've also done work recently on integrating the NVDLA into Chipyard/FireSim, so I'd love to contribute some of my learnings and help move this along.

For @farzadfch's third point, I generated the DTS compatibility string like this to integrate with the new kernel mode driver:

```scala
val dtsdevice = new SimpleDevice("nvdla",Seq("nvidia,nv_" + params.config))
```

For integrating the NVDLA Linux driver into the current FireSim Linux kernel (version 5.3), the process was not so bad, but changes were necessary. Some functions were deprecated, but have drop-in replacements, and some new `#include`s were needed (`struct` definitions moved around), but that was about the extent of it. I can give more specific input during the week 🙂.

I have a concern with the test `tests/nvdla.c`. It appears to be generated from the `.cfg` file in [this directory](https://github.com/nvdla/hw/tree/1a65f1f5b48268accaa47c95f95c2601918be095/verif/tests/trace_tests/nv_large/cdp_passthrough_8x8x32_pack_inc_int8) (one of the NVDLA hardware testbench files). However, `tests/nvdla.c` doesn't appear to do any correctness testing (such as feeding in the data from `CDP_0_input.dat` and comparing the results with the data in `CDP_0_output.dat`). It'd be great for the test to ensure correctness in the operation.

Furthermore, since the `tests/nvdla.c` test appears to be targeted at the the `nv_large` configuration, it would be great to describe a process for integrating tests for other NVDLA configurations, since [this directory](https://github.com/nvdla/hw/tree/1a65f1f5b48268accaa47c95f95c2601918be095/verif/tests/trace_tests) appears to contain tests for a number of configurations. And perhaps this test will work fine for other configurations, but it would still be valuable to make it clear how to develop other NVDLA tests to exercise other portions of the design.

I'm also curious to read through `scripts/insert-includes.py` a bit more this week, I had to jump through some hoops to get all of the RTL collateral to integrate with Chipyard/FireSim (I ended up adding a ton of `addResource` calls to `nvdla.scala`).

Regarding the [`nvdla_large.v`](https://github.com/ucb-bar/nvdla-wrapper/blob/db69836c9a310b40f97e9b8013aa4647f07f4a70/vsrc/large/nvdla_large.v) and [`nvdla_small.v`](https://github.com/ucb-bar/nvdla-wrapper/blob/db69836c9a310b40f97e9b8013aa4647f07f4a70/vsrc/small/nvdla_small.v) wrapper files, do we want to always tie together the `core_clk` and `csb_clk` inputs? I know this was done to accommodate the single-clock requirements for Black Boxes in FireSim. Is that requirement still in place? Also, could we design a system that ties these clocks together only when you're building the system for FireSim and not for an SoC?

Related FireSim vs SoC question: I haven't thought about this too much yet, but it would be good to look into the issue of selecting the NVDLA-provided [FPGA](https://github.com/ucb-bar/nvdla-wrapper/tree/db69836c9a310b40f97e9b8013aa4647f07f4a70/vsrc/large/vmod/rams/fpga/model) vs. [synthesis](https://github.com/ucb-bar/nvdla-wrapper/tree/db69836c9a310b40f97e9b8013aa4647f07f4a70/vsrc/large/vmod/rams/synth) RAM models for these two different use cases as well.

abejgonzalez:
Wow. Thanks for the quick feedback!

_Linux Testing:_
I haven't done any testing with Linux yet so that is the next step. @tymcauley would you be willing to share your changes/integration for the NVDLA Linux driver?

_CI Tests/Other Tests:_
As for the `nvdla.riscv` test, I merely copied it from @farzadfch's other work, so frankly I don't know anything about what it does. I think it would be a good idea to have a better test to make sure it is integrated properly however, I know very little about the tests in the NVDLA repo. I can try to look into some of the NVDLA tests when I have the chance but any recommendations/pointers are welcome!

_Insert Includes:_
The `scripts/insert-includes.py` is a very "dumb" script that just looks for includes and replaces them with the file it matches with. Normally, I would pre-process with Verilator then pass in the pre-processed output to the rest of the flow, however this removes any comments that may be needed later (i.e. `pragmas` and `*synopsys*`). So this script might be a reasonable band-aid? This script works for these two cases, but it can change if it needs to be more robust.

_Clock Requirements:_
Yes. I think the single-clock BB requirement is still in place (@davidbiancolin). We could potentially write the wrapper files (`nvdla_small/large.v`) in Chisel and pass in two different clocks. Thus the black boxes would be the inner modules (`NV_NVDLA_apb2csb1, `NV_nvdla`), not the wrapper module. I think that might make things work nicely with multi-clock support in FireSim. With that being said, I would like to keep extensive FireSim multi-clock changes out of this PR since we are still testing/developing that feature.

_FPGA vs SoC RAMs:_
I think this should be relatively easy. We can just pass a flag into the NVDLA preprocess Verilog makefile that switches between the two types of RAMs.

Please let me know if you guys have any other questions or concerns!

sims/vcs/Makefile
tymcauley: Ya, in general I've found that I've had to make edits to the NVDLA RTL files to fix some illegal `wire` declarations and errors of that sort. Also, I seem to recall this width-mismatch-error, and that it came from files that aren't actually needed by the this particular configuration of the design. Perhaps is was in the bridge-DMA (BDMA) module? That's not part of either `nv_large` or `nv_small`. Something to keep in mind in general about the NVDLA RTL: not all of the RTL files that are generated are needed by the design, since your configuration might omit some features. And some of those files that aren't needed by the design sometimes contain illegal RTL (such as these width mismatches, and some undeclared signals). So maybe we want to `git rm` problematic files from the `vsrc` directory?

Anyway, what was the specific width-mismatch error you were seeing? It might be due to this situation.

davidbiancolin:
It is now legal to have blackboxes with multiple clocks in FireSim, but i have not tested that case.

generators/chipyard/src/main/scala/ConfigFragments.scala
```diff
@@ -160,3 +154,12 @@ class WithControlCore extends Config((site, here, up) => {
 class WithChipTopCaughtReset extends Config((site, here, up) => {
   case BuildTop => (p: Parameters) => Module(new ChipTopCaughtReset()(p).suggestName("top"))
 })
+
+/**
+ * Config fragment to add a NVDLA to the SoC.
+ * Supports "small" and "large" configs only.
+ */
+class WithNVDLA(config: String) extends Config((site, here, up) => {
```
jerryz123, line 162: Nit, but this feels like it belongs in the NVDLA-wrapper repo, instead of here.

tymcauley: Ya, seems like a good idea to put this in a `Configs.scala` file in that repo.

tymcauley:
Here's what I did with the Linux driver:

* I started with the @farzadfch's [`firesim-nvdla` branch of `riscv-linux`](https://github.com/CSL-KU/riscv-linux/tree/afc1699e27a420f5c846bd36c0334f84d2dbe668)
* I then applied a few updates to the KMD from the [NVDLA software repo](https://github.com/nvdla/sw). I used these commits from that repo:
    * 2841094c12531252ad8b7f73a3635bbd13ca5858
    * 8df37971b901b623aaeba66e6c998f74a499c272
    * d01dfc8640b8664703047315e5a337602ab47d2d
    * 9deea34f2d61bd55abb8ec5648e7079d9665e220
    * 4022452c3124cdf1fcbb1124ce884187375a80ea
* Notably, I avoided including the changes from the following commits from the upstream `nvdla/sw` repo, as at least one of them introduced regressions in the KMD's behavior. I didn't do an exhaustive test to ensure which commit(s) were the problem though:
    * 16befcb84f8ccd2b3d1b0683b965375ae11274df
    * c906fc63eb0f20bc6ea6337fc7ef38647923339f
    * fae1eab705673fd63df03f4a927d6fda81e2a13b

At this point, I moved all of this work into the current Linux kernel fork in the FireMarshal repo. If you try to compile the Linux kernel, you'll run into a number of compilation errors. Here are the diffs I had to make to solve those errors:

* Add `#include <linux/uaccess.h>` to `include/nvdla_linux.h`
* The following changes to `nvdla_gem.c`:
    * Add `#include <drm/drm_device.h>`, `#include <drm/drm_drv.h>`, and `#include <linux/dma-mapping.h>`
    * Replace `drm_gem_object_unreference_unlocked(...)` with `drm_gem_object_put_unlocked(...)`
    * Replace `drm_dev_unref(...)` with `drm_dev_put(...)`
    * Change the `dma_declare_coherent_memory(...)` invocation to remove the final argument, `DMA_MEMORY_EXCLUSIVE`.

After making those changes, the driver compiled and functioned properly. I've only tested with the `nv_large` configuration, not `nv_small`. Also @farzadfch, I couldn't run your YOLOv3 workload yet, as there's an issue with the GLIBC headers used in the buildroot image. Haven't yet looked into how to resolve that issue.

sims/vcs/Makefile
abejgonzalez: I just re-ran VCS with the error re-enabled and I found the error in the large config when the `NV_NVDLA_CSC_dl` was instantiated in the `NV_NVDLA_csc` module.

```
"NV_NVDLA_CSC_dl u_dl( .nvdla_core_clk (nvdla_op_gated_clk_2),  .nvdla_core_rstn (nvdla_core_rstn),
  The following 10-bit expression is connected to 1-bit port
  "sc2buf_dat_rd_shift" of module "NV_NVDLA_CSC_dl", instance "u_dl".
  Expression: sc2buf_dat_rd_shift
  The expression is from module NV_NVDLA_csc, instance u_NV_NVDLA_csc
```

abejgonzalez: > Something to keep in mind in general about the NVDLA RTL: not all of the RTL files that are generated are needed by the design, since your configuration might omit some features. And some of those files that aren't needed by the design sometimes contain illegal RTL (such as these width mismatches, and some undeclared signals).

I also tried to prune all unnecessary files not needed to build the configs from the required files to reduce errors. So I believe the current set is the minimal workable set of requirements.

abejgonzalez:
Thanks @tymcauley Ill take a look at the Linux driver work. I also added a switch to the config to choose between fpga/synth rams. Let me know if that works for you.

sims/vcs/Makefile
tymcauley: Ah yes, I remember this error :)

This issue lies in `NV_NVDLA_CSC_dl.v`. Since the `CBUF_NO_SUPPORT_READ_JUMPING` macro is defined at the top of this file, the code on lines 3214 through 3216 is enabled. However, these three `wire` declarations with initializations are not legal, synthesizable verilog. This change should be made:

```diff
diff --git a/src/main/resources/vsrc/large/vmod/nvdla/csc/NV_NVDLA_CSC_dl.v b/src/main/resources/vsrc/large/vmod/nvdla/csc/NV_NVDLA_CSC_dl.v
index a2e2c1e..8f2cc21 100644
--- a/src/main/resources/vsrc/large/vmod/nvdla/csc/NV_NVDLA_CSC_dl.v
+++ b/src/main/resources/vsrc/large/vmod/nvdla/csc/NV_NVDLA_CSC_dl.v
@@ -3211,9 +3211,12 @@ assign dat_req_sub_h_1_addr_en = layer_st | ((dat_req_valid_d1 | dat_req_dummy_d
 assign dat_req_sub_h_2_addr_en = layer_st | ((dat_req_valid_d1 | dat_req_dummy_d1) & (dat_req_sub_h_d1 == 2'h2));
 assign dat_req_sub_h_3_addr_en = layer_st | ((dat_req_valid_d1 | dat_req_dummy_d1) & (dat_req_sub_h_d1 == 2'h3));
 `ifdef CBUF_NO_SUPPORT_READ_JUMPING
-wire sc2buf_dat_rd_next1_en = 1'b0;
-wire sc2buf_dat_rd_next1_en_w = 1'b0;
-wire sc2buf_dat_rd_shift = {10{1'b0}};
+wire sc2buf_dat_rd_next1_en;
+wire sc2buf_dat_rd_next1_en_w;
+wire [10 -1:0] sc2buf_dat_rd_shift;
+assign sc2buf_dat_rd_next1_en = 1'b0;
+assign sc2buf_dat_rd_next1_en_w = 1'b0;
+assign sc2buf_dat_rd_shift = {10{1'b0}};
 `endif
 `ifdef CBUF_SUPPORT_READ_JUMPING
 wire [10 -1:0] sc2buf_dat_rd_shift_w;
```

This same change is not necessary in the `nv_small` configuration, since while the same erroneous code exists, the relevant macro is not defined, so that code isn't enabled.

tymcauley:
> Thanks @tymcauley Ill take a look at the Linux driver work. I also added a switch to the config to choose between fpga/synth rams. Let me know if that works for you.

That switch looks great to me, thanks!

generators/firechip/src/main/scala/TargetConfigs.scala
```diff
@@ -63,23 +61,22 @@ class WithScalaTestFeatures extends Config((site, here, up) => {
...
-
 // Enables tracing on all cores
 class WithTraceIO extends Config((site, here, up) => {
   case BoomTilesKey => up(BoomTilesKey) map (tile => tile.copy(trace = true))
   case ArianeTilesKey => up(ArianeTilesKey) map (tile => tile.copy(trace = true))
   case TracePortKey => Some(TracePortParams())
 })
 
+class WithNVDLALarge extends chipyard.config.WithNVDLA("large")
+class WithNVDLASmall extends chipyard.config.WithNVDLA("small")
```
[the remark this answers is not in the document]
abejgonzalez: I think for you to make a build recipe like `WithNVDLALarge_SomeFireSimConfig` as the `TARGET_CONFIG` you need all of the components in the same scala package.

farzadfch:
>Also @farzadfch, I couldn't run your YOLOv3 workload yet, as there's an issue with the GLIBC headers used in the buildroot image. Haven't yet looked into how to resolve that issue.

What is the error? Does it complain about `libgomp.so.1`?

tymcauley:
> > Also @farzadfch, I couldn't run your YOLOv3 workload yet, as there's an issue with the GLIBC headers used in the buildroot image. Haven't yet looked into how to resolve that issue.
> 
> What is the error? Does it complain about `libgomp.so.1`?

No, although I did have to add `libgomp.so.1` to the `BR2_TOOLCHAIN_EXTRA_EXTERNAL_LIBS` line in `software/firemarshal/wlutil/br/buildroot-config`:

```diff
diff --git a/software/firemarshal/wlutil/br/buildroot-config b/software/firemarshal/wlutil/br/buildroot-config
index 69df8c3..54f3af4 100644
--- a/software/firemarshal/wlutil/br/buildroot-config
+++ b/software/firemarshal/wlutil/br/buildroot-config
@@ -717,7 +717,7 @@ BR2_TOOLCHAIN_EXTERNAL_INET_RPC=y
 BR2_TOOLCHAIN_EXTERNAL_PATH="$(RISCV)"
 BR2_TOOLCHAIN_EXTERNAL_PREFIX="$(ARCH)-unknown-linux-gnu"
 BR2_TOOLCHAIN_EXTERNAL_PREINSTALLED=y
-BR2_TOOLCHAIN_EXTRA_EXTERNAL_LIBS=""
+BR2_TOOLCHAIN_EXTRA_EXTERNAL_LIBS="libgomp.so.1"
 BR2_TOOLCHAIN_HAS_FULL_GETTEXT=y
 BR2_TOOLCHAIN_HAS_NATIVE_RPC=y
 BR2_TOOLCHAIN_HAS_SSP=y
```

Once I did that and ran `solo.sh`, I got this error:

```
# ./solo.sh
./darknet: /lib/libpthread.so.0: version `GLIBC_2.26' not found (required by ./darknet)
./darknet: /lib/libm.so.6: version `GLIBC_2.26' not found (required by ./darknet)
./darknet: /lib/libc.so.6: version `GLIBC_2.26' not found (required by ./darknet)
./darknet: /lib/libc.so.6: version `GLIBC_2.26' not found (required by /root/darknet-nvdla/libodlalayer.so)
./darknet: /lib/libpthread.so.0: version `GLIBC_2.26' not found (required by /root/darknet-nvdla/libdarknet.so)
./darknet: /lib/libm.so.6: version `GLIBC_2.26' not found (required by /root/darknet-nvdla/libdarknet.so)
./darknet: /lib/libc.so.6: version `GLIBC_2.26' not found (required by /root/darknet-nvdla/libdarknet.so)
./darknet: /lib/libm.so.6: version `GLIBC_2.26' not found (required by /root/darknet-nvdla/libnvdla_runtime.so)
./darknet: /lib/libc.so.6: version `GLIBC_2.26' not found (required by /root/darknet-nvdla/libnvdla_runtime.so)
```

farzadfch:
Perhaps we want to connect [this](https://github.com/ucb-bar/nvdla-wrapper/blob/43ae9612d1dfde0d3b46f0da486d384384e9bf00/src/main/scala/devices/nvdla/Periphery.scala#L21) to pbus instead of sbus? (to avoid adding a new port to the sbus Xbar)

farzadfch:
@tymcauley I think the libraries  (*.so) need to be recompiled to match the GLIBC version.

tymcauley:
> @tymcauley I think the libraries (*.so) need to be recompiled to match the GLIBC version.

Sounds good, I might not be able to get to that for a bit, but I'll try that out soon (hopefully this weekend).

Is there any way that we can run this YOLOv3 test on the nv_small configuration in addition to the nv_large? I wasn't sure if the "yolov3-subgraph{0,1,2}.nvdla" files were compatible with different NVDLA configurations.

abejgonzalez:
@farzadfch Did you ever get NVDLA working on `br-base`? I see that you modified the `fedora-base` `linux-config` to add the `CONFIG_DRM=y` and `CONFIG_DRM_GEM_CMA_HELPER=y` settings (and a lot others). When I add the same thing to the `br-base` `linux-config` it doesn't seem to work giving the following error:

```
ERROR: "drm_gem_cma_vm_ops" [/path-to-nvdla-sw/opendla.ko] undefined!
```

farzadfch:
@abejgonzalez Yes, in fact, I only tested on `br-base`. Adding `DRM_GEM_CMA_HELPER` is a bit tricky. If I remember correctly, I had to explicitly select it [here](https://github.com/CSL-KU/riscv-linux/blob/a774a2d7d196a69dc7149b13f45c23f36edfbd82/drivers/nvdla/Kconfig#L5) to be properly enabled.

I do not know how this should be done in a clean way when you compile as a module. Maybe you can create a dummy config in the Linux source that selects `DRM_GEM_CMA_HELPER`.

farzadfch:
@tymcauley I have not created the graph files but I believe that they are only compatible with the large config.

abejgonzalez:
As an update since it has been a while since I post something here.

So far, I think this RTL is working. I am trying to run YOLO3 using Darknet but I am getting a `rcu: INFO: rcu_sched self-detected stall on CPU` error? on a Large NVDLA configuration. I am currently building other loadables to test so Ill update when I get to that.

Here is the uartlog if people want to see it:

[uartlog.txt](https://github.com/ucb-bar/chipyard/files/4506981/uartlog.txt)

tymcauley:
Back this this point on the clocking idea...

> Yes. I think the single-clock BB requirement is still in place (@davidbiancolin). We could potentially write the wrapper files (`nvdla_small/large.v`) in Chisel and pass in two different clocks. Thus the black boxes would be the inner modules (`NV_NVDLA_apb2csb1`, `NV_nvdla`), not the wrapper module. I think that might make things work nicely with multi-clock support in FireSim. With that being said, I would like to keep extensive FireSim multi-clock changes out of this PR since we are still testing/developing that feature.

I like this idea a lot. The only problem module for FireSim in that case is `NV_nvdla`, since it has both the `core_clk` and `csb_clk` inputs.

Not sure if this is the best solution, but here's one idea: to keep compatibility with the single-clock-needs of FireSim, we could create a wrapper Verilog file which simply is a pass-through for `NV_nvdla`, but ties those two clocks together. We could create black-box definitions for both `NV_nvdla` and this new `NV_nvdla_single_clock` module in `nvdla.scala`. Then, in `NVDLA.scala`, we optionally use either `NV_nvdla` or `NV_nvdla_single_clock` as the top-level module. How does that sound? Any better ideas?

abejgonzalez:
That sounds reasonable to me. However, I am still not doing much for multi-clock support on this PR (yet).

Update # 2:
I'm still trying to get anything to work on the RTL (none of the regressions / YOLO3 / my ResNet50 loadable are working). I feel like this is a SW issue and not an RTL issue though. I am running off the latest `master` branch here https://github.com/abejgonzalez/sw and FireMarshal workload here https://github.com/abejgonzalez/nvdla-workload. The NVDLA SW is taken from the master NVDLA SW branch with modifications to work on the newest Kernel. In the following days, I'll try to use the direct recommendations from the comment with a lot of commits. 

If anyone has recommendations on which loadables "should" be good to run i.e. "`NN_L0_1_small_fbuf` with `Y` image finished for me", please let me know. 

Ill try to keep this thread updated on the situation.

abejgonzalez:
Update # 3:

Ok following (@ tymcauley's) advice on the driver, I narrowed down the problematic single commit and removed it. The rest of the SW is the same. I was able to run YOLO3 on NVDLA `large` as well as all the regressions (on `large` and `small`). I am still having trouble running ResNet50 but I'll start updating this to `dev` so we can get this PR in (since it is mostly working).

tymcauley:
> Update # 3:
> 
> Ok following (@ tymcauley's) advice on the driver, I narrowed down the problematic single commit and removed it. The rest of the SW is the same. I was able to run YOLO3 on NVDLA `large` as well as all the regressions (on `large` and `small`). I am still having trouble running ResNet50 but I'll start updating this to `dev` so we can get this PR in (since it is mostly working).

In my own work on the NVDLA, I just discovered why [this NVDLA SW commit](https://github.com/nvdla/sw/commit/16befcb84f8ccd2b3d1b0683b965375ae11274df) has been problematic. So, [the NVDLA Loadable format](https://github.com/nvdla/sw/blob/79538ba1b52b040a4a4645f630e457fa01839e90/umd/core/src/common/loadable.fbs) contains [a few blobs](https://github.com/nvdla/sw/blob/79538ba1b52b040a4a4645f630e457fa01839e90/umd/core/src/common/loadable.fbs#L56-L70) in it which describe useful data for the Kernel-Mode Driver (KMD). This includes the dependency graph of layer operations in the network, all of the network weights, the configuration options for each layer operation, look-up-table values, etc. Those blobs are just serialized C structs, and it turns out this commit adds a new field to one of those C structs ([this change in particular](https://github.com/nvdla/sw/commit/16befcb84f8ccd2b3d1b0683b965375ae11274df#diff-d637e5a407cb3c54c0632177ee38abbd)). Now, if you build/run the NVDLA compiler tool in the NVDLA SW repo (`umd/compiler`), it will produce an NVDLA Loadable that expects this new binary format. However, if you use one of the pre-compiled NVDLA Loadables (found at `regression/flatbufs/kmd/*/`), they _do not_ contain that field. As a result, all of the `dla_data_cube` structs deserialized by the KMD contain totally garbage data. You can view what those structs look like by changing [this define](https://github.com/nvdla/sw/blob/79538ba1b52b040a4a4645f630e457fa01839e90/kmd/firmware/include/dla_debug.h#L67) from `0` to `1` and ensuring [all of the `dla_*` logging functions](https://github.com/nvdla/sw/blob/79538ba1b52b040a4a4645f630e457fa01839e90/kmd/port/linux/nvdla_core_callbacks.c#L81-L111) are enabled (as I recall, they're commented out in the Linux kernel driver to reduce message traffic).

So, we appear to be in the position where we need to include that commit if we want to use an NVDLA Loadable generated by the NVDLA compiler, but we have to exclude that commit if we want to use one of their pre-generated unit test NVDLA Loadables. Not a great position, but it seems like including that commit is important if it means people can define their own workloads with the NVDLA compiler.

abejgonzalez:
Wow. Good catch, I was starting to trace why that commit was problematic so you beat me to the punch!

1. I just tried ResNet-50 with all the commits added back in and it works now. 
2. I'll go ahead and put and issue the `sw` repository since that seems like a major issue.

Everything here should be up to date with the latest `dev`. I'll start the review process. As for the multi-clock, I think that should move to a different PR/issue since this PR currently works in a single clock system.

abejgonzalez:
@farzadfch Feel free to comment on the PR. I can't add you as a reviewer.

generators/chipyard/src/main/scala/ConfigFragments.scala
```diff
@@ -80,12 +79,6 @@ class WithRenumberHarts(rocketFirst: Boolean = false) extends Config((site, here
   case MaxHartIdBits => log2Up(up(BoomTilesKey, site).size + up(RocketTilesKey, site).size)
 })
 
-
-
-// ------------------
-// Multi-RoCC Support
```
jerryz123, line 86: Was deleting this intentional?

README.md
```diff
@@ -65,3 +65,4 @@ These publications cover many of the internal components used in Chipyard. Howev
 [firemarshal]: https://github.com/firesim/FireMarshal/
 [ariane]: https://github.com/pulp-platform/ariane/
 [gemmini]: https://github.com/ucb-bar/gemmini
+[nvdla]: https://nvdla.org/
```
tymcauley, line 68: Beware that the `https` NVDLA site will give you a certificate warning. I've always had to use the `http` website unfortunately.

tymcauley:
> Wow. Good catch, I was starting to trace why that commit was problematic so you beat me to the punch!
> 
>     1. I just tried ResNet-50 with all the commits added back in and it works now.
> 
>     2. I'll go ahead and put and issue the `sw` repository since that seems like a major issue.
> 
> 
> Everything here should be up to date with the latest `dev`. I'll start the review process. As for the multi-clock, I think that should move to a different PR/issue since this PR currently works in a single clock system.

That's great news! And putting the multi-clock support in another PR sounds like a good idea. Thanks so much for filing that issue!

README.md
abejgonzalez: I just blindly used the `https`. Ill go ahead and update this.

generators/chipyard/src/main/scala/ConfigFragments.scala
abejgonzalez: Yes. Currently, nothing else in that file has a large header. So to keep things consistent, I deleted it.

sims/verilator/Makefile
```diff
@@ -84,13 +83,15 @@ ARIANE_VERILATOR_FLAGS = \
...
 CHIPYARD_VERILATOR_FLAGS = \
+	-Wno-fatal \
 	--assert
 
 VERILATOR_NONCC_OPTS = \
 	--top-module $(VLOG_MODEL) \
 	$(shell if ! grep -iq "module.*ariane" $(build_dir)/*.*v; then echo "$(CHIPYARD_VERILATOR_FLAGS)"; else echo "$(ARIANE_VERILATOR_FLAGS)"; fi) \
 	--output-split 10000 \
 	--output-split-cfuncs 100 \
+	--max-num-width 1048576 \
```
abejgonzalez, line 94: This was done to match RC: https://github.com/chipsalliance/rocket-chip/pull/2377

farzadfch:
@abejgonzalez sorry, it has been crazy these days. I'll get back to this soon (hopefully Tuesday).

abejgonzalez:
> @abejgonzalez sorry, it has been crazy these days. I'll get back to this soon (hopefully Tuesday).

No rush at all. Thanks!

tymcauley:
A few comments about some files in the `nvdla-wrapper` submodule:

* `src/main/resources/vsrc.mk`: Should we include some information on how these file lists were generated from the files that come out of the NVDLA tree build? It'd be great to have a relatively clear path for people who are interested in adding support for a new NVDLA configuration.
* `src/main/resources/build-hw-vmod.sh`: Probably related to the above question. I have changed this script locally so that instead of using `cp` to copy over the NVDLA RTL, it uses `rsync` with some `--exclude` and `--include` flags to gate which files we grab. Does that sound useful? If so, then we could also augment this script to auto-generate the contents of `src/main/resources/vsrc.mk`, if that seems practical. I'm thinking of something along these lines:

```bash
#!/bin/bash

set -ex

cd "$(dirname "$0")"

function build_nvdla_rtl() {
    local NVDLA_TYPE="$1"
    echo "Building NVDLA configuration: \"$NVDLA_TYPE\""
    cp tree.make hw
    sed -i -e "s/nv_small/$NVDLA_TYPE/" hw/tree.make
    cd hw
    ./tools/bin/tmake -build vmod
    cd ..
}

function install_nvdla_rtl() {
    local NVDLA_TYPE="$1"
    local SRC_DIR=hw/outdir/nv_$NVDLA_TYPE/vmod
    local DST_DIR=vsrc/$NVDLA_TYPE
    rm -rf "$DST_DIR/vmod"
    rsync -av \
        --exclude="*.vcp" \
        --exclude="*.swl" \
        --include="nv_assert_no_x.vlib" \
        --exclude="*.vlib" \
        "$SRC_DIR" "$DST_DIR"
}

function generate_vsrc_mk() {
    # TODO: generate `vsrc.mk` from the files in `vsrc/*`
    exit 1
}

NVDLA_TYPE=small
build_nvdla_rtl $NVDLA_TYPE
install_nvdla_rtl $NVDLA_TYPE

NVDLA_TYPE=large
build_nvdla_rtl $NVDLA_TYPE
install_nvdla_rtl $NVDLA_TYPE

generate_vsrc_mk
```

Then we could adjust the `install_nvdla_rtl` function to filter out files that we don't want to copy over, in case they would create issues for an auto-generated `vsrc.mk`.

abejgonzalez:
The `vsrc.mk` actually took a long time to get right since it is hard to know what the base set of files needed are. I just did a brute force approach of copying the initial `vsrc.mk` from Farzad and adding/deleting files that caused problems in VCS/Verilator/Synth. I'm not sure if there is a specific way to know what files are used. If there is a specific way to get the file list I think that would be good.

As mentioned earlier, after I used the `build-hw-vmod.sh` script to build the RTL I just brute-forced searching for the right files. I'm not opposed to changing it... but I'm not sure how many files are included versus excluded.

abejgonzalez:
@tymcauley @farzadfch I'm looking to push this through into `dev` right after the #544 PR goes through. Any final comments or questions?

@alonamid Any documentation / usability comments?

docs/Generators/NVDLA.rst
```diff
@@ -0,0 +1,16 @@
...
+The accelerator by itself exposes an AXI memory interface (or two if you use the "Large" configuration), a control interface, and an interrupt line.
+The main way to use the accelerator in Chipyard is to use the `NVDLA SW repository <https://github.com/ucb-bar/nvdla-sw>`_ that was ported to work on FireSim Linux.
+However, you can also use the accelerator in baremetal simulations (refer to ``tests/nvdla.c``).
+
+For more information on both the HW architecture and the SW, please visit their `website <http://nvdla.org/>`_.
+
+NVDLA Software with FireMarshal
+-------------------------------
+
+Located at ``software/nvdla-workload`` is a FireMarshal-based workload to boot Linux with the proper NVDLA drivers.
```
alonamid, line 15: How likely is this to get broken upon Linux kernel upgrades?

alonamid:
Why does this include a firesim bump now that the firesim configs are inside the nvdla wrapper?

alonamid:
LGTM overall, although other people tend to have stronger opinions than me about adding Verilator flags

abejgonzalez:
> Why does this include a firesim bump now that the firesim configs are inside the nvdla wrapper?

When I bump after the #544 PR, I will point to the correct FireSim repository.

docs/Generators/NVDLA.rst
abejgonzalez: Hopefully not by much... but frankly I wouldn't know until one happened.

abejgonzalez:
> LGTM overall, although other people tend to have stronger opinions than me about adding Verilator flags

The Verilator flag matches RC and fixes the same issue they were seeing so I don't see others having a problem with it.

docs/Generators/NVDLA.rst
alonamid: One is expected to happen this week? https://github.com/firesim/FireMarshal/pull/151

generators/firechip/src/main/scala/TargetConfigs.scala
```diff
@@ -63,23 +61,23 @@ class WithScalaTestFeatures extends Config((site, here, up) => {
...
 // Enables tracing on all cores
 class WithTraceIO extends Config((site, here, up) => {
   case BoomTilesKey => up(BoomTilesKey) map (tile => tile.copy(trace = true))
   case ArianeTilesKey => up(ArianeTilesKey) map (tile => tile.copy(trace = true))
   case TracePortKey => Some(TracePortParams())
 })
 
+// Adds a small/large NVDLA to the system
+class WithNVDLALarge extends nvidia.blocks.dla.WithNVDLA("large")
+class WithNVDLASmall extends nvidia.blocks.dla.WithNVDLA("small")
```
jerryz123, line 80: Why are these here?

abejgonzalez: IIRC to do something like this `WithNVDLASmall_DDR3FRFCFSLLC4MB_FireSimQuadRocketConfig` in the build recipe you need to have `WithNVDLA...` here. (so that everything is in the same scala package). I assume that this would be the easiest way to add an NVDLA for external users.

abejgonzalez: See https://github.com/ucb-bar/nvdla-wrapper/blob/master/firesim-collateral/nvdla_config_build_recipes.ini.

docs/Generators/NVDLA.rst
abejgonzalez: I updated the SW to match the bumped kernel (added extra instructions in the SW workload section to address this).

PR #509, opened 2020-04-11
Bump ESP Spike

alonamid:
**Related issue**: https://groups.google.com/forum/#!topic/chipyard/4pBzTO76gLw

**Type of change**: bug fix

**Impact**: software change

**Release Notes**
Bump esp-isa-sim 

This bump was likely missed during the Chipyard 1.2 release, which creates a mismatch between the Gemmini Spike model and the Gemmini RTL implementation


hngenc:
Lgtm

colinschmidt:
Should we have spike run some tests in CI to avoid this breakage in the future?

alonamid:
Worth discussing adding a Spike test to the esp-tools CI test

colinschmidt:
Spike tests run very fast so if we already build these tools (which I think we do) then there is pretty little downside IMO.

PR #512, opened 2020-04-15
Hammer sim integration

colinschmidt:

**Impact**: build system change

**Release Notes**
Chipyard can now run simulations with hammer, including gate-level and timing-annotated simulations.
It uses the `BINARY` flags, and the make targets `sim[-syn|-par][-timing][-debug]`


colinschmidt:
FYI: I need to write more text about this stuff

alonamid:
What is the need for the various `redo` make targets?

jwright6323:
> What is the need for the various `redo` make targets?

Iterating on a design without rerunning previous steps

sims/vcs/Makefile
```diff
@@ -25,10 +25,7 @@ sim_prefix = simv
 sim = $(sim_dir)/$(sim_prefix)-$(MODEL_PACKAGE)-$(CONFIG)
 sim_debug = $(sim_dir)/$(sim_prefix)-$(MODEL_PACKAGE)-$(CONFIG)-debug
 
-PERMISSIVE_ON=+permissive
-PERMISSIVE_OFF=+permissive-off
-
-WAVEFORM_FLAG=+vcdplusfile=$(sim_out_name).vpd
+include $(base_dir)/vcs.mk
```
abejgonzalez, line 28: This seems to be replacing the `VCS_*_OPTS` also. Can we delete those lines below and move it into this file?

vlsi/Makefile
```diff
@@ -87,6 +87,107 @@ $(SRAM_CONF): $(SRAM_GENERATOR_CONF)
...
+$(SIM_CONF): $(VLSI_RTL) $(HARNESS_FILE) $(HARNESS_SMEMS_FILE) $(sim_common_files)
+	mkdir -p $(dir $@)
+	echo "sim.inputs:" > $@
+	echo "  top_module: $(VLSI_TOP)" >> $@
+	echo "  input_files:" >> $@
+	for x in $(HARNESS_FILE) $(HARNESS_SMEMS_FILE); do \
+		echo '    - "'$$x'"' >> $@; \
+	done
+	echo "  input_files_meta: 'append'" >> $@
+	echo "  timescale: '1ns/10ps'" >> $@
```
abejgonzalez, line 108: While I understand why this is done in this way right now, I think it would be nice to directly generate this based off the `*_OPT`s in the `vcs.mk`. However, this can be future work.

vlsi/Makefile
```diff
@@ -124,4 +231,4 @@ $(OBJ_DIR)/hammer.d: $(HAMMER_D_DEPS)
 #########################################################################################
 .PHONY: clean
 clean:
-	rm -rf $(VLSI_OBJ_DIR) hammer-vlsi*.log __pycache__ output.json $(GENERATED_CONFS) $(gen_dir)
+	rm -rf $(VLSI_OBJ_DIR) hammer-vlsi*.log __pycache__ output.json $(GENERATED_CONFS) $(gen_dir) $(SIM_CONF) $(SIM_DEBUG_CONF)
```
abejgonzalez, line 234: Delete the `SIM_TIMING_CONF` also? What about the `<SYN/POWER>_CONF`s also?

vlsi/Makefile
```diff
@@ -87,6 +87,107 @@ $(SRAM_CONF): $(SRAM_GENERATOR_CONF)
...
+	echo "    'CLOCK_PERIOD=1.0'," >> $@
+	echo "    'PRINTF_COND=$(TB).printf_cond'," >> $@
+	echo "    'STOP_COND=!$(TB).reset'," >> $@
+	echo "    'RANDOMIZE_MEM_INIT'," >> $@
+	echo "    'RANDOMIZE_REG_INIT'," >> $@
+	echo "    'RANDOMIZE_GARBAGE_ASSIGN'," >> $@
+	echo "    'RANDOMIZE_INVALID_ASSIGN']" >> $@
+	echo "  defines_meta: 'append'" >> $@
+	echo "  compiler_opts: [" >> $@
+	echo "    '-I$(RISCV)/include'," >> $@
```
abejgonzalez, line 133: Is the `-CC "-I$(VCS_HOME)/include"` done by default?

vlsi/Makefile
colinschmidt: Yeah that was actually what I planned to do. I forgot that I hadn't fixed this yet. thanks for the reminder.

vlsi/Makefile
colinschmidt: Good catch.

vlsi/Makefile
colinschmidt: will double check.

sims/vcs/Makefile
colinschmidt: Yes that is the intent.

vlsi/Makefile
colinschmidt: Fixed

vlsi/Makefile
colinschmidt: Fixed

sims/vcs/Makefile
colinschmidt: Fixed.

vlsi/Makefile
```diff
@@ -87,6 +87,89 @@ $(SRAM_CONF): $(SRAM_GENERATOR_CONF)
...
+	echo "  defines_meta: 'append'" >> $@
+	echo "  compiler_opts:" >> $@
+	for x in $(filter-out -CC,$(VCS_CC_OPTS)); do \
+		echo '    - "'$$x'"' >> $@; \
+	done
+	echo "  compiler_opts_meta: 'append'" >> $@
+	echo "  execution_flags_prepend: ['$(PERMISSIVE_ON)']" >> $@
+	echo "  execution_flags_append: ['$(PERMISSIVE_OFF)']" >> $@
+	echo "  execution_flags:" >> $@
+	echo "    - '+max-cycles=$(timeout_cycles)'" >> $@
```
abejgonzalez, line 127: This is in the `SIM_FLAGS` already.

vlsi/Makefile
colinschmidt: Fixed. SYN_CONF is included in generated confs. I think that could use a clean up later. There is a makefile change in hammer that will probably change how we want to handle that.

vlsi/Makefile
colinschmidt: Will fix this after the dev merge

vcs.mk
```diff
@@ -0,0 +1,45 @@
+PERMISSIVE_ON=+permissive
```
abejgonzalez, line 1: We can now delete this from here since your other PR moves this into `variables.mk`.

vlsi/Makefile
abejgonzalez: Bump.

vlsi/Makefile
```diff
@@ -96,12 +179,19 @@ ifeq ($(CUSTOM_VLOG), )
 	GENERATED_CONFS += $(if $(filter $(tech_name), asap7), , $(SRAM_CONF))
 endif
 
+#TODO: remove filter-out once PR#572 is merged
```
abejgonzalez, line 182: This should now be addressed

vcs.mk
colinschmidt: Fixed.

vlsi/Makefile
colinschmidt: Fixed.

vlsi/Makefile
colinschmidt: Fixed.

PR #513, opened 2020-04-15
Fix bug in verilator Makefile for 'debug' rule

tymcauley:
The 'debug' rule is currently cleaning out the non-debug-model directory rather than the debug-model directory. This means a debug-mode build would cause the non-debug-mode build to get cleared out.

This PR fixes that, and changes both the debug and non-debug rules to use the variables defined for referring to these two model directories.

**Related issue**: 

**Impact**: other

**Release Notes**

### DIFF
--- a/sims/verilator/Makefile
+++ b/sims/verilator/Makefile
@@ -116,14 +116,14 @@ model_mk_debug = $(model_dir_debug)/V$(VLOG_MODEL).mk
 # build makefile fragment that builds the verilator sim rules
 #########################################################################################
 $(model_mk): $(sim_vsrcs) $(sim_common_files) $(EXTRA_SIM_REQS)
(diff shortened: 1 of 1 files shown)


abejgonzalez:
This LGTM. Could you rebase on most recent Chipyard. I would like to see if CI runs on this PR.

tymcauley:
> This LGTM. Could you rebase on most recent Chipyard. I would like to see if CI runs on this PR.

Just rebased onto the most recent `dev`.

abejgonzalez:
I've went ahead and started the CI.

tymcauley:
Looks like the CI had some problems?

abejgonzalez:
Gah. This seems to be an issue with how CircleCI deals with secrets. Im pretty sure this is fine.

Thanks for the PR.

Issue #514, opened 2020-04-15
Including Prerequisites in "Quick Start" or Script


Matthias-Raudonis:
**Impact**: software

**Description**

Following the Quick Start guide from a clean Ubuntu 18 LTS install, one encounters repeated compile errors because some packages are missing. 

In "1.4. Initial Repository Setup" I found some prerequisites are mentioned. 

I am suggesting a reference from the Quick Start Section to the above section. Why not make a script and include it into the Quick Start Process?

**What is a motivating example for changing the behavior?**
It takes a much shorter time to compile if there are no repeated errors.

Matthias-Raudonis:
Note: This was the cause of issues #455 and #436

alonamid:
Fixed by #518

Issue #455, opened 2020-02-28
recipe for target 'firrtl_temp' failed


umair-riaz:
Hi,
I have gone through the setup and installation of Chipyard and after installation, it passed all make run-asm-tests and make run-bmark-tests but got an error when running the following command.

Command:
>> make CONFIG=SmallBoomAndRocketConfig

Error:
/chipyard/common.mk:68: recipe for target 'firrtl_temp' failed

If someone has gone through it, kindly reply.

abejgonzalez:
Can you please post a build log starting from the command run to the error, please? Normally, when this breaks, the FIRRTL compile didn't pass. Did you make any changes to the code? Version of Chipyard?

umair-riaz:
I didn't make any changes but got error. I have attached the build log starting from the command run.


[build_log.txt](https://github.com/ucb-bar/chipyard/files/4292114/build_log.txt)

KlepD-SAL:
Hi,

any updates on the topic?
I also have the same problem when running the default example with `make`.

Following error occurs:
```
[...]

Issue #436, opened 2020-02-17
RISCV Toolchain Build Error


rayjyh:
Hi, I got the following error when building the riscv toolchain by running ./build-toolchains.sh. 
[log.txt](https://github.com/ucb-bar/chipyard/files/4214349/log.txt)

jingyuhe@jingyuhe-VirtualBox:~/Desktop/chipyard/scripts$ ./build-toolchains.sh > log.txt
+ set -e
+ set -o pipefail
++ pwd
+ RDIR=/home/jingyuhe/Desktop/chipyard/scripts
++ git rev-parse --show-toplevel
+ CHIPYARD_DIR=/home/jingyuhe/Desktop/chipyard
+ TOOLCHAIN=riscv-tools
+ EC2FASTINSTALL=false
+ RISCV=
+ '[' '' '!=' '' ']'
+ '[' -z '' ']'
+ INSTALL_DIR=riscv-tools-install
++ pwd
+ RISCV=/home/jingyuhe/Desktop/chipyard/scripts/riscv-tools-install
+ echo 'Installing toolchain to /home/jingyuhe/Desktop/chipyard/scripts/riscv-tools-install'
+ export RISCV=/home/jingyuhe/Desktop/chipyard/scripts/riscv-tools-install
+ RISCV=/home/jingyuhe/Desktop/chipyard/scripts/riscv-tools-install
+ cd /home/jingyuhe/Desktop/chipyard
++ pwd
+ SRCDIR=/home/jingyuhe/Desktop/chipyard/toolchains/riscv-tools
+ '[' -d /home/jingyuhe/Desktop/chipyard/toolchains/riscv-tools ']'
+ . ./scripts/build-util.sh
++ '[' -n /home/jingyuhe/Desktop/chipyard/toolchains/riscv-tools ']'
+++ getconf _NPROCESSORS_ONLN
++ ncpu=1
++ case ${ncpu} in
++ export 'MAKEFLAGS=-j 1 -j8'
++ MAKEFLAGS='-j 1 -j8'
+++ command -v make
++ MAKE=/usr/local/bin/make
++ readonly MAKE
+ '[' false = true ']'
+ module_prepare riscv-gnu-toolchain qemu
+ set -e
+ name=riscv-gnu-toolchain
+ shift
+ dir=/home/jingyuhe/Desktop/chipyard/toolchains/riscv-tools/riscv-gnu-toolchain
+ echo '=>  Starting riscv-gnu-toolchain build'
[...]

PR #518, opened 2020-04-18
Include scripts for dependencies in documentation

Matthias-Raudonis:
**Related issue**: 
Related to #455, #436 and #514: Users forget to install dependencies because there is no reference in the quick start documentation

**Impact**: documentation

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -37,6 +37,20 @@ Start by fetching Chipyard's sources. Run:
 
 This will initialize and checkout all of the necessary git submodules.
 
+Installing Dependencies
+-------------------------------------------
+Installing the recommended dependencies on Ubuntu/Debian-based platforms:
+
+.. code-block:: shell
+
+    ./scripts/ubuntu-req.sh
+    
+Installing the recommended dependencies on CentOS-based platforms:
+
+.. code-block:: shell
+
+   ./scripts/centos-req.sh
+
 Installing the RISC-V Tools
 -------------------------------------------
 


alonamid:
These scripts are currently included in the "Initial Repository Setup" of the docs (https://chipyard.readthedocs.io/en/latest/Chipyard-Basics/Initial-Repo-Setup.html)

Our goal with "Quick Start" was for it to be a fairly short segment (otherwise the docs home page become very long). Maybe in order to prevent duplicaiton, a better approach would be to include an additional link in the "Quick Start" segment to the more detailed repo setup docs page for more detailed system requirements?

jerryz123:
I agree with Matthias here, if this section can't actually get users started, then it's not a quick start. This change seems good to me.

alonamid:
And alternative could be to put links to the requirement scripts in the "Quick Start" section, rather than embedding the scripts?

Matthias-Raudonis:
I am aware of the "Initial Repository Setup". Many users are not and spend a long time compiling until errors show. The rest of the "Initial Repository Setup" is contained in the quick start.

Therefore, I believe the requirement scripts must be clearly stated in the quick start. I only included the name of the scripts to call, so it really does not take a lot of space.  

Alternatively, I could shorten it in combining the Ubuntu CentOS texts.

alonamid:
Oh, I misread the PR. Sorry about that. I thought the scripts code was included rather than just referenced. In that case I think this is fine. 
I would just recommend moving it to the "System Requirements" section within "Quick Start"? (I view dependencies as part of the System Requirements, and since we do not provide any mac/windows/BSD dependencies, the scripts probably fit there best)

Matthias-Raudonis:
I would agree on that, but the quick start guide is written step by step:
Its more convenient to call the script after cloning the repo with git.

But maybe the "System Requirements" section would be a good place for a reference to the  "Initial Repository Setup"?

PR #525, opened 2020-04-23
Get icenet and testchipip unit tests working

zhemao:
IceNet and TestChipIP previously had their own generators and testharnesses in order to run unittests. But now that generators can use configurations in other packages, that isn't necessary any more. Ideally, we'd be able to use the freechips.rocketchip.unittest TestHarness directly, but unfortunately it doesn't output the *.plusArgs file, which is necessary for verilator. 

**Related issue**: 

https://github.com/firesim/icenet/pull/24

https://github.com/ucb-bar/testchipip/pull/78

**Impact**: other

**Release Notes**

Support for running IceNet and TestChipIP unit tests

### DIFF
--- a/generators/chipyard/src/main/scala/unittest/TestHarness.scala
+++ b/generators/chipyard/src/main/scala/unittest/TestHarness.scala
@@ -1,0 +1,9 @@
+package chipyard.unittest
+
+import chisel3._
+import freechips.rocketchip.config.Parameters
+
+class TestHarness(implicit val p: Parameters) extends Module {
+  val io = IO(new Bundle { val success = Output(Bool()) })
+  io.success := Module(new UnitTestSuite).io.finished
+}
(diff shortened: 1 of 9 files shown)


davidbiancolin:
I'd consider adding [a subset] of these unit tests to the CI.

generators/chipyard/src/main/scala/unittest/TestHarness.scala
```diff
@@ -0,0 +1,9 @@
+package chipyard.unittest
+
+import chisel3._
+import freechips.rocketchip.config.Parameters
+
+class TestHarness(implicit val p: Parameters) extends Module {
+  val io = IO(new Bundle { val success = Output(Bool()) })
+  io.success := Module(new UnitTestSuite).io.finished
```
jerryz123, line 8: Can't this be done by setting `BuildSystem` to `Module(new UnitTestSuite)`, and using a `IOBinder` to connect `io.success` to `dut.io.finished`?

zhemao: But then I'd need to change the configuration inside testchipip and have it depend on chipyard or create one inside chipyard.

sims/vcs/Makefile
```diff
@@ -56,7 +56,7 @@ VCS_CC_OPTS = \
 VCS_NONCC_OPTS = \
 	+lint=all,noVCDE,noONGS,noUI \
 	-error=PCWM-L \
-	-timescale=1ns/10ps \
+	-timescale=1ns/1ps \
```
abejgonzalez, line 59: Why is this change needed?

zhemao: It's for the ClockUtilTests. They need 1ps precision timing for the clock generators.

PR #533, opened 2020-04-27
[docs] Make the link to initial setup stand out.

colinschmidt:
As discussed no one reads anything so we make this very important notice look very important.


docs/index.rst
```diff
@@ -10,10 +10,8 @@ Welcome to Chipyard's documentation!
 
 Chipyard is a framework for designing and evaluating full-system hardware using agile teams.
 It is composed of a collection of tools and libraries designed to provide an integration between open-source and commercial tools for the development of systems-on-chip.
-New to Chipyard? Jump to the :ref:`Chipyard Basics` page for more info.
 
-Quick Start
-===============================
+.. IMPORTANT:: **New to Chipyard?** Jump to the :ref:`Chipyard Basics` page for more info.
 
 System Requirements
```
sagark, line 16: Don't we want to remove this stuff too?

sagark:
I thought the agreement was to get rid of the duplicated installation instructions on the index page?

docs/index.rst
colinschmidt: Yes we do but I wanted someone who knew if anything needed to be kept to make the deletion.
I'm not sure they are 100% redundant.

sagark:
Also, depending on when we're planning to ship a release, can we just direct these changes at master and the pull back into dev?

docs/index.rst
alonamid: They aren't completely 100% redundant right now. We should probably make the toolchains installation in initial-setup more similar to the current quick-start form (so peolpe who are just looking for codeline to copy will copy the correct one).
We should also copy the "what's next" part.

docs/index.rst
```diff
@@ -10,10 +10,8 @@ Welcome to Chipyard's documentation!
 
 Chipyard is a framework for designing and evaluating full-system hardware using agile teams.
 It is composed of a collection of tools and libraries designed to provide an integration between open-source and commercial tools for the development of systems-on-chip.
-New to Chipyard? Jump to the :ref:`Chipyard Basics` page for more info.
 
-Quick Start
-===============================
+.. IMPORTANT:: **New to Chipyard?** Jump to the :ref:`Chipyard Basics` page for more info.
```
alonamid, line 14: Do we want them to just to "Chipyard Basics", or to "Initial Repository Setup"?

colinschmidt:
Ok @sagark @alonamid I've merged the two quick starts so I think we are ready to go.

docs/index.rst
```diff
@@ -10,89 +10,9 @@ Welcome to Chipyard's documentation!
 
 Chipyard is a framework for designing and evaluating full-system hardware using agile teams.
 It is composed of a collection of tools and libraries designed to provide an integration between open-source and commercial tools for the development of systems-on-chip.
-New to Chipyard? Jump to the :ref:`Chipyard Basics` page for more info.
 
-Quick Start
-===============================
+.. IMPORTANT:: **New to Chipyard?** Jump to the :ref:`Initial Repository Setup` page for more info.
```
alonamid, line 14:
```suggestion
.. IMPORTANT:: **New to Chipyard?** Jump to the :ref:`Initial Repository Setup` page for setup instructions.
```

alonamid:
LGTM other than that single suggestions

Issue #535, opened 2020-04-28
Heterogeneous-SoCs.rst should use List() or ::, not ++


cousteaulecommandant:
`docs/Customization/Heterogeneous-SoCs.rst` states this way of creating multiple dissimilar Rockets:
```scala
val rocketTile0 = RocketTileParams(...) // params for rocket core 0
val rocketTile1 = RocketTileParams(...) // params for rocket core 1
rocketTile0 ++ rocketTile1
```
However, you can't use the `++` operator (list concatenation) on RocketTileParams, since they are not lists nor have a `++` method defined (as of v1.2.0); trying to do so results in an error.
What is needed is to _create_ a list:
```scala
List(rocketTile0, rocketTile1) // more readable option
```
or
```scala
rocketTile0 :: rocketTile1 :: Nil // "Lisp-style" list creation
```

abejgonzalez:
Thanks for the catch. You are correct. `BoomTilesKey` and `RocketTilesKey` both take a `Seq` of `<Boom/Rocket>TileParams` so you can use `Seq`, `List`, etc. I went ahead and PR'ed a fix.

PR #537, opened 2020-04-29
Fix "WithHeterCoresSetup" in hetero. docs

abejgonzalez:
**Related issue**:#535

**Type of change**: bug fix

**Impact**:other

**Release Notes**
This fixes the `WithHeterCoresSetup` config. fragment to correctly use a list to combine all Tile parameters instead of `++`.


colinschmidt:
Would this have been caught if it were actual code that was included into the docs?

abejgonzalez:
> Would this have been caught if it were actual code that was included into the docs?

Yes... if that code was under CI (and if an example was made for it).

Issue #422, opened 2020-02-03
add a spi flash config


siyy123:
**Impact**: rtl

**Description**

Can we add a spi flash example? since in real tapeout the chip may need a external storage to store firmware, kernel, os and software.

jwright6323:
This would be a good thing to add, but we don't currently have access to any open-source verification IP for SPI flash.

colinschmidt:
I think we could add a config fragment that doesn't have any test harness support.
@siyy123 Would that be helpful?

jwright6323:
I haven't really looked at this much, but there appears to be a freely available QSPI model here: http://ww1.microchip.com/downloads/en/DeviceDoc/sst26wf016b.v

jwright6323:
Also: https://www.micron.com/products/nor-flash/serial-nor-flash/part-catalog/n25q00aa13g1240e

We can look into licenses to see if either are compatible with ours.

jwright6323:
I looked into these two things a bit, and unfortunately they won't work with Verilator because they have # delays. We'll probably have to build our own model.

a0u:
There are Verilog models from ISSI: http://www.issi.com/US/product-flash.shtml
The HiFive-1 board uses the IS25LP128-JBLE part.

N25Q from Micron is being replaced with the MT25Q family, but on a cursory glance, the newer models don't appear to support zero-delay simulation either.

jwright6323:
This now added into dev with #546 . It will be in the Chipyard 1.3 release.

PR #544, opened 2020-05-05
Rocket Chip Bump with Chisel 3.3 and FIRRTL 1.3

alonamid:
**Related issue**: 

**Type of change**: new feature

**Impact**: rtl change

**Release Notes**

**Todo**

- [x] Bump sifive-blocks to get https://github.com/sifive/sifive-blocks/pull/123/files
- [x] Point sifive-cache back to their repo after https://github.com/sifive/block-inclusivecache-sifive/pull/10 is merged

PR #546, opened 2020-05-06
Add SPI flash support

jwright6323:
**Related issue**: #422 

Related PR: ucb-bar/testchipip#84

**Impact**: rtl change

**Release Notes**
* Added support for a SPI flash memory model with binary memory images


jerryz123:
Note that #544 removes the need for the WithNoSpi fragments.

docs/Generators/TestChipIP.rst
```diff
@@ -84,3 +84,11 @@ output a UART log to a particular file using ``+uartlog=<NAME_OF_FILE>`` during
 
 By default, this UART Adapter is added to all systems within Chipyard by adding the
 ``WithUART`` and ``WithUARTAdapter`` configs.
+
+SPI Flash Model
+---------------
+
+The SPI flash model is a devices that models a simple SPI flash device. It currently
```
chick, line 91: NIT
devices => device
```suggestion
The SPI flash model is a device that models a simple SPI flash device. It currently
```

generators/chipyard/src/main/scala/ConfigFragments.scala
```diff
@@ -60,6 +67,10 @@ class WithNoUART extends Config((site, here, up) => {
   case PeripheryUARTKey => Nil
 })
 
+class WithNoSPIFlash extends Config((site, here, up) => {
```
abejgonzalez, line 70: Is this going to be rebased on #544 so that you can remove this? Or is the plan to merge this in before?

jwright6323: Depends on which goes in first. If #544 goes in first, yea, I'll remove this. If this goes in first, I'll make the change and commit it to the #544 branch. I imagine the former will happen.

jwright6323:
When this CI passes, let's squash this.

PR #547, opened 2020-05-06
Bump verilator to v4.034

albert-magyar:
**Related issue**:  verilator/verilator#2300
**Type of change**: bug fix
**Impact**: tool change
**Release Notes**
The default version of Verilator has changed to v4.034. Since this release adds enhanced support for Verilog timescales, a simpler behavior closer to that seen with previous versions is emulated by default with a flat `--timescale` argument.


sims/verilator/Makefile
```diff
@@ -86,7 +86,9 @@ ARIANE_VERILATOR_FLAGS = \
 CHIPYARD_VERILATOR_FLAGS = \
 	--assert
 
+# Use --timescale to approximate timescale behavior of pre-4.034
 VERILATOR_NONCC_OPTS = \
+	--timescale 1ns/1ps \
```
colinschmidt, line 91: We should consider having the argument part of this pulled out since we give the same value to both verilator and vcs.

albert-magyar: I could just put it in `SIM_FLAGS`?

colinschmidt: Unfortunately one of them wants 2 dashes and the other wants 1 so I was only suggestions lifting the standard verilog `1ns/1ps`.
Its too much work for now but it would be nice to keep things DRY.

albert-magyar: `$(SIM_FLAG_DASHES)timescale 1ns/1ps` :)

colinschmidt: wonderful :)

davidbiancolin: Where's that chef-albert garnishing gif when you need it.

albert-magyar:
Per reviewer feedback, I made the addition of the flag conditional on Verilator being new enough to support it. This will stop the new Makefile changes from breaking if people don't update their Verilator installs.

PR #551, opened 2020-05-14
Assemble the firrtl-test.jar and put it in its own directory

davidbiancolin:
1.3 has landed -- yay! Time to use the new features.

FIRRTL now provides an assembly task for `test`, which packages up a bunch of test classes that are useful for testing firrtl passes _outside_ of firrtl. This PR calls out to `test:assembly` and drops teh resulting jar in `test_lib`. Subprojects that want to use these classes (under the `firrtl.testutils` package) may either use this new directory as their unmanagedBase, or add the resulting jar to their classpath.

I picked `test_lib` to mirror what was done in RC.

 
**Impact**:  other

**Release Notes**
Assemble the firrtl-test.jar; put in separate unmanaged directory "test_lib"

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ import Tests._
 
 // This gives us a nicer handle  to the root project instead of using the
 // implicit one
-lazy val chipyardRoot = RootProject(file("."))
+lazy val chipyardRoot = Project("chipyardRoot", file("."))
 
 lazy val commonSettings = Seq(
   organization := "edu.berkeley.cs",
--- a/common.mk
+++ b/common.mk
@@ -36,11 +36,18 @@ TESTCHIPIP_CLASSES ?= "$(TESTCHIP_DIR)/target/scala-$(SCALA_VERSION_MAJOR)/class
 # jar creation variables and rules
 #########################################################################################
 FIRRTL_JAR := $(base_dir)/lib/firrtl.jar
+FIRRTL_TEST_JAR := $(base_dir)/test_lib/firrtl-test.jar
 
 $(FIRRTL_JAR): $(call lookup_srcs,$(CHIPYARD_FIRRTL_DIR),scala)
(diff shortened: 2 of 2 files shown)


build.sbt
```diff
@@ -2,7 +2,7 @@ import Tests._
 
 // This gives us a nicer handle  to the root project instead of using the
 // implicit one
-lazy val chipyardRoot = RootProject(file("."))
+lazy val chipyardRoot = Project("chipyardRoot", file("."))
```
colinschmidt, line 5: Is this just cosmetic or does it change functionality?

common.mk
```diff
@@ -36,13 +36,20 @@ TESTCHIPIP_CLASSES ?= "$(TESTCHIP_DIR)/target/scala-$(SCALA_VERSION_MAJOR)/class
...
+FIRRTL_TEST_JAR := $(base_dir)/test_lib/firrtl-test.jar
 
 $(FIRRTL_JAR): $(call lookup_srcs,$(CHIPYARD_FIRRTL_DIR),scala)
 	$(MAKE) -C $(CHIPYARD_FIRRTL_DIR) SBT="$(SBT)" root_dir=$(CHIPYARD_FIRRTL_DIR) build-scala
 	mkdir -p $(@D)
 	cp -p $(CHIPYARD_FIRRTL_DIR)/utils/bin/firrtl.jar $@
 	touch $@
 
+$(FIRRTL_TEST_JAR): $(call lookup_srcs,$(CHIPYARD_FIRRTL_DIR),scala)
+	mkdir -p $(@D)
```
colinschmidt, line 48: I think we should match the other jar creation order of commands.
I don't really care whether its one way or the other. It seems like in the order you've written sometimes test_lib is created and empty (not really a big deal).

build.sbt
davidbiancolin: It makes it possible to reference the root project using a `ProjectRef` in FireSim.

common.mk
davidbiancolin: Will do.

PR #552, opened 2020-05-15
Don't try to init nonexistent midas submodule

albert-magyar:
**Type of change**: bug fix
**Impact**: other
**Release Notes**
The submodule initialization script has been updated to no longer check out the now-inlined MIDAS project as a submodule of FireSim (see firesim/firesim#400).


albert-magyar:
Plot twist: @zhemao brought up the good point that there are still submodules inside `midas`. I changed the script to just do `dramsim2`, but the version in dev is actually already fine.

zhemao:
The dramsim2 submodule in firesim is going away, and chipyard already doesn't depend on it, as it uses its own dramsim2 submodule under tools/. IMO, you should just delete the line.

abejgonzalez:
This should be rebased with `dev` again to get the toolchain fix.

PR #554, opened 2020-05-17
Have FireChip use the standard Chipyard generator

davidbiancolin:
This further streamlines how target's can be used across Chipyard and FireSim by having FireSim reuse the standard CY generator.

Also: 
- I fixed the obviously broken initial port i wrote for "_" config concatenation. 
- I worked around an issue where targets whose Params did not define XLEN could not be generated.  

I've disabled the device-oriented target tests for now because FireSim was generating a custom makefrag snipped for those. I think the right solution there is to figure out we can have the elaborated target add additional makefrag snippets -- then we could emit the annotation from the block device itself or using an IO binder.

**Related issue**: firesim/firesim#557

**Impact**: other

**Release Notes**
- Have FireChip use the common Chipyard generator


generators/chipyard/src/main/scala/stage/phases/AddDefaultTests.scala
```diff
@@ -30,9 +31,13 @@ class AddDefaultTests extends Phase with PreservesAll[Phase] with HasRocketChipS
   private def addTestSuiteAnnotations(implicit p: Parameters): Seq[Annotation] = {
     val annotations = mutable.ArrayBuffer[Annotation]()
     val suiteHelper = new TestSuiteHelper
-    suiteHelper.addRocketTestSuites
-    suiteHelper.addBoomTestSuites
-    suiteHelper.addArianeTestSuites
+    // Use Xlen as a proxy for detecting if we are a processor-like target
+    // The underlying test suites expect this field to be defined
+    if (p.lift(XLen).nonEmpty) {
```
jerryz123, line 36: Nice

generators/firechip/src/test/scala/ScalaTestSuite.scala
```diff
@@ -127,20 +116,16 @@ abstract class FireSimTestSuite(
...
   clean
-  mkdirs
-  elaborate
-  generateTestSuiteMakefrags
-  generateArtefacts
   runTest("verilator", "rv64ui-p-simple", false, Seq(s"""EXTRA_SIM_ARGS=+trace-humanreadable0"""))
   //diffTracelog("rv64ui-p-simple.out")
   runSuite("verilator")(benchmarks)
-  runSuite("verilator")(FastBlockdevTests)
+  //runSuite("verilator")(FastBlockdevTests)
```
colinschmidt, line 122: I'm generally opposed to committing commented out code, since git preserves history it won't be hard to find this later.
I'd also suggest filing an issue to track re-enabling these tests.

davidbiancolin: Yeah, i totally agree.

davidbiancolin:
CI errors are transient failures; merging before have to yet another dev -> feature merge.

PR #560, opened 2020-05-20
Remove Dromajo build from build-toolchains.sh

abejgonzalez:
**Related issue**: 

**Type of change**: bug fix

**Impact**: other

**Release Notes**
Fixes the Dromajo build in the toolchain script and adds small comments to the generated files made by the `init-submodules` and `build-toolchains` script.


jerryz123:
How did this work before....

abejgonzalez:
The toolchain wasn't getting rebuilt because it was cached. So when I got rid of the cache it exposed this error.

PR #565, opened 2020-05-24
Upgrade verilator to support permissive args in the same way as vcs

### LINKED ISSUE #563: Fix broken verilator permissive runtime arguments
So we do need permissive and its even slightly different for verilator so its good that its overrideable:
```
simulator-example-RocketConfig +max-cycles=5000 +verbose -- +permissive +verilator+V +permissive-off output/example.TestHarness.RocketConfig/rv64ui-p-addi
using random seed 1590055841
  Version: Verilator 4.016 2019-06-16
%Error: COMMAND_LINE:0: Exiting due to command line argument (not an error)
Aborting...
Aborted (core dumped)
```
versus
```
simulator-example-RocketConfig +max-cycles=5000 +verbose +verilator+V output/example.TestHarness.RocketConfig/rv64ui-p-addi
simulator-example-RocketConfig: invalid plus-arg (Verilog or HTIF) "+verilator+V"
```

_Originally posted by @colinschmidt in https://github.com/ucb-bar/chipyard/pull/562/review_comment/create_

colinschmidt:
Fixes #563 

**Release Notes**
Verilator now supports `+permissive` and now can be passed its runtime arguments such as `+verilator+rand+reset+0`

### DIFF
--- a/generators/utilities/src/main/resources/csrc/emulator.cc
+++ b/generators/utilities/src/main/resources/csrc/emulator.cc
@@ -126,27 +126,30 @@ int main(int argc, char** argv)
   int verilog_plusargs_legal = 1;
 
   dramsim = 0;
+  opterr = 1;
 
   while (1) {
     static struct option long_options[] = {
-      {"cycle-count", no_argument,       0, 'c' },
-      {"help",        no_argument,       0, 'h' },
-      {"max-cycles",  required_argument, 0, 'm' },
-      {"seed",        required_argument, 0, 's' },
-      {"rbb-port",    required_argument, 0, 'r' },
-      {"verbose",     no_argument,       0, 'V' },
-      {"dramsim",     no_argument,       0, 'D' },
+      {"cycle-count",     no_argument,       0, 'c' },
+      {"help",            no_argument,       0, 'h' },
+      {"max-cycles",      required_argument, 0, 'm' },
+      {"seed",            required_argument, 0, 's' },
+      {"rbb-port",        required_argument, 0, 'r' },
+      {"verbose",         no_argument,       0, 'V' },
+      {"dramsim",         no_argument,       0, 'D' },
+      {"permissive",      no_argument,       0, 'p' },
+      {"permissive-off",  no_argument,       0, 'o' },
 #if VM_TRACE
-      {"vcd",         required_argument, 0, 'v' },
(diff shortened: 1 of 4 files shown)


colinschmidt:
You can now add verilator runtime arguments to `EXTRA_SIM_FLAGS` and it works as expected.

jerryz123:
If PERMISSIVE will be required for all simulation targets (VCS, Verilator), does it makes sense to remove the PERMISSIVE and PERMISSIVE_OFF variables defined in vcs/Makefile and verilator/Makefile?

colinschmidt:
Yeah I think we can hoist them out now.

generators/utilities/src/main/resources/csrc/emulator.cc
```diff
@@ -160,6 +163,8 @@ int main(int argc, char** argv)
       case 'r': rbb_port = atoi(optarg);    break;
       case 'V': verbose = true;             break;
       case 'D': dramsim = 1;                break;
+      case 'p': opterr = 0;                 break;
```
jwright6323, line 166: why not just call this variable `permissive`? This isn't telling you that there was an options error right, just masking them?

colinschmidt: I mirrored the variable name in htif/fesvr but I can name it whatever.

colinschmidt: I'll change this after the current CI run finishes in case I need to fix anything else.

jwright6323: Oh, I see. I mean that's a fine reason to leave it as-is then; do whichever you like better

variables.mk
```diff
@@ -136,6 +136,8 @@ output_dir=$(sim_dir)/output/$(long_name)
 #########################################################################################
 # helper variables to run binaries
 #########################################################################################
+PERMISSIVE_ON=+permissive
```
abejgonzalez, line 139: Why not just inline this instead of keeping around the two variables?

colinschmidt: This is slightly more DRY in my opinion. It should always be the same in both places and this will ensure that. If we inline it someone might change one or the other without realizing they need to change both.
That my opinion and if others think we should inline I'm ok with that.

abejgonzalez: Sounds good.

Issue #567, opened 2020-05-25
[docs] ScratchpadRocketConfig description in Memory-Hierarchy.rst doesn't match its behavior


cousteaulecommandant:
`docs/Customization/Memory-Hierarchy.rst` section "The L1 Caches" shows an example code of how to use a scratchpad (on-chip memory).  However, the code doesn't reflect what the description says.

What the description suggests:
* The scratchpad replaces the L1 data cache (this is what `With1TinyCore` does by setting `RocketTilesKey` -> `dcache` -> `scratch`).
* No L2 by setting number of banks to 0 (which can be achieved with `WithNBanks(0)`).
* No memory bus by setting the number of channels to 0 (`WithNMemoryChannels(0)`).
* Only a single core allowed and no external memory allowed.
* Core -> L1D Scratchpad

What the code does:
* The scratchpad hangs from the `mbus` (uses `WithBackingScratchpad`); it does NOT use L1 scratchpad.
* L2 present (`WithInclusiveCache` in use; diplomacy graph shows one L2 node).
* No memory port (`WithNoMemPort`), BUT memory bus present.
* IIUC there's no reason to limit this system to one core and remove external memory, as long as there's no address overlap.  (I haven't tried and I may be wrong.)
* Core -> L1I/D cache -> TileBus -> SystemBus -> L2 cache -> MemoryBus -> BackingScratchpad

I think both cases are useful.  The latter simply replaces the external memory with one generated inside the Top and connected to the mbus directly (which is something I wanted to do because it seemed like the logical way to provide on-chip memory, and the example has really helped me once I understood what it did), whereas the former provides a "tightly coupled memory" that is close to the core but cannot be shared among cores.  Therefore, it might be a good idea to document both cases (add an example for the former and a description for the latter).
Nevertheless, the `WithBackingScratchpad` example and documentation should be moved outside of section 6.9.1 ("The L1 Caches"), since it has nothing to do with L1, to section 6.9.5 ("The Outer Memory System") or to a new section.

abejgonzalez:
Based on the `dev` branch, this was changed accidentally from the description to a different configuration (which was similarly named but not used for the same purpose).

Faulty commit:
https://github.com/ucb-bar/chipyard/commit/708a5fb9a610c6f003d3d78f3657cd8f125f879f

Original code that matched the description:
```
    class ScratchpadSmallRocketConfig extends Config(
         new chipyard.iobinders.WithUARTAdapter ++
[...]

PR #569, opened 2020-05-25
Scratchpad Documentation Fix

abejgonzalez:
**Related issue**: #567 

**Type of change**: bug fix

**Impact**: other

**Release Notes**
Fixes the docs.

### DIFF
--- a/docs/Customization/Memory-Hierarchy.rst
+++ b/docs/Customization/Memory-Hierarchy.rst
@@ -31,8 +31,8 @@ Note that these configurations fully remove the L2 cache and mbus.
 
 .. literalinclude:: ../../generators/chipyard/src/main/scala/config/RocketConfigs.scala
     :language: scala
-    :start-after: DOC include start: scratchpadrocket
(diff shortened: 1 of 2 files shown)


jerryz123:
I'm actually in favor of not documenting or including the DTIM scratchpad option. It doesn't seem to be something we would ever use.

alonamid:
Is there a reason not to document it?

jerryz123:
It's not our job to document all of rocketchip, especially the features which we don't regularly use.

abejgonzalez:
It sounds like external users like both options described (2 external people looked at the documentation and asked questions / gave feedback on it). So I think its beneficial to keep it.

alonamid:
To the actual content of the PR - there seems to be a mismatch between the text and the code example. You remove the memory port in the L1 code example, but in the text you say that you should do that for the mbus scratchpad case.

abejgonzalez:
No I do that for both cases?

alonamid:
you're right, github diff got me again...

PR #572, opened 2020-05-26
Move all non-synthesizable constructs into the test harness

colinschmidt:
**Type of change**: bug fix

**Impact**: rtl change

**Release Notes**
Move all non-synthesizable constructs and models into the test harness and out of ChipTop

### DIFF
--- a/generators/chipyard/src/main/scala/IOBinders.scala
+++ b/generators/chipyard/src/main/scala/IOBinders.scala
@@ -9,6 +9,7 @@ import freechips.rocketchip.diplomacy.{LazyModule}
 import freechips.rocketchip.devices.debug._
 import freechips.rocketchip.subsystem._
 import freechips.rocketchip.system.{SimAXIMem}
+import freechips.rocketchip.amba.axi4.{AXI4Bundle, AXI4SlaveNode, AXI4EdgeParameters}
 import freechips.rocketchip.util._
 
 import sifive.blocks.devices.gpio._
@@ -182,6 +183,20 @@ object AddIOCells {
     port.suggestName("serial")
     (port, ios)
   }
+
+  def axi4(io: Seq[AXI4Bundle], node: AXI4SlaveNode): Seq[(AXI4Bundle, AXI4EdgeParameters, Seq[IOCell])] = {
+    io.zip(node.in).map{ case (mem_axi4, (_, edge)) => {
+      val (port, ios) = IOCell.generateIOFromSignal(mem_axi4, Some("iocell_mem_axi4"))
+      port.suggestName("mem_axi4")
+      (port, edge, ios)
+    }}
+  }
+
+  def blockDev(bdev: BlockDeviceIO): (BlockDeviceIO, Seq[IOCell]) = {
+    val (port, ios) = IOCell.generateIOFromSignal(bdev, Some("iocell_bdev"))
+    port.suggestName("bdev")
+    (port, ios)
+  }
 }
 
 // DOC include start: WithGPIOTiedOff
@@ -211,11 +226,25 @@ class WithSimSPIFlashModel(rdOnly: Boolean = true) extends OverrideIOBinder({
 })
 
 class WithSimBlockDevice extends OverrideIOBinder({
-  (system: CanHavePeripheryBlockDeviceModuleImp) => system.connectSimBlockDevice(system.clock, system.reset.asBool); Nil
+  (system: CanHavePeripheryBlockDeviceModuleImp) => system.bdev.map { bdev =>
+    val (port, ios) = AddIOCells.blockDev(bdev)
+    val harnessFn = (th: chipyard.TestHarness) => {
+      SimBlockDevice.connect(th.clock, th.reset.asBool, Some(port))(system.p)
+      Nil
+    }
+    Seq((Seq(port), ios, Some(harnessFn)))
+  }.getOrElse(Nil)
 })
 
 class WithBlockDeviceModel extends OverrideIOBinder({
-  (system: CanHavePeripheryBlockDeviceModuleImp) => system.connectBlockDeviceModel(); Nil
+  (system: CanHavePeripheryBlockDeviceModuleImp) => system.bdev.map { bdev =>
+    val (port, ios) = AddIOCells.blockDev(bdev)
+    val harnessFn = (th: chipyard.TestHarness) => {
+      BlockDeviceModel.connect(Some(port))(system.p)
+      Nil
+    }
+    Seq((Seq(port), ios, Some(harnessFn)))
+  }.getOrElse(Nil)
 })
 
 class WithLoopbackNIC extends OverrideIOBinder({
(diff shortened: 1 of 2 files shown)


generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -211,11 +226,25 @@ class WithSimSPIFlashModel(rdOnly: Boolean = true) extends OverrideIOBinder({
 })
 
 class WithSimBlockDevice extends OverrideIOBinder({
-  (system: CanHavePeripheryBlockDeviceModuleImp) => system.connectSimBlockDevice(system.clock, system.reset.asBool); Nil
+  (system: CanHavePeripheryBlockDeviceModuleImp) => system.bdev.map { bdev =>
+    val (port, ios) = AddIOCells.blockDev(bdev)
+    val harnessFn = (th: chipyard.TestHarness) => {
+      SimBlockDevice.connect(th.clock, th.reset.asBool, Some(port))(system.p)
```
colinschmidt, line 232: @davidbiancolin I'm not sure if this `asBool` will mess up your reset work. I assume we can make SimBlockDevice take an abstract reset if needed, but currently it only accepts `Bool`.

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -182,6 +183,20 @@ object AddIOCells {
...
+      port.suggestName("mem_axi4")
+      (port, edge, ios)
+    }}
+  }
+
+  def blockDev(bdev: BlockDeviceIO): (BlockDeviceIO, Seq[IOCell]) = {
+    val (port, ios) = IOCell.generateIOFromSignal(bdev, Some("iocell_bdev"))
+    port.suggestName("bdev")
+    (port, ios)
+  }
```
jerryz123, line 199: This isn't totally related to your changes, but I wonder if this pattern of adding stuff into the AddIOCells object is the best way to do this moving forwards.
These functions should be able to be inlined into the IOBInders themselves, thus localizing all the code for any IO to one place?

generators/chipyard/src/main/scala/config/RocketConfigs.scala
```diff
@@ -205,6 +205,24 @@ class SmallSPIFlashRocketConfig extends Config(
   new freechips.rocketchip.subsystem.WithCoherentBusTopology ++
   new freechips.rocketchip.system.BaseConfig)
 
+class SimAXIRocketConfig extends Config(
+  new chipyard.iobinders.WithUARTAdapter ++
+  new chipyard.iobinders.WithTieOffInterrupts ++
+  new chipyard.iobinders.WithSimAXIMem ++                       // drive the master AXI4 memory with a SimAXIMem
+  new chipyard.iobinders.WithTiedOffDebug ++
```
jerryz123, line 212: Is there a strong motivation for anyone to use this config, instead of the BlackBoxSimMem?

generators/chipyard/src/main/scala/IOBinders.scala
colinschmidt: Yeah I initially did that but then I needed the same sequence of code in both version of the `Sim` and `Model` devices so I hoisted them to make it DRY.

jwright6323: Yeah, the reason for this pattern is that you often need to reuse this part of the code for multiple `IOBinder`s.

generators/chipyard/src/main/scala/config/RocketConfigs.scala
colinschmidt: No idea. I needed to test my changes so I made a config that used it, and thought it would be easier to include it than not. We can remove it if its not useful.

generators/chipyard/src/main/scala/config/RocketConfigs.scala
```diff
@@ -205,6 +205,24 @@ class SmallSPIFlashRocketConfig extends Config(
   new freechips.rocketchip.subsystem.WithCoherentBusTopology ++
   new freechips.rocketchip.system.BaseConfig)
 
+class SimAXIRocketConfig extends Config(
+  new chipyard.iobinders.WithUARTAdapter ++
+  new chipyard.iobinders.WithTieOffInterrupts ++
+  new chipyard.iobinders.WithSimAXIMem ++                       // drive the master AXI4 memory with a SimAXIMem
```
jwright6323, line 211: Is the difference between this and simDRAM that this uses a 1-cycle magic memory and simDRAM is realistic? If so, it might be worth noting that.

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -182,6 +183,20 @@ object AddIOCells {
     port.suggestName("serial")
     (port, ios)
   }
+
+  def axi4(io: Seq[AXI4Bundle], node: AXI4SlaveNode): Seq[(AXI4Bundle, AXI4EdgeParameters, Seq[IOCell])] = {
```
jwright6323, line 187: Why does this return the edge? It doesn't seem to be used.

colinschmidt: it is used to construct both the `SimAXIMem` and the `SimDRAM`.

jwright6323: Oh I see it now. Is this `edge` just the bundle parameters? I think this is fine, it's just surprising to me to see a diplomacy edge getting passed around here.

colinschmidt: Its slightly more than the bundle parameters, but `SimAXIMem` takes in the whole edge while `SimDRAM` only takes in the bundle params.

generators/chipyard/src/main/scala/IOBinders.scala
davidbiancolin: I think it should be fine.

generators/chipyard/src/main/scala/config/RocketConfigs.scala
abejgonzalez: I think it is fine to have a config.

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -232,21 +261,38 @@ class WithSimNIC extends OverrideIOBinder({
 // accessible to the IOBinder
 // DOC include start: WithSimAXIMem
 class WithSimAXIMem extends OverrideIOBinder({
-  (system: CanHaveMasterAXI4MemPort with BaseSubsystem) => SimAXIMem.connectMem(system)(system.p); Nil
```
abejgonzalez, line 235: Are we going to remove the helper function / object?

colinschmidt: Which one? The AddIOCells function?

generators/chipyard/src/main/scala/config/RocketConfigs.scala
colinschmidt, line 211:
```suggestion
  new chipyard.iobinders.WithSimAXIMem ++                       // drive the master AXI4 memory with a SimAXIMem, a 1-cycle magic memory
```

colinschmidt: Is this what you want @jwright6323 ?

davidbiancolin: "realistic"

jwright6323: Sure, this is fine with me

generators/chipyard/src/main/scala/IOBinders.scala
abejgonzalez: Nevermind, I forgot that this was present in RC (not in TCIP)

PR #581, opened 2020-05-29
Add dramsim_lib as a dependency of hammer sims

colinschmidt:
Fix small issues with hammer-sim dut name and flags.

It would be cool to only build dramsim when our config includes
dramsim but this matches the way the other simulators handle this
in chipyard.
We also strip out empty CC flags because hammer doesn't prune these.
Its unclear whether the hammer should handle this itself or not,
if it eventually does we can remove this check.


abejgonzalez:
Cancelling CI (this isn't covered by CI).

PR #582, opened 2020-05-29
More fixes for post-syn/par hammer simulations

colinschmidt:
Symlink dramsim2 into hammer-sims rundirs
update some hammer generated makefile targets that were missed before
bump barstools to get an iocell fix

I'm currently running a design through the flow again to verify that the new iocell change works as intended. I've verified that dramsim now correctly find the ini files, and we have all the up-to-date hammer auto-generated makefile targets.

### DIFF
--- a/vlsi/Makefile
+++ b/vlsi/Makefile
@@ -99,6 +99,8 @@ SIM_TIMING_CONF = $(OBJ_DIR)/sim-timing-inputs.yml
 include $(vlsi_dir)/sim.mk
 $(SIM_CONF): $(VLSI_RTL) $(HARNESS_FILE) $(HARNESS_SMEMS_FILE) $(sim_common_files) $(dramsim_lib)
 	mkdir -p $(dir $@)
+	mkdir -p $(OBJ_DIR)/$(HAMMER_SIM_RUN_DIR)/$(notdir $(BINARY))
+	ln -sf $(base_dir)/generators/testchipip/src/main/resources/dramsim2_ini $(OBJ_DIR)/$(HAMMER_SIM_RUN_DIR)/$(notdir $(BINARY))/dramsim2_ini
 	echo "sim.inputs:" > $@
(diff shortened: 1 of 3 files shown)


vlsi/Makefile
```diff
@@ -99,6 +99,8 @@ SIM_TIMING_CONF = $(OBJ_DIR)/sim-timing-inputs.yml
 include $(vlsi_dir)/sim.mk
 $(SIM_CONF): $(VLSI_RTL) $(HARNESS_FILE) $(HARNESS_SMEMS_FILE) $(sim_common_files) $(dramsim_lib)
 	mkdir -p $(dir $@)
+	mkdir -p $(OBJ_DIR)/$(HAMMER_SIM_RUN_DIR)/$(notdir $(BINARY))
```
jwright6323, line 102: I like this; I think we should propagate this to `sims/vcs` and `sims/verilator`. Currently the outputs for `run-binary` go into the run directory

vlsi/Makefile
```diff
@@ -99,6 +99,8 @@ SIM_TIMING_CONF = $(OBJ_DIR)/sim-timing-inputs.yml
...
+	ln -sf $(base_dir)/generators/testchipip/src/main/resources/dramsim2_ini $(OBJ_DIR)/$(HAMMER_SIM_RUN_DIR)/$(notdir $(BINARY))/dramsim2_ini
```
jwright6323, line 103: is this not set as a resource with `setResource`? It feels like that would solve this automatically, but maybe you'd still need another `ln -s`

colinschmidt: Currently the dramsim blackbox in testchipip only looks in the current directory and since hammer runs the sims from a different directory based on all these factors we can't just commit the symlink like `sims/vcs` does.

vlsi/Makefile
colinschmidt: I think we should lump that into the `sims/vlsi` refactor (next release)

Issue #583, opened 2020-05-30
Make WithoutTLMonitors available to Config Strings


timsnyder-siv:
**Impact**: rtl

**Description**

[Originally discussed on FireSim mailing list]( https://groups.google.com/forum/#!topic/firesim/8loAr-29FbI)

When using FASED memory model in FireSim, if the Target memory interface width does not match FASED (64-bit at time of this issue) a TileLink width adapter is used to adjust the width between the Target and FASED.  This creates a small TileLink network.  If TLMonitor is attached to the network, `midas.passes.SimulationMapping` leaves the FIRRTL in an illegal state (contains `plusarg_reader` instantiations with no `extmodule` definition) and an uncaught exception is thrown in a subsequent `firrtl.passes.InferTypes` pass.

<details><summary>Observed Exception Backtrace</summary><p>

```
Form: HighForm
======== Finished Transform [Golden Gate] Simulation Mapping ========

======== Starting Transform midas.passes.xilinx.package$HostSpecialization$ ========
======== Starting Transform midas.passes.xilinx.package$DefineBUFGCE$ ========
------------------------------------------------------------------------------

... elide a number of uninteresting log lines ...

======== Finished Transform firrtl.passes.ResolveKinds$ ========

======== Starting Transform firrtl.passes.InferTypes$ =======
Exception in thread "main" firrtl.CustomTransformException: 
	at firrtl.stage.transforms.CatchCustomTransformExceptions.execute(CatchCustomTransformExceptions.scala:12)
	at firrtl.Transform.transform(Compiler.scala:222)
	at firrtl.Transform.transform$(Compiler.scala:222)
	at firrtl.stage.transforms.CatchCustomTransformExceptions.transform(CatchCustomTransformExceptions.scala:7)
	at firrtl.stage.transforms.UpdateAnnotations.$anonfun$internalTransform$2(UpdateAnnotations.scala:36)
	at firrtl.Utils$.time(Utils.scala:171)
	at firrtl.stage.transforms.UpdateAnnotations.internalTransform(UpdateAnnotations.scala:36)
	at firrtl.stage.transforms.UpdateAnnotations.internalTransform(UpdateAnnotations.scala:11)
	at firrtl.options.Translator.transform(Phase.scala:240)
	at firrtl.options.Translator.transform$(Phase.scala:240)
	at firrtl.stage.transforms.UpdateAnnotations.transform(UpdateAnnotations.scala:11)
[...]

PR #584, opened 2020-05-30
Alias RC WithoutTLMonitors into firesim.firesim

timsnyder-siv:
Need to add it to my GoldenGate Config String (-ggcs) otherwise
I'm seeing TLMonitors added during SimulationMapping transform in GoldenGate

Additional detail at:
https://groups.google.com/forum/\#!topic/firesim/8loAr-29FbI

**Related issue**: #583 

Not sure this is big enough to be considered a 'new feature' feel free to correct me if I'm wrong.

**Impact**: rtl change 

**Release Notes**
* alias `freechips.rocketchip.subsystem.WithoutTLMonitors` into `firesim.firesim` so that it can be used as a '_'-delimited ConfigString element.

### DIFF
--- a/generators/firechip/src/main/scala/TargetConfigs.scala
+++ b/generators/firechip/src/main/scala/TargetConfigs.scala
@@ -73,6 +73,10 @@ class WithNVDLALarge extends nvidia.blocks.dla.WithNVDLA("large")
 class WithNVDLASmall extends nvidia.blocks.dla.WithNVDLA("small")
 
 
+// Alias WithoutTLMonitors into this package so that it can be used in config strings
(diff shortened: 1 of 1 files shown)


timsnyder-siv:
@davidbiancolin please review

davidbiancolin:
Hey Tim, 

Thanks for your contribution. Could you add your alias here: https://github.com/firesim/firesim/blob/dev/sim/midas/src/main/scala/midas/Config.scala

And for now i think we should enable it by default by mixing it into SimConfig on line 69 so that TL -- you should be able to ++ <alias> if you add an extra level of Config() nesting; there are examples of this below in that file.

davidbiancolin:
Subsumed by firesim/firesim#577.

PR #587, opened 2020-05-31
Generalized Generator Interface for Easy Third-Party Core Integration

zitaofang:
**Related issue**: none

**Impact**: software change

**Release Notes**

- Replaced boilerplate code for integrating individual cores with generic code 
- Added `GenericConfig` which provides a generic method to update parameters across different cores


generators/chipyard/src/main/scala/TestSuites.scala
```diff
@@ -176,6 +95,7 @@ class TestSuiteHelper
         else            ((if (vm) rv32i else rv32pi), rv32u)
 
       addSuites(rvi.map(_("p")))
+      addSuites(rvu.map(_("p")))
       addSuites((if (vm) List("v") else List()).flatMap(env => rvu.map(_(env))))
```
jerryz123, line 99: Is giving all the cores the same test-suite generator function actually the right thing to do here? Does this still produce the same set of test-suites for each core?

generators/chipyard/src/main/scala/TestSuites.scala
```diff
@@ -64,91 +63,11 @@ class TestSuiteHelper
   def addSuites(s: Seq[RocketTestSuite]) { s.foreach(addSuite) }
 
   /**
-  * Add BOOM tests (asm, bmark, regression)
+  * Add generic tests (asm, bmark, regression) for all cores.
```
jerryz123, line 66: This comment is bit misleading if we make this function generic.
We only add tests for the core with hartid=0 (since only hart0 runs the asm tests).

generators/chipyard/src/main/scala/CoreManager.scala
```diff
@@ -0,0 +1,97 @@
...
+sealed trait CoreEntryBase {
+  def tileParamsLookup(implicit p: Parameters): Seq[TileParams]
+  def updateWithFilter(view: View, p: Any => Boolean): PartialFunction[Any, Map[String, Any] => Any]
+  def instantiateTile(crossingLookup: (Seq[RocketCrossingParams], Int) => Seq[RocketCrossingParams], logicalTreeNode: LogicalTreeNode)
+    (implicit p: Parameters, valName: ValName): Seq[(TileParams, RocketCrossingParams, BaseTile)]
+}
+
+// Implementation of third-party core entries
+class CoreEntry[TileParamsT <: TileParams with Product: TypeTag, TileT <: BaseTile : TypeTag](
+  tk: Field[Seq[TileParamsT]],
```
jerryz123, line 29: `tk` and `ck` seem like they can get better names

generators/chipyard/src/main/scala/ConfigFragments.scala
```diff
@@ -147,8 +147,10 @@ class WithControlCore extends Config((site, here, up) => {
...
 })
 
-class WithTraceIO extends Config((site, here, up) => {
-  case BoomTilesKey => up(BoomTilesKey) map (tile => tile.copy(trace = true))
-  case ArianeTilesKey => up(ArianeTilesKey) map (tile => tile.copy(trace = true))
-  case TracePortKey => Some(TracePortParams())
-})
+class WithTraceIO extends Config((site, here, up) =>
+  GenericConfig(Map("trace" -> true), {
+    case RocketTilesKey => false
```
jerryz123, line 152: Is there a reason why Rocket gets a special case here?

generators/chipyard/src/main/scala/CoreManager.scala
jerryz123: NVM I see what they represent

jerryz123:
- Have you tested this with existing BOOM/Ariane/Rocket configs?
- The GenericTestSuites function looks concerning. Is that function actually generic?
- I'd like to see this system used to reduce enumeration in a couple more places. There's a couple fragments in ConfigFragments.scala that enumerate all Tile types, can you see if those can be refactored to use this interface?
- Can you push this branch directly into the Chipyard repo? I've given you write access. That will run CI on it. You might have to close and re-open this PR

generators/chipyard/src/main/scala/TestSuites.scala
abejgonzalez: I don't think we should generalize the tests that each core runs, some cores will be RV64GC, others may not so I think it is important for the cores to determine which set of tests to run.

generators/chipyard/src/main/scala/TestSuites.scala
abejgonzalez: Agreed. We never really talk about this anywhere in the documentation either. I think it would be nice to add a better comment to this file and maybe add the "only runs on 1st hart" to the documentation.

generators/chipyard/src/main/scala/CoreManager.scala
abejgonzalez: Im still of the opinion that it should be spelled out.

generators/chipyard/src/main/scala/CoreManager.scala
```diff
@@ -0,0 +1,97 @@
...
+) extends CoreEntryBase {
+  // Use reflection to get the parameter's constructor
+  private val mirror = runtimeMirror(getClass.getClassLoader)
+  private val paramClass = mirror.runtimeClass(typeOf[TileParamsT].typeSymbol.asClass)
+  private val paramNames = (paramClass.getDeclaredFields map (f => f.getName)).zipWithIndex.toMap
+  private val paramCtr = paramClass.getConstructors.head
+
+  // Use reflection to get the tile's constructor
+  private val tileClass = mirror.runtimeClass(typeOf[TileT].typeSymbol.asClass)
+  private val tileCtr = tileClass.getConstructors.filter(ctr => ctr.getParameterTypes()(4) == classOf[Parameters]).head
```
abejgonzalez, line 40: Nit: `Ctr` in my mind normally means counter. Maybe use `cstr` `ctor`?

generators/chipyard/src/main/scala/TestSuites.scala
zitaofang: I have checked the code for Rocket, Boom and Ariane and it seems like they only have a few differences. Boom always executes `addSuites(env.map(rv64uf))` and also executes `addSuites(rvu.map(_("p")))` which no other core does. I tested Rocket and the extra suite `rvu` does not affect it, although I haven't test Boom yet. Maybe I can introduce another set of config to specify what testes they should use.

generators/chipyard/src/main/scala/TestSuites.scala
zitaofang: Sure. I will fix this comment in the next commit.

zitaofang:
Since Jerry has given me the access to this repo, I will push my branch into this repo and reopen another PR.

PR #588, opened 2020-06-01
Test Suite Simplification

zitaofang:
Check #587 for previous comments and change requests.

**Related issue**: None

**Impact**: software change

**Release Notes**

- Allow test suite to be configurable with config fragments


jerryz123:
Copying comments from the closed PR

Me:
- Have you tested this with existing BOOM/Ariane/Rocket configs?
- The GenericTestSuites function looks concerning. Is that function actually generic?
- I'd like to see this system used to reduce enumeration in a couple more places. There's a couple fragments in ConfigFragments.scala that enumerate all Tile types, can you see if those can be refactored to use this interface?

Alon:
- Generic is a somewhat confusing name, since the PR addresses only cores. I would prefer 
`addGenericTestSuites` -> `addCoreTestSuites`
`GenericConfig` -> `CoreConfig` or `TileConfig`

Abe:
- Nit: Ctr in my mind normally means counter. Maybe use cstr ctor?
- tk and ck seem like they can be spelled out
- I don't think we should generalize the tests that each core runs, some cores will be RV64GC, others may not so I think it is important for the cores to determine which set of tests to run.

generators/chipyard/src/main/scala/Subsystem.scala
```diff
@@ -53,19 +45,7 @@ trait HasChipyardTiles extends HasTiles
...
-          LazyModule(new RocketTile(r, crossing, PriorityMuxHartIdFromSeq(rocketTileParams), logicalTreeNode))
-        }
-        case b: BoomTileParams => {
-          LazyModule(new BoomTile(b, crossing, PriorityMuxHartIdFromSeq(boomTileParams), logicalTreeNode))
-        }
-        case a: ArianeTileParams => {
-          LazyModule(new ArianeTile(a, crossing, PriorityMuxHartIdFromSeq(arianeTileParams), logicalTreeNode))
-        }
-      }
+    case (param, crossing, tile) => {
```
jerryz123, line 48: Note that the sorting by hartid here was to guarantee that the Tiles  were constructed by order of hartid.
Now that you've hoisted the Tile construction out of this block, this is no longer true.

You should modify the definition of allTilesInfo to instantiateTiles by hartid.

jerryz123: Also, unrelated, but adding a `require` that no two Tiles have the same hartid would be a good sanity check to perform here.

zitaofang:
The current `GenericConfig` does not address updates to nested parameter class (like `CoreParams`), so I am redesigning it.

Also, I have investigated the previous version of `TestSuite.scala`. All three functions that add cases into the test suite are almost identical, and they would construct the suite depends on what subset of RISC-V is being used. They are copied three times because their `TilesKey` objects are different:
https://github.com/ucb-bar/chipyard/blob/c576a7e76717376b7a2cbbc0a531f51382ff87f1/generators/chipyard/src/main/scala/TestSuites.scala#L71

The Boom test suite function is slightly different from the other two at two places. It has an additional test case at
https://github.com/ucb-bar/chipyard/blob/c576a7e76717376b7a2cbbc0a531f51382ff87f1/generators/chipyard/src/main/scala/TestSuites.scala#L100
Rocket can still pass the test with this case, so I don't think having it in all cores instead of Boom only will break anything, although I still  need to test on Ariane to conclude.

Another place is here: for Boom,
https://github.com/ucb-bar/chipyard/blob/c576a7e76717376b7a2cbbc0a531f51382ff87f1/generators/chipyard/src/main/scala/TestSuites.scala#L81-L85
And for Rocket and Ariane,
https://github.com/ucb-bar/chipyard/blob/c576a7e76717376b7a2cbbc0a531f51382ff87f1/generators/chipyard/src/main/scala/TestSuites.scala#L121-L126
This two snippets are the code that deal with 64-bit base instruction set when FPU test cases are being added to the suite. Since Boom is a RV64GC core, the three cases here will always be added no matter which snippet is used, so using the code below for all cores will work. 

Therefore, it is OK to use the same function to construct the **default** test suite for all cores. If we need to have additional cases or use their own test suite instead of the generic one, I can make the `CoreManager` to load customized test suite if specified.

zitaofang:
This version should be able to handled nested parameter in TileParams and allow more fields to be changed dynamically. I can now change `WithL2TLBs` in `ConfigFragments.scala` to use the generic core config class. 

To address the concern about the single test suite for all cores, I made some changes in the test suite helper and allow people to define their own suites to replace the default test suite. I have tested the single test suite on Rocket, small Boom and dual small Boom, and all of them worked fine. I can compile Ariane with Verilator, but it's not functional (as warned in the docs); nonetheless, I verified the final TileParams for Ariane before it was passed to the tile constructor in both `dev` and `ariane-decouple` branches, and they were exactly the same.

I don't have EC2 F1 instance, so I didn't test Firesim. I will need to look at the CI check result and see if it works.

generators/chipyard/src/main/scala/Subsystem.scala
zitaofang: According to the comment, it seems like what matters is the order they are connected to the interrupt controller (which is done by `connectInterrupts()` in the remaining section in this file), so I think it is necessary to be there.

zitaofang: I forgot to included the sanity check in the commit I pushed, but I will put it into the code and push it in the next revision.

jerryz123: Can you modify core construction to be in order of hartid anyways? The test case which exposed this issue is not easy to recreate, and is not part of the regressions. Minimizing changes to functionality here would be ideal.

zitaofang: Sure. I will make another commit later.

zitaofang: Done.

davidbiancolin:
This seems closely related to: https://github.com/chipsalliance/rocket-chip/pull/2504/files

davidbiancolin:
I need to have a closer look at the implementation but my impression is:
1) There's a lot of additional software complexity that i don't think is necessary (see: reflection). 
2) there seems to be a few instances where a bunch of information needs to be centralized in a common object, which is somewhat self-defeating. The goal should be to be able to provide your own tile without modifying any of the sources. 
3) GenericConfigs underlying stringy implementation Map[String, Any] is concerning to me

Can you provide more context re: 
1) What problem GenericConfig solves
2) Address how this PR relates to henry's PR above.

davidbiancolin:
Also, how do you find using metals in Chipyard?

zitaofang:
I didn't know about the Rocket PR when I started working on this (I learned about that when I talked with Jerry this afternoon), but it seems like they finish the generic instantiation on their side. If they merge that PR, I can simplify my code that integrates tiles in `subsystem.scala`.

The current design of `CoreManager` does require modification to the code in this repo for new third-party cores, but to address this problem, I would add a method so that those cores can register themselves.

GenericConfig was intended to provide a way to change parameter for all registered cores without enumerating all cores in the config class. `WithTraceIO` in `ConfigFragments.scala` is one of these classes. It is used in a firesim config class `WithFireSimConfigTweaks`, which is designed to be attached to *any* cores and change their parameters for running on firesim (unless I misunderstood it). The current implementation of `WithTraceIO` in `dev` requires a list of all cores that it may be attached to, otherwise the cores will need to create their own version of this trait and attach it separately, which makes `WithFireSimConfigTweaks` meaningless. 

My current solution to this problem using reflection does look ugly, but that is the only way to achieve this without touching the Rocket code. I do have a better solution, but this would involve changes to a common interface and may be difficult. It would need a method to TilesParam and CoreParams (the two base traits of parameter classes) that construct a copy of that class with the specified fields visible in the traits modified. With this method, I won't need the actual type info of the parameter subclasses to copy them. If that method have a default implementation in the base trait that throw a `NotImplementedError` (use `=???`), then it won't be a breaking change unless they upgrade to the new Chipyard. 

Anyway, I would need to wait for the result of the Rocket PR before I made any change to my code.

zitaofang:
I simplified the generic config system in this repo. This version would not be able to handle the traceport config fragment since `trace` is not in the base trait TileParams, and I will need to create a new trait for those cores that need to enable traceports. This change is somewhat unrelated to this PR, and I would open another once this is merged. 

The .metal related lines in .gitignore is the auto-generated files by VSCode editor with scala plugins, and it seems like .gitignore was committed when I used `-a` flag. I would take that out in the next commit since it is unrelated, but it would be useful if someone decide to switch to VSCode in the future. Maybe another PR?

generators/chipyard/src/main/scala/GenericParams.scala
```diff
@@ -0,0 +1,177 @@
...
+  val fetchWidth: Int,
+  val decodeWidth: Int,
+  val retireWidth: Int,
+  val instBits: Int,
+  val nLocalInterrupts: Int,
+  val nPMPs: Int,
+  val pmpGranularity: Int,
+  val nBreakpoints: Int,
+  val useBPWatch: Boolean,
+  val nPerfCounters: Int,
```
jerryz123, line 70: I'm not sure this is a good idea. This object seems to attempt to create a superset of all possible core parameters, which is not a sustainable approach,

jerryz123:
Given that the new RC Tile system addresses some of the original motivations for sweeping Tile changes, I think we should rescope this PR to focus on addressing the TestSuites generation duplication.

generators/chipyard/src/main/scala/GenericParams.scala
zitaofang: This class is intended to provide a `copy()` function only for the core parameters in `TileParams` or `CoreParams`, which would be used by all cores. Chipyard is not supposed to change non-standard, core-specific parameters with generic config fragments since we don't even know what they means for different cores.

However, if the user want to create their own generic config fragments for their family of cores which contains some shared parameters not included in the standard set of parameter classes, then they can put those shared parameters in a trait and use the similar technique in this file to implement them. 

For example, there is a config fragment that enables trace port (`WithTraceIO`) in Chipyard. The parameter `trace` is not included in `TileParams` and is not used by Rocket, but it is intended to be generic for at least the cores that support toggleable trace ports generation, and it is used by other general purpose config fragments in FireSim. A new trait containing `trace` only and inherited by Boom and Ariane (which support toggleable trace port generation) can be created along with a generic case class to generalize this config fragment.

zitaofang:
After talking with Jerry, I am re-purposing this PR to include the improvement on test suite only.

jerryz123:
@colinschmidt can you check the changes for hwacha tests in this PR? It seems more elegant to me.

Issue #591, opened 2020-06-03
loadMemoryFromFile broken by Chipyard transforms


baltazarortiz:
**Impact**: rtl

**Tell us about your environment:**
*Chipyard Version:* 1.3.0
*OS:* Ubuntu 18.04
*Other:* 

* My branch demonstrating this bug by adding a `TLRAM` to the GCD example is [here](https://github.com/baltazarortiz/chipyard/tree/TLRAM_loading_bug).

* [Link I referenced that does the same thing in Rocket](https://jia.je/hardware/2020/03/17/rocket-chip-tlram-load/)

**What is the current behavior?**
* I instantiated a `TLRAM` inside my module (GCD in the example branch) and then pointed `loadMemoryFromFile` at it.

* The `TLRAM` shows up in the memory map as expected, and the annotations JSON file shows a `LoadMemoryAnnotation`. However, `LoadMemoryTransform` is not run.

* I tried adding `LoadMemoryTransform` to `TopTransforms` and/or `HarnessTransforms`, and it made no difference.

**What is the expected behavior?**
* `loadMemoryFromFile` should load a custom `TLRAM` despite the other Chipyard transforms.

**Other information**

Based on discussion in the FIRRTL chat and Chipyard mailing list, it seems that the issue is some sort of interaction between `ReplSeqMems` or the memory macrocompiler things, neither of which I am super well versed on.

However, when I removed all references to those from the makefiles, my TLRAM did get the transform applied. It was still split into 8 pieces though, which made it slightly nonobvious how I would have had to fill out the 8 files to get my memory loaded contiguously (is it just file 0 through file 7 going low to high address?). I tried using the `SimplifyMems` transform, but it didn't seem to do anything.

This also broke the actual Chipyard memory system though, so is not an actual solution, just showed that this is most likely at least part of the issue.

baltazarortiz:
(If this should be an issue against barstools instead, let me know!)

HaFred:
Is it carried out on verilator backend? How is it going now?

baltazarortiz:
Not sure what you mean re: verilator, but I still don't have this working. With thesis deadline slowly but steadily approaching I am just working around this for now with a blackbox peripheral with a giant memory in it that I can manually load. Works for my needs thankfully, but very much not integrated with the Chipyard memory system.

abejgonzalez:
@baltazarortiz Is this issue solved with this PR? https://github.com/ucb-bar/chipyard/pull/635

[...]

PR #592, opened 2020-06-03
[WIP] Load simulator memory from file

baltazarortiz:
**Related issue**: n/a

**Impact**: software change

**Release Notes**

**WIP**: Still needs debugging, but discussing this and #591 in the same mailing list thread was getting confusing. I'm guessing my c++ isn't great but I'm happy to clean it up after the functionality is correct!

In this branch + the branch of testchipip that is currently being pointed to for testing purposes, I have added command line support to initialize the simulated DRAM.

Usage (also added to the simulator help output): `./simulator-chipyard-RocketConfig +dramsim +fastloadmem --loadmem=/path/to/test.hex /path/to/test.riscv`

I attempted to do this as similarly as possible to the way that Firesim does (see [here](https://github.com/firesim/firesim/blob/b4951f2589114e3849c77c798982afd10fc11b3e/sim/midas/src/main/cc/simif_emul.cc#L86) and [here](https://github.com/firesim/firesim/blob/c2d8e3a46e59222e115a1fdaa7267592e1d3c503/sim/midas/src/main/cc/emul/mmio.cc)). Currently, the load finishes, but then the simulator segfaults during the first call to `memory_tick()`.

I have tried passing a file from [generate_memory_init.py](https://github.com/ucb-bar/midas-examples/blob/master/scripts/generate_memory_init.py) as well as a file that only has a single line, and I get the same result. I also get the same result whether or not I pass the `+dramsim` flag (though it does show the dramsim ini files being loaded if I do).

### DIFF
--- a/.circleci/do-rtl-build.sh
+++ b/.circleci/do-rtl-build.sh
@@ -53,13 +53,18 @@ else
 fi
 
 # enter the verilator directory and build the specific config on remote server
-run "make -C $REMOTE_SIM_DIR clean"
 run "export RISCV=\"$TOOLS_DIR\"; \
      export LD_LIBRARY_PATH=\"$LD_LIB_DIR\"; \
      export PATH=\"$REMOTE_VERILATOR_DIR/bin:\$PATH\"; \
      export VERILATOR_ROOT=\"$REMOTE_VERILATOR_DIR\"; \
      export COURSIER_CACHE=\"$REMOTE_WORK_DIR/.coursier-cache\"; \
+     make -C $REMOTE_SIM_DIR clean; \
      make -j$REMOTE_MAKE_NPROC -C $REMOTE_SIM_DIR JAVA_ARGS=\"$REMOTE_JAVA_ARGS\" ${mapping[$1]}"
+run "rm -rf $REMOTE_CHIPYARD_DIR/project"
+
+# copy back the final build
+
+
 run "rm -rf $REMOTE_CHIPYARD_DIR/project"
 
 # copy back the final build
--- a/.circleci/run-tests.sh
+++ b/.circleci/run-tests.sh
@@ -93,10 +93,10 @@ case $1 in
         make -C $LOCAL_SIM_DIR ${mapping[$1]} BINARY=$LOCAL_CHIPYARD_DIR/tests/nvdla.riscv run-binary
         ;;
     icenet)
-        make run-none-fast -C $LOCAL_SIM_DIR ${mapping[$1]}
+        make run-binary-fast BINARY=none -C $LOCAL_SIM_DIR ${mapping[$1]}
         ;;
     testchipip)
-        make run-none-fast -C $LOCAL_SIM_DIR ${mapping[$1]}
+        make run-binary-fast BINARY=none -C $LOCAL_SIM_DIR ${mapping[$1]}
         ;;
     *)
         echo "No set of tests for $1. Did you spell it right?"
--- a/.githooks/ignore-certain-dirs-commit-msg
+++ b/.githooks/ignore-certain-dirs-commit-msg
@@ -8,9 +8,9 @@ fi
 changes=( `git diff --name-only --cached` )
 
 # Load the patterns we want to skip into an array
-mapfile -t blacklist < .ciignore
+mapfile -t blocklist < .ciignore
 
-for i in "${blacklist[@]}"
+for i in "${blocklist[@]}"
 do
 	# Remove the current pattern from the list of changes
 	changes=( ${changes[@]/$i/} )
--- a/.gitmodules
+++ b/.gitmodules
@@ -3,7 +3,7 @@
 	url = https://github.com/ucb-bar/rocket-chip.git
 [submodule "testchipip"]
 	path = generators/testchipip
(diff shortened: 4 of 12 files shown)


davidbiancolin:
Cool! You can have a look at @colinschmidt's pr #598 to see how to more cleaning add new plusargs.

baltazarortiz:
Thanks! I'll check that out. Any ideas about the segfault? Is there a different format of input file I should be using?

baltazarortiz:
I switched the plusarg parsing to match #598, thanks for pointing this out! 

I'm still unsure about why the model segfaults as soon as the load finishes. I remembered that I had to replace `MEM_DATA_BITS` with a constant - is there any equivalent in Chipyard to the [automatic header generation in midas](https://github.com/firesim/firesim/blob/c2d8e3a46e59222e115a1fdaa7267592e1d3c503/sim/midas/src/main/scala/midas/core/FPGATop.scala#L281)? I didn't see anything obvious.

baltazarortiz:
I thought I had a hack that worked for my system (blackboxed MMIO peripheral containing a huge memory I could directly initialize in the verilog), but it's looking to be insufficient. @davidbiancolin any ideas on things I should look into to debug this approach crashing / any examples of what using this flag on firesim looks like? I still haven't been able to find any examples.

zhemao:
This line here needs one more level of indirection.

    load_mem((void**)mm->get_data(), loadmem.c_str(), 64 / 8, 1);

The `load_mem` function expects an array of pointers, one to each memory channel. If you don't care about multichannel, you can change it to this:

    load_mem((void **) &mm->get_data(), loadmem.c_str(), 64, 1);

Also, note that the line_size argument is specified in bytes and not in words.

zhemao:
The proper way of doing this that makes it compatible with multichannel would be a bit trickier. We'd have to rearchitect SimDRAM so that it knows about all the channels.

baltazarortiz:
Thanks! I'll check this out. I personally do not need multichannel, just some way to load memory. It definitely looks like making this PR fully compatible in general would require a lot of changes to automatically pull parameters out of Chipyard, since the ones the firesim implementation grabs things from headers that come out of Midas/golden gate if I understand correctly?

baltazarortiz:
To confirm, is the format of file that comes out of [generate_memory_init.py](https://github.com/ucb-bar/midas-examples/blob/master/scripts/generate_memory_init.py) the correct format to load via this function?

zhemao:
Yes, it should work. The format is just a list of hex numbers in little-endian order.

baltazarortiz:
I updated the argument to be a double pointer and changed line_size to be in bytes. It isn't segfaulting now but also doesn't seem to be loading anything.

I tried providing it a file that is entirely filled with `f`'s and then running a hello world + memory print:

```
#include <stdio.h>                                                              
#include <stdint.h>                                                             
int main() {                                                                    
        printf("Hello, World!\n");                                                   
              
        #define RAM ( *(volatile uint32_t*)0x80000000 )                                                 
        printf("RAM val: %x\n", RAM);                                           
                     
        return 0;                                                               
} 
```

However, the print is consistently `RAM val: 41014081`. My SoC is set up with a default memory starting at `0x80000000`, but I'm linking my application to start at `0x85000000`. I'll make a separate TLRAM for loading my program into to see if something about the way FESVR loads applications is overwriting the initialized memory.

baltazarortiz:
I wasn't able to get the TLRAM to store the program I'm trying to run, but I did find that if I change my link address to `0x80000000`, the program still runs fine, which presumably means no memory is being loaded (I would assume that the program would be overwritten and not be able to run?)

zhemao:
HTIF generally runs after loadmem, so it's most likely overwriting the memory with the program.

Are you invoking it with the correct arguments? In the invocation you provided, you use `--loadmem=`, but your code is expecting `+loadmem`.

baltazarortiz:
That makes sense, does HTIF start writing from the beginning of memory though? I would expect that to mean that I'd see significantly slower load times for loading at `0x85000000` vs `0x80000000`, but (without timing exactly) that doesn't seem to be the case.

I am, good catch - `--loadmem=` was my initial implementation before I updated based on #598. I'm invoking with 


```
make run-binary-debug CONFIG=MyRocketConfig BINARY=../../tests/hello.riscv SIM_FLAGS="+fastloadmem +loadmem=/home/ubuntu/test.hex"
```

which results in the simulator invocation being

```
/home/ubuntu/chipyard/sims/verilator/simulator-chipyard-MyRocketConfig-debug +permissive +fastloadmem +loadmem=/home/ubuntu/test.hex +dramsim +dramsim_ini_dir=/home/ubuntu/chipyard/generators/testchipip/src/main/resources/dramsim2_ini +max-cycles=10000000  +verbose -v/home/ubuntu/chipyard/sims/verilator/output/chipyard.TestHarness.MyRocketConfig/hello.vcd +permissive-off ../../tests/hello.riscv
```

The print statement I added indicating that the fastloadmem is happening is also popping up, so the code is definitely being reached. Just unsure if the load is happening correctly.

colinschmidt:
Closing this as the dev branch now includes #635 and #639 which should allow you to use loadmem and also have HTIF skip memory that has been loaded via the sideband.

PR #596, opened 2020-06-05
Use output_dir for run-binary logs and waveforms

jerryz123:
* Dump run-binary files in `output/$(long_name)` instead of current directory
 * Remove run-none rules, these were equivalent to run-binary BINARY=none

**Impact**: software change

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -129,34 +129,30 @@ verilog: $(sim_vsrcs)
 # helper rules to run simulations
 #########################################################################################
 .PHONY: run-binary run-binary-fast run-binary-debug run-fast
-run-binary: $(sim)
+run-binary: $(output_dir) $(sim)
 	(set -o pipefail && $(sim) $(PERMISSIVE_ON) $(SIM_FLAGS) $(EXTRA_SIM_FLAGS) $(VERBOSE_FLAGS) $(PERMISSIVE_OFF) $(BINARY) </dev/null 2> >(spike-dasm > $(sim_out_name).out) | tee $(sim_out_name).log)
 
 #########################################################################################
 # helper rules to run simulator as fast as possible
 #########################################################################################
-run-binary-fast: $(sim)
+run-binary-fast: $(output_dir) $(sim)
 	(set -o pipefail && $(sim) $(PERMISSIVE_ON) $(SIM_FLAGS) $(EXTRA_SIM_FLAGS) $(PERMISSIVE_OFF) $(BINARY) </dev/null | tee $(sim_out_name).log)
 
 #########################################################################################
 # helper rules to run simulator with as much debug info as possible
 #########################################################################################
-run-binary-debug: $(sim_debug)
+run-binary-debug: $(output_dir) $(sim_debug)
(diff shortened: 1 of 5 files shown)


common.mk
```diff
@@ -141,22 +141,18 @@ run-binary-fast: $(sim)
...
-
 #########################################################################################
 # run assembly/benchmarks rules
 #########################################################################################
-$(output_dir)/%: $(RISCV)/riscv64-unknown-elf/share/riscv-tests/isa/%
-	mkdir -p $(output_dir)
+$(output_dir):
+	mkdir -p $@
+
+$(output_dir)/%: $(RISCV)/riscv64-unknown-elf/share/riscv-tests/isa/% $(output_dir)
```
colinschmidt, line 155: Nice this always annoyed me.

zhemao:
I'm not sure about getting rid of the output dependencies none.out/none.run/none.vpd. The difference between "run-none" and "run-binary BINARY=none" is that the former will only execute if the simulation executable has changed, while the latter will run every time. So it seems that the tradeoff is between having this dependency tracking feature versus having a bunch of extra rules that may need to be changed if the arguments for running simulations drastically change in the future. @jwright6323 @colinschmidt What do you think? My vote is to not change it for now.

colinschmidt:
I'm usually in favor of these DRY changes but if the issue Howie mentioned can't be worked around and thats a feature people use then I think we should not remove it.

jerryz123:
Is that actually a useful feature though? In what cases would a user request `make run` without wanting to rerun the simulation?

colinschmidt:
I wouldn't do that with `none` although sometimes I do `make run` where only the binary has changed and it no rerunning lets me know I didn't update the binary as expected.

jerryz123:
That functionality currently only exists for the asm/bmark tests. I still don't see a use-case for this dependency tracking for `none`, since `run-none` has no binary to depend on.

zhemao:
The dependency is on the VCS/Verilator executable.

jerryz123:
In what case would you type `make run-none` and NOT want the simulation to run? That seems to be the only functionality which is lost here.

zhemao:
As Colin mentioned, if you think the VCS binary should have changed, but `make run-none` does nothing, it tells you the binary hasn't actually changed. I realize this is a pretty rare case though.

jerryz123:
Oh sorry, I was confused between the simulator binary, and the test binary (which does not exist for the none rules). That does seem like a marginally useful feature.

However, I argue that the DRYness is useful here, as the list of simulator arguments is a easy place for bugs to show up. Note that the existing `run-none-fast` rule has a bug in its construction of the simulator arguments, 
```
$(output_dir)/none.run: $(sim)
    (set -o pipefail && $(sim) $(PERMISSIVE_ON) $(SIM_FLAGS) $(EXTRA_SIM_FLAGS) $(PERMISSIVE_OFF) $< </dev/null | tee $<.log) && touch $@
```

zhemao:
Maybe we should talk about this at the next chipyard meeting.

jerryz123:
Sure

common.mk
```diff
@@ -141,22 +141,18 @@ run-binary-fast: $(sim)
 #########################################################################################
 # helper rules to run simulator with as much debug info as possible
 #########################################################################################
-run-binary-debug: $(sim_debug)
```
zhemao, line 144: You need to add the output_dir dependency to the other run-binary rules as well.

PR #598, opened 2020-06-06
Allow dramsim_ini folder to be set at the command line

colinschmidt:
verilator supports the amount of `vpi` we need to do our own argument parsing without messing with emulator.cc every time.


colinschmidt:
Rocket-chip doesn't support permissive (so doesn't support arbitrarty additions or verilator runtime arguments), it also doesn't support TSI.
We could probably crib the way they do async reset rising pulse, and then it would be a smaller diff but ours still adds several useful features for chipyard.

PR #601, opened 2020-06-10
Warn if RISCV unset

jerryz123:
Does anyone foresee problems with this? Technically it was possible to run Hammer with $RISCV unset, but user who want to do that can just set RISCV to a dummy variable.

**Impact**: software change

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -3,6 +3,14 @@
 #########################################################################################
 SHELL=/bin/bash
 
+
+ifndef RISCV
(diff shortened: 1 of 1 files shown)


colinschmidt:
I don't really have a problem with this. I've set `RISCV` to an empty string in other instances where I don't care and know that what I'm doing will be fine.

jwright6323:
If you're using hammer through chipyard you're likely using RISCV at some point, so I don't foresee this causing any trouble.

jerryz123:
Lol turns out this breaks all the CI

PR #612, opened 2020-06-20
[firechip] Add a small target that should fit on all hosts

davidbiancolin:
This is a convenience for eventually getting zynq back up in FireSim. 

**Related issue**: 

**Impact**: other 

**Release Notes**
- Add a small FireChip config that should fit on all FireSim hosts


davidbiancolin:
It also removed TraceIO. I wanted something flat so that it wouldn't change underneath me.

PR #613, opened 2020-06-22
Bloop Support 

davidbiancolin:
This PR brings up support for [bloop](https://scalacenter.github.io/bloop/) to do scala source compiles and for invoking scala executables. Builds/runs with bloop are enabled by setting ENABLE_BLOOP=1. I'll add documentation explaining installation.
 
The Bloop flow is faster across the board, but for a common iteration loop, where i touch a scala source (IOBinder.scala) and regenerate verilog (`make verilog` in sims/verilator) i've measured:

F1 c4.4xlarge: 5m21s SBT, 3m04s Bloop  (2 previous elaborations)
My 2019 mbpro:  3m4s SBT, 1m47s Bloop  (1 previous elaboration)

Neither build tool recompiles sources since i've only touched the files, but bloop seems to be ~30% faster compiling chipyard's sources.  For the figures above, much of the win comes from not having to relaunch SBT repeatedly, but having a long-lived JVM means the JIT can warm up and stay warm. For smaller designs the speed up becomes more profound as the SBT launches don't scale in design size. 

Unrelated fixes:
- Stop re-initializing firesim in firesim-setup.sh
- Delete dead `CLASSES` variables in Makefile

**Impact**:  other

**Release Notes**
- Support optionally using bloop as the scala build tool

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -34,6 +34,8 @@ lookup_srcs = $(shell find -L $(1)/ -name target -prune -o -iname "*.$(2)" -prin
 SOURCE_DIRS = $(addprefix $(base_dir)/,generators sims/firesim/sim tools/barstools/iocell)
 SCALA_SOURCES = $(call lookup_srcs,$(SOURCE_DIRS),scala)
 VLOG_SOURCES = $(call lookup_srcs,$(SOURCE_DIRS),sv) $(call lookup_srcs,$(SOURCE_DIRS),v)
+# This assumes no SBT meta-build sources
+SBT_SOURCES = $(call lookup_srcs,$(base_dir),sbt)
 
 #########################################################################################
 # rocket and testchipip classes
--- a/common.mk
+++ b/common.mk
@@ -60,11 +60,19 @@ $(FIRRTL_TEST_JAR): $(call lookup_srcs,$(CHIPYARD_FIRRTL_DIR),scala)
 	cp -p $(CHIPYARD_FIRRTL_DIR)/utils/bin/firrtl-test.jar $@
 	touch $@
 
+
+#########################################################################################
+# Bloop Project Definitions
+#########################################################################################
+$(BLOOP_CONFIG_DIR)/TIMESTAMP: $(base_dir)/build.sbt
+	cd $(base_dir) && $(SBT) "project chipyardRoot" "bloopInstall"
+	touch $@
+
 #########################################################################################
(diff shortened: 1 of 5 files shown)


common.mk
```diff
@@ -60,11 +55,19 @@ $(FIRRTL_TEST_JAR): $(call lookup_srcs,$(CHIPYARD_FIRRTL_DIR),scala)
...
+	cd $(base_dir) && $(SBT) "project chipyardRoot" "bloopInstall"
+	touch $@
+
 #########################################################################################
 # create list of simulation file inputs
 #########################################################################################
-$(sim_files): $(call lookup_srcs,$(base_dir)/generators/utilities/src/main/scala,scala) $(FIRRTL_JAR)
-	cd $(base_dir) && $(SBT) "project utilities" "runMain utilities.GenerateSimFiles -td $(build_dir) -sim $(sim_name)"
+$(sim_files): $(call lookup_srcs,$(base_dir)/generators/utilities/src/main/scala,scala) $(FIRRTL_JAR) $(SCALA_BUILDTOOL_DEPS)
+	$(call run_scala_main,utilities,utilities.GenerateSimFiles,-td $(build_dir) -sim $(sim_name))
```
colinschmidt, line 70: Is there any difference is variable expansion when moving from double quotes to parens?

variables.mk
```diff
@@ -125,9 +125,24 @@ JAVA_ARGS ?= -Xmx$(JAVA_HEAP_SIZE) -Xss8M -XX:MaxPermSize=256M
...
 
+BLOOP ?= bloop
+BLOOP_CONFIG_DIR ?= $(base_dir)/.bloop
+
+SCALA_BUILDTOOL_DEPS = $(SBT_SOURCES)
+
+ifdef ENABLE_BLOOP
+override SCALA_BUILDTOOL_DEPS += $(BLOOP_CONFIG_DIR)/TIMESTAMP
+define run_scala_main
+	cd $(base_dir) && bloop run $(shell echo $(1) | sed 's/{.*}//') --main $(2) -- $(3)
```
colinschmidt, line 138: What is the sed command filtering out here?

davidbiancolin: The `{file:/<path>} ` that needs to be prepending to provide a project URI in SBT but not for bloop.

davidbiancolin: That sometimes* needs to be.

PR #614, opened 2020-06-23
Diplomatic multiclock

jerryz123:
**Related issue**: 
https://github.com/chipsalliance/rocket-chip/pull/1795

**Impact**: rtl change 

- [x] https://github.com/ucb-bar/ariane-wrapper/pull/8
- [x] https://github.com/riscv-boom/riscv-boom/pull/481
- [x] https://github.com/ucb-bar/hwacha/pull/22
- [x] https://github.com/ucb-bar/testchipip/pull/96
- [x] https://github.com/firesim/firesim/pull/623


generators/chipyard/src/main/scala/ChipTop.scala
```diff
@@ -4,8 +4,11 @@ import chisel3._
 
 import scala.collection.mutable.{ArrayBuffer}
 
+import freechips.rocketchip.prci._
+import freechips.rocketchip.diplomacy._
+import freechips.rocketchip.subsystem.{BaseSubsystem}
 import freechips.rocketchip.config.{Parameters, Field}
-import freechips.rocketchip.diplomacy.{LazyModule}
+import freechips.rocketchip.diplomacy.{LazyModule, LazyModuleImp, LazyRawModuleImp, LazyModuleImpLike}
```
colinschmidt, line 11: double import of diplomacy.

generators/chipyard/src/main/scala/ChipTop.scala
```diff
@@ -4,8 +4,11 @@ import chisel3._
 
 import scala.collection.mutable.{ArrayBuffer}
 
+import freechips.rocketchip.prci._
```
colinschmidt, line 7: Do we really need all of prci?

generators/chipyard/src/main/scala/ChipTop.scala
```diff
@@ -49,71 +52,121 @@ case object GlobalResetAsynchronousFull extends GlobalResetScheme with HasAsyncI
...
+      iocells ++= _iocells.flatten
+      harnessFunctions ++= _harnessFunctions.flatten
+    }
+
+    // Set the System clock/reset to the diplomatic clock bundle
+    systemClock := systemClockSinkNode.in.head._1.clock
+    systemReset := systemClockSinkNode.in.head._1.reset
+
+    // The ModuleImp might be a LazyRawModule, in which case there is no clock
+    lSystem.module match { case l: LazyModuleImp => {
```
colinschmidt, line 114: Is the idea that they would have a custom ChipTop or use binders to drive the clock of the lSystem themselves?

generators/chipyard/src/main/scala/ChipTop.scala
```diff
@@ -49,71 +52,121 @@ case object GlobalResetAsynchronousFull extends GlobalResetScheme with HasAsyncI
...
-    iocells ++= _iocells.flatten
-    harnessFunctions ++= _harnessFunctions.flatten
-  }
 
+  // The provider of all clock groups. Subclasses define this
+  def provideDiplomaticClocks(): Unit
+
+  // This function generates RTL that may be necessary to drive the diplomatic clocks
+  // Subclasses define this
+  def driveClockGroups(systemClock: Clock, systemReset: Reset): (Seq[IOCell], Seq[TestHarnessFunction])
```
colinschmidt, line 76: I wonder if we often will need to use systemClock and systemReset in this function.
It seems like a recipe for many self-inflicted wounds where someone uses systemClock or more insidiously systemReset to drive the diplomatic clocks including what will become systemClock/Reset.

generators/chipyard/src/main/scala/ChipTop.scala
jerryz123: I guess there's three ways to specify clock input.

1. As implicit Chisel clocks. The current setup supports this case.
2. As a `ClockNode` within the system. The current code matches `BaseSubsystem`, but I think it should more accurately match `HasConfigurablePRCILocations` to match any `System` which might provide diplomatic clocks.
3. As some other named IO. I hope that we can define some generic "ChipTopClockingKey" to provide enough flexibility to support these cases.

generators/chipyard/src/main/scala/ChipTop.scala
jerryz123: Yeah this is pretty confusing already. It looks like in this example the `ResetCatchAndSync` can be clocked to the incoming clock from the TestHarness. 
It also seems like reset-generation can be refactored to be more elegantly tied to clock-generation.

jerryz123:
I cleaned this up substantially. Changes include

- `ChipTop` is now `LazyModule`, as it contains `ClockNodes`
- `ChipyardClockKey` lets you define a function to drive the diplomatic clocks of the system
- Two SW-simulation demonstrations of `ChipyardClockKey`, `harnessClock` just passes through the clock from the `TestHarness`, and passes it everywhere. `harnessMultiClock` divides the `harnessClock`, and uses that for the uncore
- `MultiClockRocketConfig` uses the `harnessMultiClock` setting, but it actually doesn't simulate correctly, because `SimSerial` and `SimDRAM` receive the fast TestHarness clock, instead of the divided clock. The fix for this will come later, as it is more involved. 
- `ConfigValName` is removed, I think this also resolves the naming issues with `TestHarness` appearing everywhere
- `FireSimMultiClockPOC` is removed, as the base `FireSim` design now supports multiclock through the `ChipyardClockKey`.
- `FireSimMulticlockRocketConfig` demonstrates the old `FireSimMulticlockPOC` config
- `FireSim` now builds `ChipTop`, instead of `DigitalTop`, removing a bit of code duplication.

.circleci/do-rtl-build.sh
```diff
@@ -59,7 +59,7 @@ run "export RISCV=\"$TOOLS_DIR\"; \
      export VERILATOR_ROOT=\"$REMOTE_VERILATOR_DIR\"; \
      export COURSIER_CACHE=\"$REMOTE_WORK_DIR/.coursier-cache\"; \
      make -C $REMOTE_SIM_DIR clean; \
-     make -j$REMOTE_MAKE_NPROC -C $REMOTE_SIM_DIR JAVA_ARGS=\"$REMOTE_JAVA_ARGS\" ${mapping[$1]}"
+     make -j$REMOTE_MAKE_NPROC -C $REMOTE_SIM_DIR FIRRTL_LOGLEVEL=info JAVA_ARGS=\"$REMOTE_JAVA_ARGS\" ${mapping[$1]}"
```
colinschmidt, line 62: Does this slow down CI at all?

generators/chipyard/src/main/scala/ChipTop.scala
```diff
@@ -4,118 +4,64 @@ import chisel3._
...
+  lazy val module: LazyModuleImpLike = new LazyRawModuleImp(this) {
+    // These become the implicit clock and reset to the System
+    val implicit_clock = implicitClockSinkNode.in.head._1.clock
+    val implicit_reset = implicitClockSinkNode.in.head._1.reset
+
+
+    // The implicit clock and reset for the system is also, by convention, used for all the IOBinders
+    // TODO: This may not be the right thing to do in all cases
+    withClockAndReset(implicit_clock, implicit_reset) {
+      val (_ports, _iocells, _harnessFunctions) = p(IOBinders).values.flatMap(f => f(lSystem) ++ f(lSystem.module)).unzip3
```
colinschmidt, line 54: I think we will want to understand how to generate the clock for the test harness side of things.
I don't think we want to have the parts inside the system punch their clock into the test harness.
Either the testharness module should be able to recover the clock from an explicit pin that would exist in a real ASIC (e.g. DDR or serial links with clock recovery), or it should explicitly require some configuration of the test harness component to work correctly (e.g. uart baud rate).

But we can discuss this more later.

generators/chipyard/src/main/scala/ChipTop.scala
```diff
@@ -4,118 +4,64 @@ import chisel3._
...
-      val asyncResetCore = Wire(Input(AsyncReset()))
-      systemReset := ResetCatchAndSync(systemClock, asyncResetCore.asBool)
-      IOCell.generateIOFromSignal(asyncResetCore, Some("iocell_reset"), abstractResetAsAsync = true)
+  val lSystem = LazyModule(p(BuildSystem)(p)).suggestName("system")
+
+  // The implicitClockSinkNode provides the implicit clock and reset for the System
+  val implicitClockSinkNode = ClockSinkNode(Seq(ClockSinkParameters()))
+
+  // Generate Clocks and Reset
+  p(ChipyardClockKey)(this)
```
colinschmidt, line 39: I kind of wish this was named something like `ChipyardClockingScheme`.

generators/chipyard/src/main/scala/ChipTop.scala
```diff
@@ -4,118 +4,64 @@ import chisel3._
...
+    // These become the implicit clock and reset to the System
+    val implicit_clock = implicitClockSinkNode.in.head._1.clock
+    val implicit_reset = implicitClockSinkNode.in.head._1.reset
+
+
+    // The implicit clock and reset for the system is also, by convention, used for all the IOBinders
+    // TODO: This may not be the right thing to do in all cases
+    withClockAndReset(implicit_clock, implicit_reset) {
+      val (_ports, _iocells, _harnessFunctions) = p(IOBinders).values.flatMap(f => f(lSystem) ++ f(lSystem.module)).unzip3
+      // We ignore _ports for now...
```
colinschmidt, line 55: I don't understand what this means? Are ports handled by some other infrastructure or is it something chipyard doesn't support at all?

generators/chipyard/src/main/scala/Clocks.scala
```diff
@@ -0,0 +1,176 @@
...
+      val div_clock = Pow2ClockDivider(clock_wire, 2)
+
+      implicitClockSourceNode.out.unzip._1.map { o =>
+        o.clock := div_clock
+        o.reset := reset_wire
+      }
+
+      simpleClockGroupSourceNode.map { n => n.out.unzip._1.map { out: ClockGroupBundle =>
+        out.member.elements.map { case (name, data) =>
+          // This is mega hacks, how are you actually supposed to do this?
```
colinschmidt, line 163: This can't be the way, it seems hackier than before there was diplomatic clocking.

generators/chipyard/src/main/scala/ConfigFragments.scala
```diff
@@ -105,7 +98,8 @@ class WithMultiRoCCHwacha(harts: Int*) extends Config(
     case MultiRoCCKey => {
       up(MultiRoCCKey, site) ++ harts.distinct.map{ i =>
         (i -> Seq((p: Parameters) => {
-          LazyModule(new Hwacha()(p)).suggestName("hwacha")
+          val hwacha = LazyModule(new Hwacha()(p)).suggestName("hwacha")
```
colinschmidt, line 101: What is the purpose of explicitly creating this val?

generators/chipyard/src/main/scala/TestHarness.scala
```diff
@@ -5,33 +5,41 @@ import chisel3._
...
+  val success: Bool
+}
+
+class TestHarness(implicit val p: Parameters) extends Module with HasHarnessUtils {
   val io = IO(new Bundle {
     val success = Output(Bool())
   })
 
-  val dut = p(BuildTop)(p)
+  val ldut = LazyModule(p(BuildTop)(p)).suggestName("chiptop")
```
colinschmidt, line 31: I believe this is a change in name of the dut module. If so we need to update it in the vlsi makefragments.

generators/chipyard/src/main/scala/TestHarness.scala
```diff
@@ -5,33 +5,41 @@ import chisel3._
...
+  val success = io.success
+
   // dutReset assignment can be overridden via a harnessFunction, but by default it is just reset
   val dutReset = WireDefault(if (p(GlobalResetSchemeKey).pinIsAsync) reset.asAsyncReset else reset)
 
-  dut.harnessFunctions.foreach(_(this))
-
-  def success = io.success
-  def harnessReset = this.reset.asBool
+  ldut.harnessFunctions.foreach(_(this))
```
colinschmidt, line 42: None of the IOBinders depended on being called before the success val is set right?

generators/chipyard/src/main/scala/config/AbstractConfig.scala
```diff
@@ -17,6 +17,7 @@ class AbstractConfig extends Config(
   new chipyard.config.WithBootROM ++                             // use default bootrom
   new chipyard.config.WithUART ++                                // add a UART
   new chipyard.config.WithL2TLBs(1024) ++                        // use L2 TLBs
+  new chipyard.config.WithNoSubsystemDrivenClocks ++             // drive the subsystem diplomatic clocks from ChipTop instead of using implicit clocks
```
colinschmidt, line 20: Glad we an abstract config 👍 
I would be tempted to make the default config the divided version, since that is more "realistic".

generators/chipyard/src/main/scala/config/RocketConfigs.scala
```diff
@@ -184,3 +184,18 @@ class MMIORocketConfig extends Config(
...
+// always get the TestHarness clock. The Tiles and Uncore receive the correct clocks
+class DividedClockRocketConfig extends Config(
+  new chipyard.config.WithTileDividedClock ++                       // Put the Tile on its own clock domain
+  new freechips.rocketchip.subsystem.WithRationalRocketTiles ++   // Add rational crossings between RocketTile and uncore
+  new freechips.rocketchip.subsystem.WithNBigCores(1) ++
+  new chipyard.config.AbstractConfig)
+
+
+class TestClockRocketConfig extends Config(
+  //new chipyard.config.WithTileMultiClock ++                       // Put the Tile on its own clock domain
```
colinschmidt, line 198: Can't you just use the same mixin as above? It should still work through the async crossing.

.circleci/do-rtl-build.sh
jerryz123: No. It only prints the passes as they go by, which does not swamp the console.

generators/chipyard/src/main/scala/ChipTop.scala
jerryz123: This is from John's original IOCells implementation. I interpret the comment to mean that the list of ports returned by the IOBinders evaluation isn't used in the current generator, although a future extension may choose to reference the list of generated ports.

generators/chipyard/src/main/scala/config/AbstractConfig.scala
jerryz123: That would be cool, if only it worked :cry:

generators/chipyard/src/main/scala/TestHarness.scala
jerryz123: I can leave it as `dut`, but I feel `chiptop` is a more accurate name

generators/chipyard/src/main/scala/TestHarness.scala
jerryz123: I don't think so? IOBinders can either set or read the success signal. Are you envisioning a case where the order of evaluation matters?

generators/chipyard/src/main/scala/config/RocketConfigs.scala
jerryz123: Yeah, you are right.

generators/chipyard/src/main/scala/TestHarness.scala
colinschmidt: I also like the new name and was mostly just alerting you that this is used in other infrastructure.

generators/chipyard/src/main/scala/Clocks.scala
jerryz123: :skull: 

At least everything is contained in one place now.

generators/chipyard/src/main/scala/config/AbstractConfig.scala
davidbiancolin: I think we should do that once it does work.

generators/firechip/src/test/scala/ScalaTestSuite.scala
```diff
@@ -106,8 +106,8 @@ class BoomF1Tests extends FireSimTestSuite("FireSim", "DDR3FRFCFSLLC4MB_FireSimL
 class RocketNICF1Tests extends FireSimTestSuite("FireSim", "WithNIC_DDR3FRFCFSLLC4MB_FireSimRocketConfig", "BaseF1Config")
 // Multiclock tests
 class RocketMulticlockF1Tests extends FireSimTestSuite(
-  "FireSimMulticlockPOC",
-  "FireSimQuadRocketMulticlockConfig",
+  "FireSim",
```
davidbiancolin, line 109: This is going to require changing the `ini`s.

generators/chipyard/src/main/scala/ConfigFragments.scala
jerryz123: Making it a `val` makes the `suggestName` redundant, but this was an artifact from testing. I'll remove this.

generators/chipyard/src/main/scala/ChipTop.scala
jerryz123: Changed to `ClockingSchemeKey`

generators/chipyard/src/main/scala/TestHarness.scala
jerryz123: I changed the Makefile in `vlsi`

generators/chipyard/src/main/scala/Clocks.scala
```diff
@@ -0,0 +1,176 @@
...
+}
+
+
+case object ChipyardClockKey extends Field[ChipTop => Unit](ClockDrivers.harnessClock)
+
+
+
+object ClockDrivers {
+  // A simple clock provider, for testing
+  val harnessClock: ChipTop => Unit = { chiptop =>
```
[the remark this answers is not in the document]
jerryz123: Changed to `ClockingSchemeGenerators`

generators/firechip/src/test/scala/ScalaTestSuite.scala
jerryz123: I will do this when I regenerate AGFIs

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -366,13 +368,14 @@ class WithTiedOffDebug extends OverrideIOBinder({
   (system: HasPeripheryDebugModuleImp) => {
     val (psdPort, resetctrlOpt, debugPortOpt, ioCells) =
       AddIOCells.debug(system.psd, system.resetctrl, system.debug)(system.p)
-    val harnessFn = (th: chipyard.TestHarness) => {
+    val harnessFn = (th: HasHarnessUtils) => {
       Debug.tieoffDebug(debugPortOpt, resetctrlOpt, Some(psdPort))(system.p)
       // tieoffDebug doesn't actually tie everything off :/
       debugPortOpt.foreach { d =>
-        d.clockeddmi.foreach({ cdmi => cdmi.dmi.req.bits := DontCare; cdmi.dmiClock := th.clock })
+        // TODO: Using harness clock/reset will be incorrect when systemClock =/= harnessClock
```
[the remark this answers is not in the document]
jerryz123: So it turns out the way we drive the debug module clocks is pretty wrong, debug.clock should not be punched off-chip to be driven.
We should fix the WithSimDebug and WithTiedOffDebug IOBinders in another PR.

jerryz123:
@davidbiancolin @colinschmidt 
I favor merging this PR as-is (after agfi regeneration), assuming no major changes are still necessary. There are still a number of problems to solve before multi-clock designs can be cleanly enabled. Those should be addressed in follow-up PRs, as they may require extensive changes through sub-projects, and additional rocketchip bumps.

davidbiancolin:
I'm fine with that -- but you should put a failing require or assert in the divided config.

generators/chipyard/src/main/scala/IOBinders.scala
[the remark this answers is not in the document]
davidbiancolin: Sounds good.

colinschmidt:
Yeah. Merging is fine as we have plenty of time before the next release. 
Is there a list of things that need to be fixed or is that an unknown unknown? If you know of a list or at least a partial list its probably worth filing a tracking issue.

generators/chipyard/src/main/scala/IOBinders.scala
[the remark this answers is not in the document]
colinschmidt: I think there is probably some confusion between the DMI_DTM signals going off chip (which is only intended as a simulation config) and the JTAG_DTM version which has a pretty clear set of signals that should go off-chip (and is intended to be actually implemented in the real world). Perhaps this confusion is carried through to the IOBinders?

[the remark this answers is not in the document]
jerryz123: Yeah that is accurate. I think the problem is that we are plumbing through the entire `DebugIO` bundle, instead of just `DebugIO.systemjtag` or `DebugIO.clockeddmi`. 

`DebugIO.clock` appears to require a clock which is synchronous to the Tilelink buses (https://github.com/chipsalliance/rocket-chip/blob/3176ed81c2bfa1a24622c511a1e4b0777bc9377c/src/main/scala/devices/debug/Debug.scala#L1775-L1780), so it should not be generated offchip. I feel like we should be calling `Debug.connectClockAndReset` somewhere within the chip...

Issue #616, opened 2020-06-29
Top AXI slave port


pbert519:
**Impact**: rtl

**Tell us about your environment:**
*Chipyard Version:* 
*OS:* 

I try to generate to generate verilog wit the sim/verilator makefile, which has the AXI Slave port available in the ChipTop module. The configuration is the following:
`   class RocketConfig extends Config(`
    `new chipyard.iobinders.WithUARTAdapter ++ `                     
    `new chipyard.iobinders.WithTieOffInterrupts ++   `              
    `new chipyard.iobinders.WithBlackBoxSimMem ++  `             
    `new chipyard.iobinders.WithTiedOffDebug ++     `                
    `new chipyard.iobinders.WithSimSerial ++            `           
    `new chipyard.iobinders.WithTieOffL2FBusAXI ++`
    `new testchipip.WithTSI ++                                      `
    `new chipyard.config.WithBootROM ++                    `     
    `new chipyard.config.WithUART ++                         `       
    `new chipyard.config.WithL2TLBs(1024) ++             `      
    `new freechips.rocketchip.subsystem.WithNoMMIOPort ++  `        
    `//new freechips.rocketchip.subsystem.WithNoSlavePort ++      `   
    `new freechips.rocketchip.subsystem.WithInclusiveCache ++     ` 
    `new freechips.rocketchip.subsystem.WithNExtTopInterrupts(0) ++ `
    `new freechips.rocketchip.subsystem.WithNBigCores(1) ++        ` 
    `new freechips.rocketchip.subsystem.WithCoherentBusTopology ++  `
    `new freechips.rocketchip.system.BaseConfig)`                   

**What is the current behavior?**
During '[info] running barstools.tapeout.transforms.GenerateTopAndHarness' i get the following errors:
'[error] (run-main-0) firrtl.passes.PassExceptions:' 
'[error] firrtl.passes.CheckInitialization$RefNotInitializedException:  @[ChipTop.scala 71:68:chipyard.TestHarness.RocketConfig.fir@278078.4] : [module ChipTop]  Reference system is not fully initialized.'
'[error]    : system.l2_frontend_bus_axi4.0.aw.bits.size <= VOID'
This is just an example, there is an error for every AXI Signal. I attached the full log here:
[log.txt](https://github.com/ucb-bar/chipyard/files/4846481/log.txt)

**What is the expected behavior?**
A generated verilog ChipTop module with the AXI Slave as IO.

**Other information**
[...]

PR #618, opened 2020-06-30
Fixes for AXI4 MMIO and FBus ports

jerryz123:
**Related issue**: #616 

**Type of change**: bug fix | new feature 

**Impact**: rtl change 

 * Fixes bug with AXI4 MMIO ports not being generated properly due to
   IOBinders issue. Additionally adds IOCells to AXI4 ports so that they
   appear in ChipTop
 * Change IOBinders to also require passing p: Parameters
   to child functions. Serialization of type targets via ClassTags fails
   for compound types, so we cannot use `BaseSubsystem with HasSomeTrait`
   as the type target in OverrideIOBinders.


generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -249,11 +258,11 @@ class WithBlockDeviceModel extends OverrideIOBinder({
...
-  (system: CanHavePeripheryIceNICModuleImp) => system.connectNicLoopback(); Nil
+  (system: CanHavePeripheryIceNICModuleImp, p) => system.connectNicLoopback(); Nil
 })
 
 class WithSimNIC extends OverrideIOBinder({
-  (system: CanHavePeripheryIceNICModuleImp) => system.connectSimNetwork(system.clock, system.reset.asBool); Nil
+  (system: CanHavePeripheryIceNICModuleImp, p) => system.connectSimNetwork(system.clock, system.reset.asBool); Nil
 })
 
 // Note: The parameters instance is accessible only through the BaseSubsystem
```
davidbiancolin, line 268: Delete this comment.

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -185,10 +185,19 @@ object AddIOCells {
...
-      val (port, ios) = IOCell.generateIOFromSignal(mem_axi4, Some(s"iocell_mem_axi4_${i}"))
-      port.suggestName(s"mem_axi4_${i}")
+      val (port, ios) = IOCell.generateIOFromSignal(mem_axi4, Some(s"iocell_${name}_axi4_slave_${i}"))
+      port.suggestName(s"${name}_axi4_slave_${i}")
+      (port, edge, ios)
+    }}
+  }
+  def axi4(io: Seq[AXI4Bundle], node: AXI4MasterNode, name: String): Seq[(AXI4Bundle, AXI4EdgeParameters, Seq[IOCell])] = {
+    io.zip(node.out).zipWithIndex.map{ case ((mem_axi4, (_, edge)), i) => {
+      //val (port, ios) = IOCell.generateIOFromSignal(mem_axi4, Some(s"iocell_${name}_axi4_master_${i}"))
```
jwright6323, line 197: I would like this to be fixed, but I also understand that I'm in the critical path for that :). I'll take a look, but I won't block this PR because of this.

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -262,16 +271,16 @@ class WithSimNIC extends OverrideIOBinder({
...
-  (system: CanHaveMasterAXI4MemPort with BaseSubsystem) => {
-    val peiTuples = AddIOCells.axi4(system.mem_axi4, system.memAXI4Node)
+  (system: CanHaveMasterAXI4MemPort, p) => {
+    val peiTuples = AddIOCells.axi4(system.mem_axi4, system.memAXI4Node, "mem")
     // TODO: we are inlining the connectMem method of SimAXIMem because
     //   it takes in a dut rather than seq of axi4 ports
     val harnessFn = (th: chipyard.TestHarness) => {
       peiTuples.map { case (port, edge, ios) =>
-        val mem = LazyModule(new SimAXIMem(edge, size = system.p(ExtMem).get.master.size)(system.p))
+        val mem = LazyModule(new SimAXIMem(edge, size = p(ExtMem).get.master.size)(p))
```
jwright6323, line 280: One other thought- would some kind of match work?

```suggestion
        val p = system match {
          case BaseSubsystem => system.p
          case _ => assert(false)
        }
        val mem = LazyModule(new SimAXIMem(edge, size = p(ExtMem).get.master.size)(p))
```

generators/chipyard/src/main/scala/IOBinders.scala
jerryz123: Moved the comment to where we match for `LazyModule` or `LazyModuleImpLike`

generators/chipyard/src/main/scala/IOBinders.scala
jerryz123: The new version does this basically.

jwright6323: I was still thinking you'd call the object function here, rather than pass `p` as an argument

jerryz123: Ok, I can do it that way.

PR #625, opened 2020-07-08
Override default baud rate for FireChip

a0u:
**Related issue**: 

**Impact**: rtl change

**Release Notes**

This avoids target software needing to explicitly set the divisor to match the [UART bridge](https://github.com/firesim/firesim/blob/f5374ffcdfdac657245274670ba18577fb8561c9/sim/firesim-lib/src/main/scala/bridges/UARTBridge.scala#L48).

### DIFF
--- a/generators/firechip/src/main/scala/TargetConfigs.scala
+++ b/generators/firechip/src/main/scala/TargetConfigs.scala
@@ -69,7 +69,8 @@ class WithNVDLASmall extends nvidia.blocks.dla.WithNVDLA("small")
 class WithFireSimConfigTweaks extends Config(
   // Required*: When using FireSim-as-top to provide a correct path to the target bootrom source
   new WithBootROM ++
-  // Optional*: Removing this will require target-software changes to properly capture UART output
+  // Optional*: Removing this will require adjusting the UART baud rate and
+  // potential target-software changes to properly capture UART output
   new WithPeripheryBusFrequency(BigInt(3200000000L)) ++
   // Required: Existing FAME-1 transform cannot handle black-box clock gates
   new WithoutClockGating ++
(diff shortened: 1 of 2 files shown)


generators/firechip/src/main/scala/TargetConfigs.scala
```diff
@@ -69,7 +69,8 @@ class WithNVDLASmall extends nvidia.blocks.dla.WithNVDLA("small")
 class WithFireSimConfigTweaks extends Config(
   // Required*: When using FireSim-as-top to provide a correct path to the target bootrom source
   new WithBootROM ++
-  // Optional*: Removing this will require target-software changes to properly capture UART output
+  // Optional*: Removing this will require adjusting the UART baud rate and
+  // potential target-software changes to properly capture UART output
```
timsnyder, line 73: @a0u Any tips on the "potential target-software changes"?  In the past [Sagar pointed to adjusting the UART div in riscv-pk](https://groups.google.com/forum/#!msg/firesim/3-w8vt95OTU/ewEteglcDAAJ).  If I'm using the br-base-bin linux image from Firemarshal, would I need to adjust the kconfig somewhere adjusting riscv-pk/machine/uart.c like Sagar mentioned?

a0u: As of firesim/firesim#598, both the UART device and bridge are initialized with the same baud rate from `UARTParams`, so target software no longer needs to adjust the divisor after reset.  This has recently enabled FireMarshal to switch back to upstream riscv-pk.  At this point, a user is largely free to change the `initBaudRate` parameter and pbus frequency without worrying about software, but the cautionary note still applies to at least two situations:

* As a fundamental constraint, the baud rate cannot exceed the pbus frequency (divided by 2).
* The _Linux_ UART driver normally defaults to 115200 baud, unless specified otherwise by the boot command line or `/etc/inittab` (`getty`).  Most FireSim users do not need to be concerned about this since FireMarshal kernels use the SBI console driver, which is agnostic to the baud rate and bus frequency.

timsnyder: Thanks for your reply @a0u.  This information is very helpful.

PR #629, opened 2020-07-17
Add RANDOM_SEED variable to set random init for VCS and Verilator simulations

jerryz123:
Setting `RANDOM_SEED` now lets you control the random seed for VCS and Verilator. Leaving `RANDOM_SEED` unset results in randomly generated seed.

**Related issue**: 

**Type of change**:  new feature

**Impact**: software change

**Release Notes**

### DIFF
--- a/sims/vcs/Makefile
+++ b/sims/vcs/Makefile
@@ -47,13 +47,11 @@ VCS_OPTS = -notice -line $(VCS_CC_OPTS) $(VCS_NONCC_OPTS) $(VCS_DEFINE_OPTS) $(E
 # vcs simulator rules
 #########################################################################################
 $(sim): $(sim_vsrcs) $(sim_common_files) $(dramsim_lib) $(EXTRA_SIM_REQS)
-	rm -rf csrc && $(VCS) $(VCS_OPTS) -o $@ \
-	-debug_pp
+	rm -rf csrc && $(VCS) $(VCS_OPTS) -o $@
 
 $(sim_debug): $(sim_vsrcs) $(sim_common_files) $(dramsim_lib) $(EXTRA_SIM_REQS)
 	rm -rf csrc && $(VCS) $(VCS_OPTS) -o $@ \
-	+define+DEBUG \
-	-debug_pp
+	+define+DEBUG
 
 #########################################################################################
 # create a vcs vpd rule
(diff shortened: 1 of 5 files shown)


sims/vcs/Makefile
```diff
@@ -47,13 +47,11 @@ VCS_OPTS = -notice -line $(VCS_CC_OPTS) $(VCS_NONCC_OPTS) $(VCS_DEFINE_OPTS) $(E
 # vcs simulator rules
 #########################################################################################
 $(sim): $(sim_vsrcs) $(sim_common_files) $(dramsim_lib) $(EXTRA_SIM_REQS)
-	rm -rf csrc && $(VCS) $(VCS_OPTS) -o $@ \
-	-debug_pp
```
jerryz123, line 51: `-debug_pp` is already defined in `VCS_OPTS`

sims/verilator/Makefile
```diff
@@ -29,6 +29,7 @@ sim = $(sim_dir)/$(sim_prefix)-$(MODEL_PACKAGE)-$(CONFIG)
 sim_debug = $(sim_dir)/$(sim_prefix)-$(MODEL_PACKAGE)-$(CONFIG)-debug
 
 WAVEFORM_FLAG=-v$(sim_out_name).vcd
+SEED_FLAG=+verilator+seed+I$(RANDOM_SEED)
```
colinschmidt, line 32: Good thing we added support for these flags 😝

variables.mk
```diff
@@ -167,6 +167,7 @@ PERMISSIVE_OFF=+permissive-off
 BINARY ?=
 override SIM_FLAGS += +dramsim +dramsim_ini_dir=$(TESTCHIP_DIR)/src/main/resources/dramsim2_ini +max-cycles=$(timeout_cycles)
 VERBOSE_FLAGS ?= +verbose
+RANDOM_SEED ?= 1
```
colinschmidt, line 170: I think we should have this actually be a random number most of the time.
And when the user suspects nondeterminism they would fix the seed to a particular value.

This is assuming the current behavior has a random seed.

jerryz123: The current behavior in VCS uses 1 as the seed for every run. I think having default be deterministic makes more sense, otherwise it will always be annoying to reproduce some non-determinism-induced bug.

Users looking for randomness can just set `RANDOM_SEED=$RANDOM`

davidbiancolin:
I'm with Colin on this. We should just print the seed if we don't already.

jerryz123:
The seed is already printed in the `TestDriver.v`, and can be seen in the `.out/.log` file. 
The new default behavior is random init in both verilator and VCS, with the option to set a seed with the `RANDOM_SEED` variable.

PR #615, opened 2020-06-29
Documentation for Third-Party Core Integration

zitaofang:
This is the documentation for core integration under the new Rocket API introduced in June. 

**Related issue**:  Not Applicable

**Type of change**: other enhancement

**Impact**: other

**Release Notes**

- Add instructions for integrating third-party cores with Chipyard

PR #632, opened 2020-07-20
Documentation for Available Config Keys

zitaofang:
This PR intends to address the comment on the appendix list of available keys in PR #615. I am currently listing all config keys presented in Chipyard repo, but not all of them are useful to users that want to integrate their cores. Input needed. 

**Related issue**: N/A

**Impact**: other

**Release Notes**

- Add documentation for available config keys


jerryz123:
It seems like documentation on keys can be better displayed as inline comments. It may be better to open a PR against Rocketchip with concise inline comments for some of their keys.

jerryz123:
Additionally, many of the keys listed here are not "free variables", so-to-speak. There are many implicit, hidden dependencies between these keys, and we don't want to suggest to users that these keys are all independent variables.

Documentation should focus on the user-facing interface for customization (ConfigFragments), not the underlying implementation.

alonamid:
This has long been a contentious discussion issue in Chipyard. The main concern is stale documentation.
However, in contrast the Jerry, I do believe there is value in a *centralized* description of config keys. The problem is how to not make it grow stale. One option could be a script which tries to search for config keys which match the naming convention, and point to their locations in the codebase (which could then reference the comments Jerry is talking about). I'm open to other options.

colinschmidt:
Perhaps we should have a developers only doc section with a huge flashing warning at the top saying "Any and all of this information may be out of date, wrong, or bad for your health. Proceed at your own risk.".
Then we can put things like random case classes there.

I think most of the documentation should be focused on what users can do. If there are mixins that we are missing documents for we should add that. More on topic with the PR that spawned this one, we should add some documentation on how to integrate an external core including some subset of keys that should be modified but providing an exhaustive list seems prohibitive and problematic.

zitaofang:
Sorry for the late response since I was working on another task. It is a challenging task to write a script to detect config keys and generate documentation for them, and we will still need to manually write descriptions for the keys since many existing config keys don't have Scaladoc comments. In addition, most of the features documented here are only for advanced users who will normally read the code themselves, and the majority of users only want to integrate their own cores (which can be done without touching those keys), the best way to deal with this IMO is to have (the small group of) advanced users contact us if they have any questions on the keys. 

Therefore, I am closing this PR without merging.

PR #636, opened 2020-07-31
updated openroad hash

ssteffl:
updated vlsi/hammer hash with openroad updates


colinschmidt:
In the future we should try to roll this into the previous PR.

PR #635, opened 2020-07-30
Implement fast loadmem feature

zhemao:
**Related issue**: https://github.com/ucb-bar/testchipip/pull/94

**Type of change**: new feature

**Impact**: software change

**Release Notes**

Implement fast loadmem feature for VCS/Verilator simulation

PR #639, opened 2020-08-03
Esp tsi loadmem

colinschmidt:
This build on #635 to add a `run-binary*-hex` target that creates a hex file from the executable and loads it for simulation.

This speeds up simulations for large executables considerably. 
Requires https://github.com/ucb-bar/esp-isa-sim/pull/16

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -161,9 +161,7 @@ run-fast: run-asm-tests-fast run-bmark-tests-fast
 # helper rules to run simulator with fast loadmem via hex files
 #########################################################################################
 $(binary_hex): $(output_dir) $(BINARY)
-	hex_bytes=$(shell readelf --segments --wide $(BINARY) | grep LOAD | tail -n 1 | tr -s [:space:] | cut -f4,6 -d' ' | tr -d x | tr [:lower:] [:upper:] | tr ' ' + | sed 's/0000000008/ibase=16;/' | bc);\
-	power_2_bytes=`python -c "import math; print int(pow(2,math.ceil(math.log($$hex_bytes)/math.log(2))))"`;\
-	elf2hex 64 $$power_2_bytes $(BINARY) $(shell echo "ibase=16;$(LOADMEM_ADDR)" | bc) > $(binary_hex)
+	$(base_dir)/scripts/smartelf2hex.sh $(BINARY) > $(binary_hex)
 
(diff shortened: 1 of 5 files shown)


jerryz123:
Do the new make rules only work when the user builds the esp-toolchain?

colinschmidt:
If you don't have the new esp-fesvr from this PR then TSI will be unable to skip the loading. The simulation will still run correctly it just won't be sped up.
I think the change is upstream-able but we should discuss in the chipyard meeting.

zhemao:
I've merged my PR, so you can go ahead and rebase now.

zhemao:
Also, you may want to check that this actually works in verilator. The main function in emulator.cc does some complicated transformations on the arguments that might end up dropping +loadmem and +loadmem_addr.

colinschmidt:
I believe we no longer need that mangling in emulator. Since my `+permissive` handling changes a while back.
You were right that this loadmem change wouldn't have worked in verilator until this fix.

common.mk
```diff
@@ -157,6 +157,30 @@ run-binary-debug: $(output_dir) $(sim_debug)
 
 run-fast: run-asm-tests-fast run-bmark-tests-fast
 
+#########################################################################################
+# helper rules to run simulator with fast loadmem via hex files
+#########################################################################################
+$(binary_hex): $(output_dir) $(BINARY)
+	hex_bytes=$(shell readelf --segments --wide $(BINARY) | grep LOAD | tail -n 1 | tr -s [:space:] | cut -f4,6 -d' ' | tr -d x | tr [:lower:] [:upper:] | tr ' ' + | sed 's/0000000008/ibase=16;/' | bc);\
```
zhemao, line 164: This is a pretty complicated series of commands. Since you're invoking python anyway, maybe it would be better to write a standalone python script and then call it here. There are some libraries you can use to read the ELF. If that's too much of a pain, you can just make it a bash script with the same commands.

colinschmidt: Do we have a place in chipyard to store python scripts?

jerryz123: `./scripts`

colinschmidt:
Ok. I've fixed the feedback, and added a doc snippet.

colinschmidt:
Someone else's attempt to make elf2hex better: https://github.com/sifive/elf2hex

Issue #109, opened 2019-06-01
Docker images, installation instructions


jerryz123:
Several retreat attendees mentioned Docker images as a distribution format. I think this is a good idea.

At the very least, we should make the init scripts install all required packages that are not in a base Ubuntu install.

Currently, the CI runs off a manually uploaded docker image. We should rebuild this docker image in the CI, and deploy it as well.

- [ ] Move Dockerfile (used in CI) from `riscv-boom` to `chipyard`
- [x] Add script(s) to install required packages in base OS(s)
- [ ] Rebuild Docker container and deploy in CI

faddat:
I am working on this currently, my first attempt was with Arch Linux and I don't know if it's going to prove to be successful. If it is, I will send a pull request your way.

If it is not, I will do it with Ubuntu and also send a pull request your way.

I have been researching open CPU cores for about 5 months now and the first two maybe three months was spent getting the environment right, then learning Well, almost everything.

I started this as a very advanced Linux user without much knowledge of semiconductor design and engineering.  Actually, I still probably don't know that much about semiconductor design and engineering but I think that making systems like chipyard easier to use from the get-go will make these kinds of engineering tasks more approachable for more people and lead to a larger ecosystem of open source hardware.

faddat:
https://gitlab.com/virgohardware/core/-/blob/master/chipyard/Dockerfile

faddat:
It seems to be working steadily now though I do have some concerns about dependencies

faddat:
The current Dockerfile isn't appropriate for people who want to **run** shipyard.  

While I'll be using it in CI, I can't actually build an image from that Dockerfile, I get some version conflicts in the python packages.  

So I'm reworking it from bare Debian Sid.  

I've updated the Dockerfile referenced in #641

faddat:
In the discussion on #641, @jerryz123 and I talked about the proper form of a Chipyard Docker image.  

After several days of fiddling, I think my thoughts have changed.  Arch Linux has high quality precompiled binaries for the riscv toolchain, and they're kept up to date.  With https://gitlab.com/virgohardware/core I'm trying to make it as easy as possible to:

* modify / create an SOC
* build the SOC
* test the SOC
* compile Linux for the SOC
* run linux on it in simulation
* run linux on it using a cloud FPGA, probably using fire.sim
* download FPGA bitstreams for user-owned FPGAs
* prep the core for manufacturing and download the needed binaries for that

[...]

PR #641, opened 2020-08-09
Add Dockerfile

faddat:
Added Dockerfile. It's possible that it is missing some dependencies, or should simply be switched to a Ubuntu base.  Curious to learn your review and thoughts. The  Dockerfile is in use in this CI  system: https://gitlab.com/virgohardware/core/-/pipelines

Creating it, I opted to use as many precompiled binaries as possible to reduce the image build time.  It's also possible to restructure the setup scripts in a docker format, if that will work out better.  

**Related issue**: 

#109 

**Type of change**: bug fix | new feature | other enhancement

Add a Dockerfile for easier setup and orientation.  

**Impact**: rtl change | software change | unknown | other

Other

**Release Notes**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,39 +1,244 @@
-FROM archlinux
+### Note: This DockerFile is adapted from https://github.com/ucb-bar/chipyard/blob/master/.circleci/images/Dockerfile which was adapted from: https://github.com/CircleCI-Public/example-images/openjdk
+# This is a full chipyard setup, which will be built manually on-demand in the Tendermint Hardware Project at https://gitlab.com/virgohardware/core/
 
-# This Dockerfile was created for the Tendermint Hardware project and funded by Tendermint, Inc.
-# It is focused on speed and ease of use, so it uses pre-compiled tools availalbe in Arch Linux.
-# This Dockerfile can be used in CI systems like the one found here: https://gitlab.com/virgohardware/core
-# Individual developers can also use this Dockerfile to experiment with Chipyard without having to worry about configuring a development environment.
+FROM debian:sid
+
 MAINTAINER jacobgadikian@gmail.com
 
-# Every risc-v package in community
-RUN pacman -Syyu --noconfirm python-pythondata-cpu-vexriscv riscv32-elf-binutils riscv32-elf-gdb riscv32-elf-newlib riscv64-elf-binutils riscv64-elf-gcc riscv64-elf-gdb riscv64-elf-newlib riscv64-linux-gnu-binutils riscv64-linux-gnu-gcc riscv64-linux-gnu-gdb riscv64-linux-gnu-glibc riscv64-linux-gnu-linux-api-headers
+# man directory is missing in some base images
+# https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=863199
+RUN apt-get update && \
+    apt-get upgrade -y && \
+    mkdir -p /usr/share/man/man1 && \
+    apt-get install -y \
+               bzip2 \
+               ca-certificates \
+               curl \
+               git \
+               gnupg \
+               gzip \
+               libfl2 \
+               libfl-dev \
+               locales \
+               mercurial \
+               python2 \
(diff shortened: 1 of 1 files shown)


jerryz123:
Thanks for looking at this. I have a few comments..

- Chipyard is not guaranteed to be compatible with packaged versions of various risc-v tools (for example, FESVR). For various reasons, we find it more flexible to build and deploy specific versions of these tools tracked in Chipyard. 
- We don't encourage setting the `RISCV` path to `/usr/bin`. The chipyard `build-toolchains.sh` script should generate a tools-install directory and a bash script to set `RISCV`.

We already have a Dockerfile for CI that installs all prerequisites of Chipyard (without installing Chipyard itself). This is in `.circleci/images`. I think a "deploy" image should inherit from that Dockerfile.

faddat:
Thanks very much for your comments :).  

I'll give the same task a try with your recommendations applied, including inheriting from the Dockerfile in `.circleci/images`.  

Should I close this PR and submit another later, or just update my Dockerfile?

Cheers!

faddat:
Have some more comments, but will move them to #109

faddat:
I've updated this so that:

* It gets maven, gradle, ant, jq and sbt from apt
* It is no longer configured for docker in docker
* It allows the apt versions in debian Sid to be standard
* Build steps that have been omitted have been commented out for visibility

I will use it to build all of the cores that we've got in our CI system-- every "config" found in this repo, without exception.  

What else should I be testing?

I reckon that this will work, but of course we want a bit more than just a reckoning :).

Dockerfile
```diff
@@ -0,0 +1,244 @@
...
+#    && autoconf && ./configure && make && make install
+
+
+# Add HOME environment variable
+ENV HOME="/home/riscvuser"
+
+# Update PATH for RISCV toolchain (note: hardcoded for CircleCI)
+ENV RISCV="$HOME/riscv-tools-install"
+ENV LD_LIBRARY_PATH="$RISCV/lib"
+ENV PATH="$RISCV/bin:$PATH"
```
jerryz123, line 210: Do these still need to be set here?

faddat: Since we are not currently sourcing env.sh, yes.  They ensure that when the Docker image is used, environment variables are set even without sourcing env.sh.  

I should probably fire up the image to make sure that CI defaults to using user riscvuser, and it may be preferable to just source env.sh when using this in CI.  

For users who aren't using CI, these env vars ensure that they don't need to source env.sh when using the image.  

**goal:** This Docker image should allow anyone to play with Chipyard quickly and easily, using the exact same image that we use in the Tendermint Hardware [CI system](https://gitlab.com/virgohardware/core/-/pipelines) so that:

* questions about differing build environments are eliminated
* time spent configuring a build environment is drastically reduced
* need to own/rent a build machine is eliminated

all of those were a big problems for me when I first began researching Open Source Semiconductors.  

* users should be able to modify a core, create a branch in our repo, and build/test it so that the pace of development, especially for non-professionals, increases

jerryz123: I agree with the goals you specified.

I believe that setting `RISCV` here in the Dockerfile does not actually adjust the location of the eventual riscv-tools install, as the `build-toolchains.sh` hard-code a path (`{riscv/esp}-tools-install`) unless the `--prefix` argument is given. You will probably observe this when you inspect the final image.

Since Chipyard provides two toolchain options for users, I think its best to leave RISCV unset in final Dockerimage. The Chipyard makefiles will already provide a informative error if running RISCV unset, and the documentation additionally stresses that RISCV must be set by sourcing one of the `env` scripts.

faddat: Thanks for your advice!

I'll take that out, and see what I find :).

faddat:
This is currently pretty much broken; I only thought it was working :/.  

Will continue to bang on it until it's working.

abejgonzalez:
Im going to close this in favor of #771  since that PR seems to work nicely and achieves the same result.

PR #645, opened 2020-08-13
Update ubuntu-req.sh

faddat:
Use more cores when we can, and fewer when we can't use more

**Impact**: software


alonamid:
Thank you for your contribution!

PR #647, opened 2020-08-15
Fix Verilator Simulation run-binary-debug Error 

zitaofang:
The change to the line https://github.com/ucb-bar/chipyard/blob/dev/variables.mk#L175 breaks run-binary-debug (and other debug target) in Verilator as there is no whitespace between `-v` option and the `.vcd` output path following it now. This PR addresses the issue.  

**Related issue**: N/A

**Type of change**: bug fix

**Impact**: other

**Release Notes**


jerryz123:
I'm not sure what is broken here is the flag. What version of verilator are you using?

colinschmidt:
I'm 95% positive there should be no space there.

zitaofang:
I am using the latest version of Verilator (I just pulled from their stable branch and rebuild it before I tested it on my machine). Here is the output if I don't have the space: https://gist.github.com/zitaofang/50e3c509b3a747112f599690309337ed

zitaofang:
If you look at the gist above, you will see that right after the `+verbose` flag, the `-v` flag sticks to the path right without whitespace, which cause the error. Adding a space there fixed the problem.

jerryz123:
> If you look at the gist above, you will see that right after the `+verbose` flag, the `-v` flag sticks to the path right without whitespace, which cause the error. Adding a space there fixed the problem.

Do you have evidence the whitespace caused the error? I'm also 95% sure the whitespace is not supposed to be there.

zitaofang:
You may reproduce this problem by running `make run-binary-debug BINARY=$RISCV/riscv64-unknown-elf/share/riscv-tests/isa/rv32mi-p-ma_addr` (or any test program) under the latest dev. I tested this in a clean Chipyard repo (with toolchain built in `dev` immediately after cloning the repo and switching to that branch).

zitaofang:
I should verify this issue on other environments before opening a PR (my bad), and since I do not have access to other Linux machine, I would open a new issue about this later and close this PR.

jerryz123:
I'm seeing really strange behavior here. Can someone else confirm this? 

- ```./simulator-chipyard-RocketConfig-debug +permissive -vabcdef +permissive-off test.riscv```: SUCCEEDS, with waves in ./abcdef
- ```./simulator-chipyard-RocketConfig-debug +permissive -vabcdefh +permissive-off test.riscv```: FAILS with htif parsing error
- ```./simulator-chipyard-RocketConfig-debug +permissive -v abcdef +permissive-off test.riscv```: SUCCEEDS, with waves in ./abcdef
- ```./simulator-chipyard-RocketConfig-debug +permissive -v abcdefh +permissive-off test.riscv```: SUCCEEDS with waves in ./abcdefh

My conclusion is that without a space after the `-v`, HTIF cannot properly parse the args when the desired output vpd has the letter `h` in its string (probably getting mistaken for the `h` command to ask for HTIF help text).

Sorry, @zitaofang, for our skepticism earlier. You may have actually found a bizarre bug in our system.

zitaofang:
The `tsi` is now instantiated in https://github.com/ucb-bar/testchipip/blob/master/src/main/resources/testchipip/csrc/SimSerial.cc, and the argument list is retrieved from the VPI. The previous version of `emulator.cc` passed the untruncated argument list into the verilator environment, which contains the `-v` flags. The argument parser in `htif_t` will interpret the entire `-v<filename>` string as a list of singe character flags and show the help list if there is "h" in the file name, causing the program to exit.

generators/utilities/src/main/resources/csrc/emulator.cc
```diff
@@ -163,6 +163,7 @@ int main(int argc, char** argv)
 #if VM_TRACE
       case 'v': {
         vcdfile_name = optarg;
+        // printf("%s\n", vcdfile_name);
```
jerryz123, line 166: Remove this please.

generators/utilities/src/main/resources/csrc/emulator.cc
```diff
@@ -258,14 +258,19 @@ int main(int argc, char** argv)
     return 1;
   }
 
+  int htif_argc = 1 + argc - optind;
+  char** htif_argv = (char **) malloc((htif_argc) * sizeof (char *));
+  htif_argv[0] = argv[0];
+  for (int i = 1; optind < argc;) htif_argv[i++] = argv[optind++];
```
colinschmidt, line 264: I think you need to do something more clever here to have this still work with permissive args see #563 
I could be wrong but you should test that you can still pass permissive args in.

colinschmidt (changes requested):
I want to make sure this doesn't regress on the permissive args.

generators/utilities/src/main/resources/csrc/emulator.cc
zitaofang: Sure. I copied this code from `emulator.cc` before the multithreading commit, and it should work if the code worked before. I will check it anyway though to make sure.

zitaofang:
@colinschmidt I verified the three commands in that issue and most of them behave the same before and after the change in this PR, except for this command:

```
./simulator-chipyard-RocketConfig +max-cycles=50000 +verbose -- +permissive +verilator+rand+reset+0 +permissive-off output/chipyard.TestHarness.RocketConfig/rv64ui-p-addi
```

The current code in `dev` does not recognize `+max-cycles`, but the version in this PR can. Here is the output of the `dev` version:

```
using random seed 1601526642
This emulator compiled with JTAG Remote Bitbang client. To enable, use +jtag_rbb_enable=1.
Listening on port 34561
[UART] UART0 is here (stdin/stdout).
terminate called after throwing an instance of 'std::runtime_error'
  what():  could not open +max-cycles=50000 (did you misspell it? If VCS, did you forget +permissive/+permissive-off?)
```

The PR version runs to the end.

colinschmidt:
Ok thanks for double checking.

colinschmidt (approved):

zitaofang:
Since using the original code will break simulations that accept Verilog plusargs, I am filtering all arguments beginning with a '-' from the argument list passed into the Verilator.

colinschmidt (changes requested):
I believe this breaks the permissive arguments that verilator supports. i.e. passing `+verilator+rand+reset` between `+permissive` and `+permissive-off`
I think the most robust solution is to specifically filter out arguments handled in the parsing code in emulator.cc.

zitaofang:
I did not notice that there is a list of plusargs in `emulator.cc` so I simply filter for arguments that begin with '+'. I will change it to make it copy HTIF and Verilog plusargs only.

zitaofang:
For some reasons the CI didn't work in the past few days, but it is running now.

jerryz123:
@colinschmidt Does this look ok now? Can we merge?

generators/utilities/src/main/resources/csrc/emulator.cc
```diff
@@ -235,6 +246,7 @@ int main(int argc, char** argv)
             c = '?';
           } else {
             c = 'p';
```
colinschmidt, line 248: This should be a capital `P`.
We have found a verilog plusarg not the argument `+permissive` which is what lowercase `p` means.

colinschmidt: For my sanity I tried to derive a case where this would manifest but,
This bug appears to be undetectable because you only arrive to this case if `opterr` is already 0 so the side effect of the lowercase `p` is already in effect.
However the intent is clearly that we should be setting this to capital `P`

colinschmidt (changes requested):
I think there is one very subtle bug remaining, another lurking in the code from what I assume was the time I added support for `+permissive`, and one nit fix from the same time period.

Let me know if you have a question about what my comments mean. Hopefully this careful review I did will be the last time we need to read/write this code for a while.

colinschmidt:
Sorry for the delay @zitaofang and @jerryz123 . After looking at the changes the other day I realized I needed to do a full read through and it took a while for me to set aside that time.

generators/utilities/src/main/resources/csrc/emulator.cc
```diff
@@ -213,6 +223,7 @@ int main(int argc, char** argv)
             verilog_plusargs_legal = 0;
```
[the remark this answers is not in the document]
zitaofang: Actually the code here will not distinguish whether the removed arguments are outside the permissive section or not. It will simply copy all arguments defined in the plug-args list generated from Verilog or the HTIF argument lists, as well as the arguments `+permissive` and `+permissive-off`. The only two consumers of arguments passed into the Verilator environment now are HTIF parser and Verilog plus-args, and we will need to modify `emulator.cc` anyway if we add another parser to make sure that they are not considered invalid arguments.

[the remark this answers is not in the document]
zitaofang: Never mind, I will need to look at this field more closely.

generators/utilities/src/main/resources/csrc/emulator.cc
zitaofang: Fixed.

generators/utilities/src/main/resources/csrc/emulator.cc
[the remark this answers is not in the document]
zitaofang: I moved `verilated_argv` to the 'P' clause of the `switch` code block. If we have a plusargs, it will only leaves the last two if-else clauses by setting `c` to either 'P' (when `emulator.cc` does not detect any issue) or '?' (when `emulator.cc` will report an illegal argument and terminate the simulation). This ensure that all arguments not recognized in `emulator.cc` will be added to `verilated_argv` unless it is an illegal argument, in which case the simulation terminated directly. 
The remaining plusargs defined in `emulator.cc` are not used anywhere else except for `+permissive` and `+permissive-off` which is used by HTIF parser, therefore I must put those two into `verilated_argv`.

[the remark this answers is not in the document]
zitaofang: I just found that we can still leave those two if-else clauses by `goto done_processing`, and I am looking at this right now.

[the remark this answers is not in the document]
zitaofang: Fixed. Hitting HTIF arguments will make the program exit the while loop, and I have code after the while loop that copies all the remaining arguments (which can only be HTIF) and the binary name onto the verilator argument stack.

colinschmidt:
Ok I think this looks good now. Hopefully we can avoid touching this code for a long time now.

colinschmidt (approved):

jerryz123 (approved):

PR #650, opened 2020-08-18
Make help target and VCS/Verilator Cleanup

abejgonzalez:
**Related issue**: #562

**Type of change**: other enhancement

**Impact**: other

**Release Notes**
This is a split of PR #562 with Makefile change minus the Multi-threading support. It includes changes to VCS so that it aligns with the same changes done to Verilator.

PR #562, opened 2020-05-22
Multithread verilator

ssteffl:
**Related issue**: 

**Type of change**: new features

**Impact**: simulation infrastructure

**Release Notes**

Enabled verilator multithreading, profiling, fst waveforms
Added `bin/numa_prefix` wrapper for multi-socket multithreaded verilator
Enabled selective hardware printfs in verilator.
Added `scripts/build-isa-sim.sh`

PR #654, opened 2020-08-21
Multithreaded Verilator

abejgonzalez:
**Related issue**: Split up from PR #562 

**Impact**: other

**Release Notes**
This splits up #562 to just include the multi-threaded Verilator improvements. The main differences from the prior PR is that `numa_prefix` is moved to `scripts/`, `numa_prefix` is directly pointed to in `common.mk`, and `NUMA_PREFIX` is added in more places (only where Verilator might be called).

Finally, I caught extra flags being passed to `LDFLAGS` that were unneeded and used the `TRACE_OPTS` that was missed in PR #650 .

### DIFF
--- a/scripts/numa_prefix
+++ b/scripts/numa_prefix
@@ -1,4 +1,4 @@
-#!/usr/bin/env perl
+#!/usr/bin/env python
 
 #============================================================================
 # - really simple script, which just prints out the numactl cmd to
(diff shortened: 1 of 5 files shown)


abejgonzalez:
This is untested. Additionally, someone may want to double-check the `emulator.cc` changes (it looks like the functionality changed from it was last PR'ed)

scripts/numa_prefix
```diff
@@ -23,45 +23,54 @@
...
-    }
-    print("numactl -m $best_node_id -C $best_cpus --");
-  } else {
-    die("malformed head line: $head_line\n");
-  }
-}
+        # loop through available nodes, selecting the node with the most free mem
+        for i in avail_nodes:
+            cpu_line      = lines[line_idx]
+            mem_size_line = lines[line_idx + 1]
```
colinschmidt, line 52: This isn't used anywhere. I think we should just remove it and comment on mem_free_line that we skip the size line.

scripts/numa_prefix
```diff
@@ -0,0 +1,76 @@
...
+    if node_match:
+        avail_nodes = node_match.group(1)
+        best_node_id = ""
+        best_cpus = ""
+        best_free_size = 0
+
+        # loop through available nodes, selecting the node with the most free mem
+        for i in avail_nodes:
+            cpu_line      = lines[line_idx]
+            mem_size_line = lines[line_idx + 1]
```
abejgonzalez, line 52:
```suggestion
            # mem. size unused. skip and use mem. free
```

PR #659, opened 2020-08-24
WIP - Simple divider-only PLL generation flow

davidbiancolin:
This is an implementation sketch of one means we could instantiate a PLL in ChipTop. 

It functions by collecting the async clock group and generating frequencies based on the `take` parameter in ClockSinkParameters.

Notes:
- Clocks that do not provide a take parameter are assumed to be driven by a clock that shares the same name parameter in the group. `ClockGroupDealiaser` removes extra clocks under that assumption.
- This relies on injecting a clock node into the tile to set the take parameter. This is almost certainly not the right way to do this, but it works without modifying RC source or without overriding RC-defined methods. Overriding `connectPRC` seems like a reasonable solution though.
- The implicit clock is driven with the sbus clock mostly by coincidence (?) right now (since they both specify no name, and because of the way the ClockGroupAggregator combines group names (it doesn't)). 

How we want to get derived clocks to IOBinders remains a problem. 

TODOs:
- [ ] Harness side clock generation should be driven by the choose PLL reference frequency.
- [ ] Reset handling?


generators/chipyard/src/main/scala/clocking/ClockGroupDealiaser.scala
```diff
@@ -0,0 +1,51 @@
...
+import chisel3._
+
+import freechips.rocketchip.config.{Parameters}
+import freechips.rocketchip.diplomacy._
+import freechips.rocketchip.prci._
+
+/**
+  * Somewhat hacky. Since not all clocks in a clock group specify a taken frequency
+  * current, this LazyModule attempts to dealias them, by finding a specified
+  * clock whose name has the longest matching prefix.
```
davidbiancolin, line 12: This comment is out-of-date.

generators/chipyard/src/main/resources/vsrc/ClockDividerN.sv
```diff
@@ -0,0 +1,40 @@
+// See LICENSE for license details.
+
+/**
+  * An unsynthesizable divide-by-N clock divider.
+  * Duty cycle is 100 * (ceil(DIV / 2)) / 2.
+  */
+
+module ClockDividerN #(parameter DIV)(output logic clk_out = 1'b0, input clk_in);
```
jwright6323, line 8: I think this is fine to put here, but we have some of this stuff in testchipip as well- it might be good to consolidate it at some point. We also have a synthesizable divide-by-N (N>=1) in there as well (I think), which is just a counter and a mux.

generators/chipyard/src/main/resources/vsrc/ClockDividerN.sv
```diff
@@ -0,0 +1,40 @@
+// See LICENSE for license details.
+
+/**
+  * An unsynthesizable divide-by-N clock divider.
+  * Duty cycle is 100 * (ceil(DIV / 2)) / 2.
```
colinschmidt, line 5: This is confusing. I would expect the duty cycle to be in percentages, but this formula doesn't seem to produce something sensible when compared to the code.
i.e. for DIV=50 it returns 1250 but for DIV=10 it returns 250, and for DIV=11 it returns 275
I think the formula should be more like `100 * (ceil(DIV/2))/DIV` which would give 50 for the first two cases above and 54.54 for the last one.

davidbiancolin: It's just wrong lol, yeah i didn't normalize correctly (your formula is correct).  This is what i get working on this late.

generators/chipyard/src/main/resources/vsrc/ClockDividerN.sv
davidbiancolin: FWIW, this one _should_ be deterministic. But i might be wrong.

generators/chipyard/src/main/scala/clocking/IdealizedPLL.scala
```diff
@@ -0,0 +1,68 @@
...
+
+  lazy val module = new LazyRawModuleImp(this) {
+    require(node.out.size == 1, "Must use a ClockGroupAggregator")
+    val (refClock, ClockEdgeParameters(_, refSinkParam, _, _)) = node.in.head
+    val (outClocks, ClockGroupEdgeParameters(_, outSinkParams,  _, _)) = node.out.head
+
+    val referenceFreq = refSinkParam.take.get.freqMHz
+    val requestedFreqs = outSinkParams.members.map(m => m.name -> m.take)
+
+    val dividedClocks = mutable.HashMap[Int, Clock]()
```
colinschmidt, line 47: Does this need to use a TreeMap or whatever firrtl/chisel use to ensure that each time we build a PLL with the same input/outputs they are named the same?
Looking at it I think it should be fine, because we never iterate over the map only access elements in it.

generators/chipyard/src/main/scala/config/RocketConfigs.scala
```diff
@@ -199,3 +199,12 @@ class TestClockRocketConfig extends Config(
   new freechips.rocketchip.subsystem.WithAsynchronousRocketTiles(8, 3) ++   // Add rational crossings between RocketTile and uncore
   new freechips.rocketchip.subsystem.WithNBigCores(1) ++
   new chipyard.config.AbstractConfig)
+
+// Multiclock Sketch
+class ForcedClockRocketConfig extends Config(
+  new chipyard.config.WithForcedTileFrequency(200) ++
```
colinschmidt, line 205: What is the assumed frequency for the rest of the system? 100Mhz?

generators/chipyard/src/main/scala/GenericAttachParams.scala
```diff
@@ -0,0 +1,38 @@
...
+    params.forceSeparateClockReset)
+}
+
+case class GenericallyAttachableTile[TT <: BaseTile](
+    tileParams: InstantiableTileParams[TT],
+    crossingParams: GenericCrossingParams,
+    lookup: LookupByHartIdImpl) extends CanAttachTile {
+  type TileType = TT
+}
+
```
jerryz123, line 38: I wonder if this generic tile params object can be PR'd into RC... it would make the rest of the config fragments cleaner.

generators/chipyard/src/main/scala/config/RocketConfigs.scala
jerryz123: I believe so, but is that because the Dealiser thing is setting everything to use the `take` frequency of the PBus, which is the only clock in the `subsystem_sbus` clock group which has its `take` parameter set? https://github.com/chipsalliance/rocket-chip/blob/6f0cf5463a0e2942cb49ab69ab048e6ba2b4378c/src/main/scala/subsystem/Configs.scala#L29-L32

What is a reasonable frequency for everything else? 1GHz?

generators/chipyard/src/main/scala/Clocks.scala
```diff
@@ -171,6 +173,40 @@ object ClockDrivers {
...
+    implicit val p = chiptop.p
+
+    // Requires existence of undriven asyncClockGroups in subsystem
+    val systemAsyncClockGroup = chiptop.lSystem match {
+      case l: BaseSubsystem if (p(SubsystemDriveAsyncClockGroupsKey).isEmpty) =>
+        l.asyncClockGroupsNode
+    }
 
+    val aggregator = ClockGroupAggregator()
+    chiptop.implicitClockSinkNode := ClockGroup() := aggregator
```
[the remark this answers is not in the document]
davidbiancolin: I think technically no because implicitClockSink is not a ClockGroup, but as far as connectivity is concerned it is.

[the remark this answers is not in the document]
davidbiancolin: Technically not because the implicitClockSinkNode is not a clock group, but connectivity-wise, yes.

generators/chipyard/src/main/scala/clocking/IdealizedPLL.scala
```diff
@@ -0,0 +1,68 @@
...
+import freechips.rocketchip.prci._
+
+import scala.collection.mutable
+
+object FrequencyUtils {
+  def computeReferenceFrequencyMHz(
+    requestedOutputs: Seq[ClockParameters],
+    maximumAllowableDivisor: Int = 0xFFFF): ClockParameters = {
+    require(requestedOutputs.nonEmpty)
+    require(!requestedOutputs.contains(0.0))
```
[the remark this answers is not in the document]
davidbiancolin: Yeah. Reasonably these checks could be promoted into `ClockParameters` if they aren't already there.

generators/chipyard/src/main/scala/clocking/IdealizedPLL.scala
davidbiancolin: Since the insertion order is the same, and since we aren't iterating through the KVPs of the map, the usual gotcha, i think it's alright? I'll check though.

generators/chipyard/src/main/scala/GenericAttachParams.scala
davidbiancolin: hmm, good point.

generators/chipyard/src/main/scala/clocking/IdealizedPLL.scala
```diff
@@ -0,0 +1,68 @@
...
+import freechips.rocketchip.config.{Parameters}
+import freechips.rocketchip.diplomacy._
+import freechips.rocketchip.prci._
+
+import scala.collection.mutable
+
+object FrequencyUtils {
+  def computeReferenceFrequencyMHz(
+    requestedOutputs: Seq[ClockParameters],
+    maximumAllowableDivisor: Int = 0xFFFF): ClockParameters = {
```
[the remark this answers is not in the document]
davidbiancolin: It is. I need to look at verilator performance when using large divisions, and make sure everything works alright.

davidbiancolin:
I'm rebasing this against dev.

Issue #606, opened 2020-06-18
Re-reference Ariane as CVA6


jerryz123:
**Impact**: software

https://github.com/openhwgroup/cva6

faddat:
I'll take it.

The correct way to do this would be to make changes in `dev` which will be promoted to `master`, correct?

jerryz123:
Yeah, PR into `dev`.

alonamid:
PR #710 fixes this

PR #661, opened 2020-08-25
ariane -> cva6

faddat:
**Related issue**:  #606 

**Type of change**:  other enhancement

**Impact**:  other

**Release Notes**
Renamed Ariane to CVA6


.circleci/defaults.sh
```diff
@@ -56,7 +56,7 @@ mapping["chipyard-boom"]=" CONFIG=SmallBoomConfig"
 mapping["chipyard-blkdev"]=" CONFIG=SimBlockDeviceRocketConfig"
 mapping["chipyard-hwacha"]=" CONFIG=HwachaRocketConfig"
 mapping["chipyard-gemmini"]=" CONFIG=GemminiRocketConfig"
-mapping["chipyard-ariane"]=" CONFIG=ArianeConfig"
+mapping["chipyard-cva6"]=" CONFIG=Cva6Config"
```
colinschmidt, line 59: I think every place we have `Ariane` we should have `CVA6` as that seems to be their preferred capitalization.

generators/firechip/src/main/scala/TargetConfigs.scala
```diff
@@ -179,10 +179,10 @@ class SupernodeFireSimRocketConfig extends Config(
   new FireSimRocketConfig)
 
 //**********************************************************************************
-//* Ariane Configurations
+//* CVA6 Configurations
 //*********************************************************************************/
-class FireSimArianeConfig extends Config(
+class FireSimCva6Config extends Config(
```
jerryz123, line 184: This requires changing the default build recipes defined in FireSim: 
https://github.com/firesim/firesim/blob/master/deploy/sample-backup-configs/sample_config_build_recipes.ini

.circleci/defaults.sh
faddat: Sure, I can do that.

generators/firechip/src/main/scala/TargetConfigs.scala
faddat: OK.  

I think the whole thing requires changing the wrapper for Ariane, as well, right?

jerryz123: Yes, that submodule would need to be updated and bumped as well. Our standard routine is to open PRs for and submodules which need to be bumped, and link to them in this PR. Then we merge everything together once everything is approved.

alonamid:
subsumed by #710

PR #663, opened 2020-08-28
Add firemarshal's dev branch to its allowed-branch list

jerryz123:
**Related issue**: 

**Type of change**: bug fix


jerryz123:
Those projects don't have `dev` branches, so this change should have no effect on those.
I think its a reasonable sacrifice for the sake of verbosity.

PR #664, opened 2020-08-28
Only punch realistic subset of DebugIO through chiptop | default to JTAG+Serial

jerryz123:
`rocketchip.debug.DebugIO` has a bunch of stuff that should not be punched offchip through the IOCell generators.
Only punch out the subset of signals relevant to a real chip, namely the JTAG or clockedDMI signals.

This PR also bumps various riscv toolchain components that should have been bumped earlier. The dmiRocketconfig is also added to the regressions.

- [x] https://github.com/ucb-bar/esp-tests/pull/5
- [x] https://github.com/ucb-bar/esp-isa-sim/pull/17

**Type of change**: bug fix

**Impact**: rtl change

**Release Notes**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -199,6 +199,11 @@ jobs:
         steps:
             - prepare-rtl:
                 project-key: "chipyard-rocket"
+    prepare-chipyard-dmirocket:
+        executor: main-env
+        steps:
+            - prepare-rtl:
+                project-key: "chipyard-dmirocket"
     prepare-chipyard-sha3:
         executor: main-env
         steps:
@@ -225,11 +230,6 @@ jobs:
         steps:
             - prepare-rtl:
                 project-key: "chipyard-boom"
-    prepare-rocketchip:
-        executor: main-env
-        steps:
-            - prepare-rtl:
-                project-key: "rocketchip"
     prepare-chipyard-blkdev:
         executor: main-env
         steps:
@@ -297,6 +297,11 @@ jobs:
         steps:
             - run-tests:
                 project-key: "chipyard-rocket"
+    chipyard-dmirocket-run-tests:
+        executor: main-env
+        steps:
+            - run-tests:
+                project-key: "chipyard-dmirocket"
     chipyard-sha3-run-tests:
         executor: main-env
         steps:
@@ -323,11 +328,6 @@ jobs:
         steps:
             - run-tests:
                 project-key: "chipyard-boom"
-    rocketchip-run-tests:
-        executor: main-env
-        steps:
-            - run-tests:
-                project-key: "rocketchip"
     chipyard-hwacha-run-tests:
         executor: main-env
         steps:
@@ -451,6 +451,11 @@ workflows:
                     - install-riscv-toolchain
                     - install-verilator
 
+            - prepare-chipyard-dmirocket:
+                requires:
+                    - install-riscv-toolchain
+                    - install-verilator
+
             - prepare-chipyard-sha3:
                 requires:
                     - install-riscv-toolchain
@@ -472,11 +477,6 @@ workflows:
                     - install-verilator
 
             - prepare-chipyard-boom:
-                requires:
-                    - install-riscv-toolchain
-                    - install-verilator
-
-            - prepare-rocketchip:
                 requires:
                     - install-riscv-toolchain
                     - install-verilator
@@ -547,6 +547,10 @@ workflows:
                 requires:
                     - prepare-chipyard-rocket
 
+            - chipyard-dmirocket-run-tests:
+                requires:
+                    - prepare-chipyard-dmirocket
+
             - chipyard-sha3-run-tests:
                 requires:
                     - prepare-chipyard-sha3
@@ -566,10 +570,6 @@ workflows:
             - chipyard-boom-run-tests:
                 requires:
                     - prepare-chipyard-boom
-
-            - rocketchip-run-tests:
-                requires:
-                    - prepare-rocketchip
 
             - chipyard-hwacha-run-tests:
                 requires:
--- a/docs/Advanced-Concepts/Chip-Communication.rst
+++ b/docs/Advanced-Concepts/Chip-Communication.rst
@@ -130,7 +130,8 @@ Using the JTAG Interface
 ------------------------
 
 The main way to use JTAG with a Rocket Chip based system is to instantiate the Debug Transfer Module (DTM)
-and configure it to use a JTAG interface. The default Chipyard designs configure the DTM to use JTAG. you may attach OpenOCD and GDB to any of the default JTAG-enabled designs.
+and configure it to use a JTAG interface. The default Chipyard designs instantiate the DTM and configure it
+to use JTAG. You may attach OpenOCD and GDB to any of the default JTAG-enabled designs.
 
 Debugging with JTAG
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/Advanced-Concepts/Chip-Communication.rst
+++ b/docs/Advanced-Concepts/Chip-Communication.rst
@@ -130,38 +130,7 @@ Using the JTAG Interface
 ------------------------
 
 The main way to use JTAG with a Rocket Chip based system is to instantiate the Debug Transfer Module (DTM)
-and configure it to use a JTAG interface (by default the DTM is setup to use the DMI interface mentioned above).
-
-Creating a DTM+JTAG Config
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-First, a DTM config must be created for the system that you want to create.
-This step is similar to the DMI simulation section within the :ref:`Starting the TSI or DMI Simulation` section.
-The configuration is very similar to a DMI-based configuration. The main difference
-is the addition of the ``WithJtagDTM`` config fragment that configures the instantiated DTM to use the JTAG protocol as the
-bringup method.
-
-.. literalinclude:: ../../generators/chipyard/src/main/scala/config/RocketConfigs.scala
-    :language: scala
-    :start-after: DOC include start: JtagRocket
-    :end-before: DOC include end: JtagRocket
-
-Building a DTM+JTAG Simulator
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-After creating the config, call the ``make`` command like the following to build a simulator for your RTL:
-
-.. code-block:: bash
-
-    cd sims/verilator
-    # or
-    cd sims/vcs
-
-    make CONFIG=jtagRocketConfig
-
-In this example, the simulation will use the config that you previously specified, as well as set
-the other parameters that are needed to satisfy the build system. After that point, you
-should have a JTAG enabled simulator that you can attach to using OpenOCD and GDB!
+and configure it to use a JTAG interface. The default Chipyard designs configure the DTM to use JTAG. you may attach OpenOCD and GDB to any of the default JTAG-enabled designs.
 
 Debugging with JTAG
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/generators/chipyard/src/main/scala/IOBinders.scala
+++ b/generators/chipyard/src/main/scala/IOBinders.scala
@@ -7,6 +7,7 @@ import chisel3.experimental.{Analog, IO}
 import freechips.rocketchip.config.{Field, Config, Parameters}
 import freechips.rocketchip.diplomacy.{LazyModule, LazyModuleImpLike}
 import freechips.rocketchip.devices.debug._
+import freechips.rocketchip.jtag.{JTAGIO}
 import freechips.rocketchip.subsystem._
 import freechips.rocketchip.system.{SimAXIMem}
 import freechips.rocketchip.amba.axi4.{AXI4Bundle, AXI4SlaveNode, AXI4MasterNode, AXI4EdgeParameters}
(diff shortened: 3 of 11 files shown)


.circleci/config.yml
```diff
@@ -225,11 +230,6 @@ jobs:
         steps:
             - prepare-rtl:
                 project-key: "chipyard-boom"
-    prepare-rocketchip:
-        executor: main-env
-        steps:
-            - prepare-rtl:
-                project-key: "rocketchip"
```
jerryz123, line 232: The "rocketchip" set of tests did the exact same thing as the "chipyard-rocket" set of tests, so I removed these.

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -163,30 +165,72 @@ object AddIOCells {
...
-    resetctrlPortOpt.foreach(_.suggestName("resetctrl"))
-
-    psdPort.suggestName("psd")
-    (psdPort, resetctrlPortOpt, debugPortOpt, psdIOs ++ debugIOs ++ resetctrlIOs)
+  def debug(system: HasPeripheryDebugModuleImp)(implicit p: Parameters): (Seq[Bundle], Seq[IOCell]) = {
+    system.debug.map { debug =>
+      val tlbus = system.outer.asInstanceOf[BaseSubsystem].locateTLBusWrapper(p(ExportDebug).slaveWhere)
+      val debug_clock = Wire(Clock()).suggestName("debug_clock")
+      val debug_reset = Wire(Reset()).suggestName("debug_reset")
+      debug_clock := false.B.asClock
```
colinschmidt, line 178: Isn't this dead code? Does boring utils require this?

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -163,30 +165,72 @@ object AddIOCells {
...
+      debug_reset := false.B
+      BoringUtils.bore(tlbus.module.clock, Seq(debug_clock))
+      BoringUtils.bore(tlbus.module.reset, Seq(debug_reset))
+
+      // We never use the PSDIO, so tie it off on-chip
+      system.psd.psd.foreach { _ <> 0.U.asTypeOf(new PSDTestMode) }
+
+      // Set resetCtrlOpt with the system reset
```
colinschmidt, line 186: Typos in comment. The variable here is just called `resetctrl` not `resetCrtlOpt`,  and its driven by the reset that drives the debug module.

IMO this comment doesn't really add anything over reading the code below.

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -163,30 +165,72 @@ object AddIOCells {
...
+          t.in.req := false.B
+          t.out.ack := t.out.req
+        }
+
+        // Tie off disableDebug
+        d.disableDebug.foreach { d => d := false.B }
+
+        // Drive JTAG on-chip IOs
+        d.systemjtag.map { j =>
+          j.reset := debug_reset
```
colinschmidt, line 201: In the rocket-chip method this is an async reset. Are you sure its safe to reset this with what I assume is a synchronous reset?

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -163,30 +165,72 @@ object AddIOCells {
...
+          j.mfr_id := system.p(JtagDTMKey).idcodeManufId.U(11.W)
+          j.part_number := system.p(JtagDTMKey).idcodePartNum.U(16.W)
+          j.version := system.p(JtagDTMKey).idcodeVersion.U(4.W)
+        }
+      }
+
+
+      // Connect DebugClockAndReset to system implicit clock. TODO this should use the clock of the bus the debug module is attached to
+
```
colinschmidt, line 210: Why are there so many empty new lines here?

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -163,30 +165,72 @@ object AddIOCells {
...
+        d.systemjtag.map { j =>
+          j.reset := debug_reset
+          j.mfr_id := system.p(JtagDTMKey).idcodeManufId.U(11.W)
+          j.part_number := system.p(JtagDTMKey).idcodePartNum.U(16.W)
+          j.version := system.p(JtagDTMKey).idcodeVersion.U(4.W)
+        }
+      }
+
+
+      // Connect DebugClockAndReset to system implicit clock. TODO this should use the clock of the bus the debug module is attached to
```
colinschmidt, line 209: You've fixed this todo

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -364,40 +408,53 @@ class WithTieOffL2FBusAXI extends OverrideIOBinder({
...
     }
-    Seq((Seq(psdPort) ++ resetctrlOpt ++ debugPortOpt.toSeq, Nil, Some(harnessFn)))
+    Seq((ports, iocells, Some(harnessFn)))
   }
 })
 
-// TODO we need to rethink what this does. The current system punches out excessive IOs.
-// Some of the debug clock/reset should be driven from on-chip
-class WithSimDebug extends OverrideIOBinder({
+class WithTiedOffDebug extends OverrideIOBinder({
```
colinschmidt, line 431: Are we going to add an option that actually removes the debug IOs from going off-chip? This currently just makes the debug unit unusable without making it so that CAD tools would be able to prune portions of it.

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -364,40 +408,53 @@ class WithTieOffL2FBusAXI extends OverrideIOBinder({
...
-      val dtm_success = Wire(Bool())
-      Debug.connectDebug(debugPortOpt, resetctrlPortOpt, psdPort, th.harnessClock, th.harnessReset.asBool, dtm_success)(system.p)
-      when (dtm_success) { th.success := true.B }
-      th.dutReset := th.harnessReset.asBool | debugPortOpt.map { debug => AsyncResetReg(debug.ndreset).asBool }.getOrElse(false.B)
+      ports.map {
+        case d: ClockedDMIIO =>
+          d.dmi.req.valid := false.B
+          d.dmi.req.bits  := DontCare
+          d.dmi.resp.ready := true.B
+          d.dmiClock := th.harnessClock
```
colinschmidt, line 440: Why can't we make these `false.B` and `true.B`?

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -364,40 +408,53 @@ class WithTieOffL2FBusAXI extends OverrideIOBinder({
...
+          d.dmiReset := th.harnessReset
+        case j: JTAGIO =>
+          j.TCK := true.B.asClock
+          j.TMS := true.B
+          j.TDI := true.B
+          j.TRSTn.foreach { r => r := true.B }
```
colinschmidt, line 446: Do we need to change the other fields in the bundle that contains this field? rocket-chip does this but I think if we aren't intending this to be a hardware removal tie off then its probably unnecessary.

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -163,30 +164,63 @@ object AddIOCells {
...
-    val resetctrlTuple = resetctrlOpt.map(d =>
-      IOCell.generateIOFromSignal(d, Some("iocell_resetctrl"), abstractResetAsAsync = p(GlobalResetSchemeKey).pinIsAsync))
-    val resetctrlPortOpt: Option[ResetCtrlIO] = resetctrlTuple.map(_._1)
-    val resetctrlIOs: Seq[IOCell] = resetctrlTuple.map(_._2).toSeq.flatten
-    resetctrlPortOpt.foreach(_.suggestName("resetctrl"))
-
-    psdPort.suggestName("psd")
-    (psdPort, resetctrlPortOpt, debugPortOpt, psdIOs ++ debugIOs ++ resetctrlIOs)
+  def debug(system: HasPeripheryDebugModuleImp)(implicit p: Parameters): (Seq[Bundle], Seq[IOCell]) = {
+    system.debug.map { debug =>
```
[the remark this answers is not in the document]
jerryz123: Yeah. I wish the RocketChip API made it clearer which `connect` methods should only be called in a "TestHarness" setting, and which can be called in a "Chip" setting.

generators/chipyard/src/main/scala/IOBinders.scala
jerryz123: Yes, I'll add a comment here. Leaving these wires uninitialized causes a quiet failure of `BoringUtils`.

generators/chipyard/src/main/scala/IOBinders.scala
jerryz123: I agree.

generators/chipyard/src/main/scala/IOBinders.scala
jerryz123: I think this should probably be async reset then.

generators/chipyard/src/main/scala/IOBinders.scala
jerryz123: What do you mean? 
A system with no DebugIOs going off chip can be built by setting `DebugModuleKey` to `None`. Do we want to support a case where there is a hardware debug module, but no IOs to interact with it?

jerryz123: I don't think we should expect CAD tool pruning to be the proper method for eliminating unwanted hardware. In general, IOs which appear in ChipTop, should be expected to be IOs desired on the chip.

generators/chipyard/src/main/scala/IOBinders.scala
jerryz123: I don't think any ChipTop tie-offs should be used as hardware-removal tie-offs.

generators/chipyard/src/main/scala/IOBinders.scala
colinschmidt: Is that a bug we should report?

jerryz123: I'll ask chisel-dev

generators/chipyard/src/main/scala/IOBinders.scala
colinschmidt: I think the question is a little more nuanced than that. If the system requires this to be brought up with the  top-level async reset and not one that has been synchronized on-chip to avoid some sort of reset-race we might need to adjust the entire reset strategy for this block.
Fortunately I think this will show up in simulation (and would show up in the first few cycles), so assuming we get a jtag only version to run code it is probably fine.

generators/chipyard/src/main/scala/IOBinders.scala
colinschmidt: Ok I didn't realize you could set `DebugModuleKey` to `None`.
The use case for this function then is to support implementations that have a debug module but don't want to use it now? What does this provide? Can't we just always connect it and let the outer level simulation decide which to drive at runtime?

jerryz123: > Ok I didn't realize you could set DebugModuleKey to None.
This is untested, but I'll verify this works. I assume we don't need to prioritize this use case? 

> Can't we just always connect it and let the outer level simulation decide which to drive at runtime?
If we attach both SimDTM and SimSerial in the TestHarness, than both bringup methods will read the fesvr args and attempt to communicate with the DUT, which causes undefined behavior. The primary use case is to support systems with both clockedDMI and serial interfaces.

Perhaps we should default to having the debug module use JTAG, instead of clockedDMI. In that case, `WithTiedOffDebug` would not be useful, since we can support simulations with both SimSerial and SimJTAG
.

generators/chipyard/src/main/scala/IOBinders.scala
jerryz123: This is expected (but perhaps, uintuitive) behavior. I'll leave a comment explaining why this is necessary

generators/chipyard/src/main/scala/IOBinders.scala
colinschmidt: Hmm. I wonder if we can make SimDTM or SimSerial able to cooperate with the other, and then eliminate this tieOff altogether.

jerryz123: That may be challenging, as they instantiate separate `htif_t` objects. We could possibly make this work, but should we? We don't build real chips with the clockedDMI interface, so perhaps we should just make the default case more accurate, with JTAG+Serial, instead of clockedDMI+Serial.

colinschmidt: Yeah I'm happy making that the default. I think it might be nice to `assert` that we don't support DMI+Serial.

generators/chipyard/src/main/scala/IOBinders.scala
jerryz123: Tested working with JTAG-only

generators/chipyard/src/main/scala/IOBinders.scala
jerryz123: Its a bit tricky to `assert` this actually, since the IOBinders are by design isolated from each other.
I think its ok to not assert on this for now, since the dmiRocketConfig provides an example on supporting a DMI+Serial system, and since most people seem to want either serial interface or JTAG.

colinschmidt: Ok. How does that config/system figure out which interface is used for hosted-bringup/testing?

jerryz123: It has `WithTiedOffSerial`, which does not attach the `SimSerial` widget.

colinschmidt: Ok. I guess whoever really uses that config first can figure out the software issues.

docs/Advanced-Concepts/Chip-Communication.rst
```diff
@@ -130,38 +130,7 @@ Using the JTAG Interface
...
-    cd sims/verilator
-    # or
-    cd sims/vcs
-
-    make CONFIG=jtagRocketConfig
-
-In this example, the simulation will use the config that you previously specified, as well as set
-the other parameters that are needed to satisfy the build system. After that point, you
-should have a JTAG enabled simulator that you can attach to using OpenOCD and GDB!
+and configure it to use a JTAG interface. The default Chipyard designs configure the DTM to use JTAG. you may attach OpenOCD and GDB to any of the default JTAG-enabled designs.
```
colinschmidt, line 133: "The default Chipyard designs instantiate the DTM and configure it to use JTAG. You may attach OpenOCD and GDB to any of default JTAG-enabled designs."

generators/chipyard/src/main/scala/IOBinders.scala
jerryz123: To be clear, the included dmiRocketConfig works. What doesn't work is a config with
```
new WithSimSerial ++
new WithSimDebug ++
new WithDMIDTM
```

colinschmidt (approved):
LGTM

generators/chipyard/src/main/scala/IOBinders.scala
colinschmidt: yeah I understand. The software I'm talking about is the FESVR side of things that would let you use serial as off-chip memory while using the DMI to do interactive debugging.

PR #666, opened 2020-09-02
Initial draft of local FPGA support, targeting Arty

jamesdunn:
**Type of change**:  new feature

**Impact**: rtl change

**Release Notes**

This is a first draft of local FPGA support, initially targeting the Xilinx ARTY dev board. Most of this initial implementation has been taken from SiFive's Freedom platform. I have created a subdirectory `fpga` where the Chisel, Makefiles, bootrom, and fpga-shells submodule live. fpga-shells has not been forked or modified other than to update the commit from that included with Freedom. I am open to changing the directory structure or naming, since perhaps a top-level `fpga` directory could be confusing when FireSim lives in `generators`.

I have also added initialization to the init scripts.

To test this implementation, run
```
./scripts/init-fpga.sh
```
chdir into fpga, and run:
```
make -f Makefile.e300artydevkit verilog
make -f Makefile.e300artydevkit mcs
```
to generate the Verilog and bitfile. It can then be programmed to an Arty with Vivado and interacted with via the freedom-e-sdk.

To build with a MMIO Chipyard generator like GCD, add its mixin to `fpga/src/main/scala/arty` and change the base address in GCD.scala to something like 0x10017000. The re-run the make steps above. I've written a simple GCD program in freedom-e-sdk that tests the GCD hardware, and it is working.

As mentioned in PR #669, all `vsrcs` should be included with the `addResources` FIRRTL utility.


fpga/src/main/scala/arty/FPGAChip.scala
```diff
@@ -0,0 +1,193 @@
...
+import sifive.blocks.devices.spi._
+
+import sifive.fpgashells.shell.xilinx.artyshell.{ArtyShell}
+import sifive.fpgashells.ip.xilinx.{IBUFG, IOBUF, PULLUP, PowerOnResetFPGAOnly}
+
+//-------------------------------------------------------------------------
+// E300ArtyDevKitFPGAChip
+//-------------------------------------------------------------------------
+
+class E300ArtyDevKitFPGAChip(implicit override val p: Parameters) extends ArtyShell {
```
alonamid, line 21: This might be a total nit, but I think FireSim uses FPGATop for its top level FPGA name. I'm wondering whether we want any type of naming convention consistency between the two? @davidbiancolin @sagark , thoughts?

alonamid:
A couple of organizational discussion points:
- Do we want to put this under `chipyard/fpga` vs. `chipyard/sims/fpga`? One the one hand, it's definitely not a sim, so from that perspective, putting it in the top-level is more correct. One the other hand, we do want to encourage people to understand the difference between FPGA prototyping and emulation/simulation, and this organization might just lead them automatically to prototyping without even considering thinking about what is the actual solution that fits their problem. I tend to agree with the current organization (chipyard/fpga), but I thought it's worth opening up the discussion.

- In `fpga/src/main/scala/arty/` , should there be an additional level of hierarchy (`fpga/src/main/scala/platforms/arty/`) looking forward to other FPGA platforms?

fpga/src/main/scala/arty/FPGAChip.scala
jerryz123: This class should eventually be similar to the SW sim `TestHarness` class, which is also similar to firesim's `FireSim` class. Perhaps `TestHarness`, `FireSimHarness`, `FPGAHarness`?

build.sbt
```diff
@@ -217,3 +217,10 @@ lazy val firechip = conditionalDependsOn(project in file("generators/firechip"))
     testGrouping in Test := isolateAllTests( (definedTests in Test).value ),
     testOptions in Test += Tests.Argument("-oF")
   )
+lazy val fpgaShells = (project in file("./fpga/fpga-shells"))
+  .dependsOn(rocketchip, sifive_blocks)
+  .settings(commonSettings)
+
+lazy val freedomPlatforms = (project in file("./fpga"))
```
colinschmidt, line 224: I think we should just call it `fpgaPlatforms` or `fpgaPrototyping`. We're going to rapidly evolve away from the original (unmaintained) sources.

fpga/src/main/scala/arty/System.scala
```diff
@@ -0,0 +1,51 @@
...
+import sifive.blocks.devices.pwm._
+import sifive.blocks.devices.spi._
+import sifive.blocks.devices.uart._
+import sifive.blocks.devices.i2c._
+
+//-------------------------------------------------------------------------
+// E300ArtyDevKitSystem
+//-------------------------------------------------------------------------
+
+class E300ArtyDevKitSystem(implicit p: Parameters) extends RocketSubsystem
```
colinschmidt, line 24: I think even this first step should be putting a chipyard based generator on the fpga.

fpga/src/main/scala/arty/Config.scala
```diff
@@ -0,0 +1,65 @@
...
+  case PeripheryI2CKey => List(
+    I2CParams(address = 0x10016000))
+  case PeripheryMockAONKey =>
+    MockAONParams(address = 0x10000000)
+  case PeripheryMaskROMKey => List(
+    MaskROMParams(address = 0x10000, name = "BootROM"))
+})
+
+// Freedom E300 Arty Dev Kit Peripherals
+class E300ArtyDevKitConfig extends Config(
```
colinschmidt, line 55: I'd like this to be built up from chipyard config fragments.
I'm glad this commit is here and will be useful to start moving forward but I think before this is merged we should get to something that looks like this config but made of chipyard config fragments running on the fpga.

fpga/common.mk
```diff
@@ -0,0 +1,119 @@
...
+
+export BUILD_DIR
+
+EXTRA_FPGA_VSRCS ?=
+PATCHVERILOG ?= ""
+BOOTROM_DIR ?= ""
+
+base_dir=$(abspath ..)
+export rocketchip_dir := $(base_dir)/generators/rocket-chip
+SBT ?= java -jar $(rocketchip_dir)/sbt-launch.jar ++2.12.10
```
colinschmidt, line 31: I think we want to dedupe these variables as much as possible to integrate with the chipyard make system.

colinschmidt (changes requested):
Not a full review yet.
I think this is a good start but before we merge I think we should hit a few more concrete integration goals.

- [ ] Use the chipyard make system to run the scala compile. This would make this directory more similar to e.g. the vlsi directory where the local makefile in `fpga/Makefile` is just responsible for turning verilog into a bitstream.

- [ ] Build up a config that looks like the e300arty config from chipyard config fragments, and use it as the default config. Hoppefully this can be done while leaving the harness/platform specialization as it stands currently in this PR.

fpga/common.mk
abejgonzalez: I think I solved this with #669

abejgonzalez:
Closing for now. Hope to have an update by next release.

PR #667, opened 2020-09-03
Don't run find in base_dir to avoid slow filesystem search

jerryz123:
This should drastically improve performance for users who store lots of stuff in `$(base_dir)`.
In my workspace, running out of `/nscratch`, this change improved runtime of `make` with no source changes from 60 seconds to 3 seconds.

**Related issue**: 

**Type of change**: bug fix

**Impact**: software change

**Release Notes**


colinschmidt (approved):
LGTM

PR #669, opened 2020-09-04
Misc Additions to Local FPGA Branch

abejgonzalez:
**Related issue**: #666 

**Impact**: rtl change + other

**Release Notes**
This builds off of #666 to have it use the Chipyard make system and try to make the compile work with the newest RC changes. Note, this doesn't check for RTL functionality... still needs to be checked on an Arty board.


fpga/.gitignore
```diff
@@ -0,0 +1,3 @@
+*
```
colinschmidt, line 1: I think we should make the builds in here look more standard so we can change this ignore all to something like `generated-src`,`output`, etc.

fpga/Makefile
```diff
@@ -0,0 +1,124 @@
...
+# import other necessary rules and variables
+#########################################################################################
+include $(base_dir)/common.mk
+
+#########################################################################################
+# copy from other directory
+#########################################################################################
+all_vsrcs := \
+	$(sim_vsrcs) \
+	$(base_dir)/generators/sifive-blocks/vsrc/SRLatch.v \
```
colinschmidt, line 51: Why these included via the standard firrtl method?

fpga/Makefile
```diff
@@ -0,0 +1,124 @@
...
+	$(fpga_dir)/common/vsrc/PowerOnResetFPGAOnly.v \
+	$(build_dir)/$(long_name).rom.v
+
+#########################################################################################
+# build rom for the fpga
```
colinschmidt, line 56: rom->bootrom

fpga/Makefile
```diff
@@ -0,0 +1,124 @@
...
+#########################################################################################
+# needed for bootrom makefile
+export BUILD_DIR=$(build_dir)
+export ROCKETCHIP_DIR
+export LONG_NAME=$(long_name)
+export ROMCONF=$(build_dir)/$(long_name).rom.conf
+
+romgen := $(build_dir)/$(long_name).rom.v
```
colinschmidt, line 64: I don't like this name. It is the generated rom not a rom generator. I think `rom_v` `rom_file` etc are better names.

fpga/Makefile
```diff
@@ -0,0 +1,124 @@
...
+.PHONY: romgen
+romgen: $(romgen)
+
+#########################################################################################
+# vivado rules
+#########################################################################################
+# combine all sources into single .F
+f := $(build_dir)/$(long_name).vsrcs.F
+$(f): $(sim_common_files) $(all_vsrcs)
+	$(foreach file,$(all_vsrcs),echo "$(file)" >> $@;)
```
colinschmidt, line 80: Does vivado handle duplicates? I think we have had dedupe code for these `*.f` files in the past.

fpga/Makefile
```diff
@@ -0,0 +1,124 @@
...
+	cat $(sim_common_files) >> $@
+
+bit := $(build_dir)/obj/$(MODEL).bit
```
colinschmidt, line 83: Again I think we should disambiguate the name of the variable for the output from the name of the target.

fpga/Makefile
```diff
@@ -0,0 +1,124 @@
...
+		-top-module "$(MODEL)" \
+		-F "$(f)" \
+		-ip-vivado-tcls "$(shell find '$(build_dir)' -name '*.vivado.tcl')" \
+		-board "$(BOARD)"
+
+.PHONY: bit
+bit: $(bit)
+
+# Build .mcs
+mcs := $(build_dir)/obj/$(MODEL).mcs
```
colinschmidt, line 98: As above. bad name.

fpga/Makefile
```diff
@@ -0,0 +1,124 @@
...
+$(mcs): $(bit)
+	cd $(build_dir); vivado -nojournal -mode batch -source $(fpga_common_script_dir)/write_cfgmem.tcl -tclargs $(BOARD) $@ $<
+
+.PHONY: mcs
+mcs: $(mcs)
+
+#########################################################################################
+# mircosemi rules
+#########################################################################################
+# Build Libero project
```
colinschmidt, line 108: Do we know what this is? Does it apply to the subset we are including?

fpga/bootrom/xip/Makefile
```diff
@@ -1,12 +1,17 @@
...
 CFLAGS=-march=rv32imac -mabi=ilp32 -O2 -std=gnu11 -Wall -I. -nostartfiles -fno-common -g
 LFLAGS=-static -nostdlib
 
-dtb := $(BUILD_DIR)/$(CONFIG_PROJECT).$(CONFIG).dtb
-$(dtb): $(BUILD_DIR)/$(CONFIG_PROJECT).$(CONFIG).dts
+dtb := $(BUILD_DIR)/$(LONG_NAME).dtb
+$(dtb): $(BUILD_DIR)/$(LONG_NAME).dts
 	dtc -I dts -O dtb -o $@ $<
 
 .PHONY: dtb
```
colinschmidt, line 17: Hopefully we can apply the same `s/filename[^:]/filename_file/` rewrite rule to all these intermediate files.

fpga/bootrom/xip/Makefile
```diff
@@ -35,11 +40,11 @@ hex: $(hex)
```
colinschmidt, line 40: I can't comment on the above lines since they aren't in the diff for this PR, but I have one question.
Is the hardcoded target_address coming from an RTL config fragment? If so we should think about how to do this without repeating ourselves.

fpga/bootrom/xip/Makefile
```diff
@@ -35,11 +40,11 @@ hex: $(hex)
...
 $(romgen): $(hex)
-	$(rocketchip_dir)/scripts/vlsi_rom_gen $(ROMCONF) $< > $@
+	$(ROCKETCHIP_DIR)/scripts/vlsi_rom_gen $(ROMCONF) $< > $@
 
 .PHONY: romgen
 romgen: $(romgen)
 
 .PHONY: clean
 clean::
-	rm -rf $(hex) $(elf)
```
colinschmidt, line 45: Nice

fpga/src/main/scala/arty/Platform.scala
```diff
@@ -51,6 +51,7 @@ class E300ArtyDevKitPlatformIO(implicit val p: Parameters) extends Bundle {
 //-------------------------------------------------------------------------
 
 class E300ArtyDevKitPlatform(implicit val p: Parameters) extends Module {
+  //val sys = Module(LazyModule(new E300ArtyDevKitSystem).module) This can be DigitalTop?
```
colinschmidt, line 54: Is this just remnants from some other testing?

fpga/src/main/scala/arty/System.scala
```diff
@@ -31,6 +31,12 @@ class E300ArtyDevKitSystem(implicit p: Parameters) extends RocketSubsystem
     with HasPeripheryGPIO
     with HasPeripheryPWM
     with HasPeripheryI2C {
+  val bootROM  = p(BootROMLocated(location)).map { BootROM.attach(_, this, CBUS) }
+  val maskROMs = p(MaskROMLocated(location)).map { MaskROM.attach(_, this, CBUS) }
+
+  val maskROMResetVectorSourceNode = BundleBridgeSource[UInt]()
+  tileResetVectorNexusNode := maskROMResetVectorSourceNode
```
colinschmidt, line 38: It seems weird to support both `bootROM` and `maskROM` above but then not do some connection here.
Do we require a `maskROM` but optionally allow a `bootROM` that never can be the reset location?

generators/utilities/src/main/scala/Simulator.scala
```diff
@@ -82,26 +85,31 @@ object GenerateSimFiles extends App with HasGenerateSimConfig {
...
-  ) ++ (sim match { // simulator specific files to include
-    case VerilatorSimulator => Seq(
-      "/csrc/emulator.cc",
-      "/csrc/verilator.h",
-    )
-    case VCSSimulator => Seq(
-      "/vsrc/TestDriver.v",
-    )
-  })
+    ) ++ (sim match {
```
colinschmidt, line 89: Do we need this empty `Seq` at the beginning?

colinschmidt (approved):
Seem pretty good. I think we don't need to consume the weird style habits in the makefile, and instead can make the names more similar to the rest of chipyard.

Also obviously testing would be good, but I assume that can happen after the merge into the orig PR.

Not blocking merge since this is just a PR to a PR

fpga/Makefile
```diff
@@ -0,0 +1,124 @@
...
+SBT_PROJECT       := freedomPlatforms
+MODEL             := E300ArtyDevKitFPGAChip
+VLOG_MODEL        := E300ArtyDevKitFPGAChip
+MODEL_PACKAGE     := sifive.freedom.everywhere.e300artydevkit
+CONFIG            := E300ArtyDevKitConfig
+CONFIG_PACKAGE    := sifive.freedom.everywhere.e300artydevkit
+GENERATOR_PACKAGE := chipyard
+TB                := none # unused
+TOP               := E300ArtyDevKitPlatform
+
```
jerryz123, line 30: Should the goal be to eventually eliminate most of these overrides? I believe eventually everything can just live as part of the core chipyard generator.

colinschmidt: I think thats related to the other checkbox in my review of the original PR. I believe the answer is yes.

fpga/.gitignore
abejgonzalez: Reasonable goal.

fpga/Makefile
abejgonzalez: Yes. The goal is to get rid of these. I just hard-coded them so that this particular example works all the time. Ideally, this will change with more configuration exposed.

fpga/Makefile
abejgonzalez: These are just black boxes and do not have a resource associated with them. So we would have to make a PR for these to be included through the "normal automatic" way.

abejgonzalez: For now, I just blindly included them, later I think a PR should be opened in both places to make this automatic.

fpga/Makefile
abejgonzalez: Sure. Just copied the names.

fpga/Makefile
abejgonzalez: Sure. Just copied the names.

fpga/Makefile
abejgonzalez: The `sim_common_files` is deduped. It shouldn't collide with the `all_vsrcs`. As for Vivado... I don't know.

fpga/Makefile
abejgonzalez: I don't know, I just copied this. I think we are just using the Vivado stuff.

fpga/bootrom/xip/Makefile
abejgonzalez: Yea. Im not sure. I need to look into this more... they end up using a `MaskROM` over a `BootROM` in the code which is a bit confusing to me.

fpga/bootrom/xip/Makefile
abejgonzalez: #vim>emacs

fpga/src/main/scala/arty/Platform.scala
abejgonzalez: Oh. Yea. This should be deleted.

fpga/src/main/scala/arty/System.scala
abejgonzalez: Yea. This was a hack to try and do what the original code wanted. I need to look into this.

generators/utilities/src/main/scala/Simulator.scala
abejgonzalez: Sorry? Are you referring to L90? If so, yes, you want to not include the simulation files (or you have to filter them out in the `Makefile` when using `sim_common_files`)

fpga/Makefile
colinschmidt: Issue to track?

fpga/Makefile
colinschmidt: Ok cool as long as we reusing that code.

fpga/Makefile
colinschmidt: I think we should delete it for now, rather than keep around dead code (and perhaps make users think we support this).

fpga/bootrom/xip/Makefile
colinschmidt: I believe `MaskROM` implies a specific kind of `ROM` that is reprogrammable with just a single VLSI mask update. I think there is an overload used here sometimes for an FPGA `ROM` that can be updated without the rest of the bit stream being changed.

Besides that definition, I think a `MaskROM` can be a `BootROM` (a `ROM` that contains code to boot, and is where the core resets to), but not all `BootROM`s would be `MaskROM`s and not all `MaskROM`s would be `BootROM`s.

generators/utilities/src/main/scala/Simulator.scala
colinschmidt: Oh this has changed I think since I looked at it before, or I misread before.
I thought it was:
```
Seq(
) ++ (sim match {
...
```

fpga/bootrom/xip/Makefile
abejgonzalez: So it looks like the `MaskROM` is used to load in a ROM after elaboration using `readmemh`. This differs from the BootROM which bakes the `.img` directly into the module in Chisel.

abejgonzalez: Should be good to just swap `MaskROM` for the `BootROM`

abejgonzalez:
Updated. Now uses the CY BootRom, addressed some comments, and now uses `DigitalTop` (aka can now access all CY projects). Need to see what the `MockAON` is since I don't think it can be "disabled" but so far this is getting closer.

PR #670, opened 2020-09-04
Split IOBinders into IOBinders and HarnessBinders | punch out clocks to harness for simwidgets and bridges

jerryz123:
IOBinders now generate only chip-side hardware. Specifically, IOBinders instantiate IOCells for chip-like IOs, and directly punch through the signals for simulation-only IOs.
A new set of HarnessBinders generate hardware which sits in the TestHarness (or the FireSim "harness"). A portMap created during evaluation of the IOBinders holds references to ChipTop IOs, and is used during HarnessBinder evaluation to pass references to the corresponding IOs for each HarnessBinder.

FireSim's BridgeBinders are also moved into the FireSim "harness" (outside of ChipTop, where they previously lived), and now override existing HarnessBinders. FireSim uses the same IOBinders as SW-sim, so IOCell models are also instantiated for FireSim

Due to the BridgeBinders refactoring, it was also necessary to punch out clocks to BridgeBinders/HarnessBinders for simulation-only blocks. This fixes multiclock simulations

- [x] https://github.com/firesim/icenet/pull/27
- [x] https://github.com/ucb-bar/testchipip/pull/97
- [x] https://github.com/firesim/firesim/pull/625
- [x] https://github.com/ucb-bar/barstools/pull/86

**Related issue**: 

**Impact**: rtl change

**Release Notes**


jwright6323:
Cool, this seems a lot less repetitive

docs/Customization/IOBinders.rst
```diff
@@ -1,41 +1,45 @@
...
+
 IOBinders
 =========
 
-In Chipyard we use a special ``Parameters`` key, ``IOBinders`` to instantiate IO cells in the ``ChipTop`` layer and determine what modules to bind to the IOs of a ``ChipTop`` in the ``TestHarness``.
+The ``IOBinder`` functions are responsible for instantiating IO cells and IOPorts in the ``ChipTop`` layer.
+
+``IOBinders`` are typically defined using the ``OverrideIOBinder`` or ``ComposeIOBinder`` macros. An ``IOBInder`` consists of a function matching ``Systems`` with a given trait that generates IO ports and IOCells, and returns a list of generated ports and cells.
+
+For example, the ``WithUARTIOCells`` IOBinder specifies will, for any ``System`` that might have UART ports (``HasPeripheryUARTModuleIMP``, generate ports within the ``ChipTop`` (``ports``) as well as IOCells with the appropriate type and direction (``cells2d``). This function returns a the list of generated ports, and the list of generate IOCells. The list of generated ports is passed to the ``HarnessBinders`` such that they can be connected to ``TestHarness`` devices.
```
alonamid, line 13:
```suggestion
For example, the ``WithUARTIOCells`` IOBinder will, for any ``System`` that might have UART ports (``HasPeripheryUARTModuleIMP``, generate ports within the ``ChipTop`` (``ports``) as well as IOCells with the appropriate type and direction (``cells2d``). This function returns a the list of generated ports, and the list of generate IOCells. The list of generated ports is passed to the ``HarnessBinders`` such that they can be connected to ``TestHarness`` devices.
```

docs/Customization/IOBinders.rst
```diff
@@ -1,41 +1,45 @@
...
-This function is responsible for instantiating logic inside the ``TestHarness`` to appropriately drive the ``ChipTop`` IO ports created by the ``IOBinder``.
-Conveniently, because the ``IOBinder`` is generating the port, it may also use the port inside this function, which prevents the ``BaseChipTop`` code from ever needing to access the port ``val``, thus having the ``IOBinder`` house all port specific code.
-This scheme prevents the need to have two separate binder functions for each ``System`` trait.
-When creating custom ``IOBinders`` it is important to use ``suggestName`` to name ports; otherwise Chisel will raise an exception trying to name the IOs.
-The example ``IOBinders`` demonstrate this.
+The ``HarnessBinder`` functions determine what modules to bind to the IOs of a ``ChipTop`` in the ``TestHarness``. The ``HarnessBinder`` interface is designed to be reused across various simulation modes, enabling decoupling of the target design from simulation and testing concerns.
```
alonamid, line 24:
```suggestion
The ``HarnessBinder`` functions determine what modules to bind to the IOs of a ``ChipTop`` in the ``TestHarness``. The ``HarnessBinder`` interface is designed to be reused across various simulation/implementation modes, enabling decoupling of the target design from simulation and testing concerns.
```

docs/Customization/IOBinders.rst
```diff
@@ -1,41 +1,45 @@
...
-The test harness simply ties these off, but additional logic could be inserted to perform some kind of test in the ``TestHarness``.
+ * For SW RTL simulations, the default set of ``HarnessBinders`` instantiate software-simulated models of various devices, for example external memory or UART, and connect those models to the IOs of the ``ChipTop``.
+ * For FireSim simulations, FireSim-specific ``HarnessBinders`` instantiate ``Bridges``, which faciliate cycle-accurate simulation across the simulated chip's IOs. See the FireSim documentation for more details.
+ * In the future, a Chipyard FPGA prototyping flow may use ``HarnessBinders`` to connect ``ChipTop`` IOs to other devices or IOs in the FPGA harness.
 
-.. literalinclude:: ../../generators/chipyard/src/main/scala/IOBinders.scala
-   :language: scala
-   :start-after: DOC include start: WithGPIOTiedOff
-   :end-before: DOC include end: WithGPIOTiedOff
+For FireSim simulations, the ``HarnessBinder`` attach ``Bridge`` modules (See the FireSim documentation for more details).
```
alonamid, line 30:
```suggestion
For FireSim simulations, the ``HarnessBinder`` attaches ``Bridge`` modules (See the FireSim documentation for more details). 
```

docs/Customization/IOBinders.rst
```diff
@@ -1,41 +1,45 @@
...
+ * For FireSim simulations, FireSim-specific ``HarnessBinders`` instantiate ``Bridges``, which faciliate cycle-accurate simulation across the simulated chip's IOs. See the FireSim documentation for more details.
+ * In the future, a Chipyard FPGA prototyping flow may use ``HarnessBinders`` to connect ``ChipTop`` IOs to other devices or IOs in the FPGA harness.
 
-.. literalinclude:: ../../generators/chipyard/src/main/scala/IOBinders.scala
-   :language: scala
-   :start-after: DOC include start: WithGPIOTiedOff
-   :end-before: DOC include end: WithGPIOTiedOff
+For FireSim simulations, the ``HarnessBinder`` attach ``Bridge`` modules (See the FireSim documentation for more details). 
 
+Like ``IOBinders``, ``HarnessBinders`` are defined using macros (``OverrideHarnessBinder, ComposeHarnessBinder``), and matches ``Systems`` with a given trait. However, ``HarnessBinders`` are also passed a reference to the ``TestHarness`` (``th: HasHarnessSignalReferences``) and the list of ports generated by the corresponding ``IOBinder`` (``ports: Seq[Data]``).
```
alonamid, line 32:
```suggestion
Like ``IOBinders``, ``HarnessBinders`` are defined using macros (``OverrideHarnessBinder, ComposeHarnessBinder``), and match ``Systems`` with a given trait. However, ``HarnessBinders`` are also passed a reference to the ``TestHarness`` (``th: HasHarnessSignalReferences``) and the list of ports generated by the corresponding ``IOBinder`` (``ports: Seq[Data]``).
```

docs/Customization/IOBinders.rst
```diff
@@ -1,41 +1,45 @@
...
 
-.. literalinclude:: ../../generators/chipyard/src/main/scala/IOBinders.scala
+.. literalinclude:: ../../generators/chipyard/src/main/scala/HarnessBinders.scala
    :language: scala
-   :start-after: DOC include start: WithSimAXIMem
-   :end-before: DOC include end: WithSimAXIMem
+   :start-after: DOC include start: WithUARTAdapter
+   :end-before: DOC include end: WithUARTAdapter
+
+The ``IOBinder`` and ``HarnesBinder`` system is designed to enable decoupling of concerns between target design and simulation ssystem.
```
alonamid, line 41:
```suggestion
The ``IOBinder`` and ``HarnessBinder`` system is designed to enable decoupling of concerns between the target design and the simulation system.
```

docs/Customization/IOBinders.rst
```diff
@@ -1,41 +1,45 @@
...
 
-These classes are all ``Config`` objects, which can be mixed into the configs to specify IO connection behaviors.
+For a given set of chip IOs, there may be not only multiple simulation platforms ("harnesses", so-to-speak), but also multiple simulation strategies. For example, the choice of whether to connect the backing AXI4 memory port to a accurate DRAM model (``SimDRAM``) or a simple simulated memory model (``SimAXIMem``) is isolated in ``HarnessBinders``, and does not affect target RTL generation.
```
alonamid, line 43:
```suggestion
For a given set of chip IOs, there may be not only multiple simulation platforms ("harnesses", so-to-speak), but also multiple simulation strategies. For example, the choice of whether to connect the backing AXI4 memory port to an accurate DRAM model (``SimDRAM``) or a simple simulated memory model (``SimAXIMem``) is isolated in ``HarnessBinders``, and does not affect target RTL generation.
```

generators/chipyard/src/main/scala/config/TutorialConfigs.scala
```diff
@@ -81,66 +45,26 @@ class TutorialMMIOConfig extends Config(
...
-  new freechips.rocketchip.subsystem.WithNoMMIOPort ++
-  new freechips.rocketchip.subsystem.WithNoSlavePort ++
-  new freechips.rocketchip.subsystem.WithNExtTopInterrupts(0) ++
-  new freechips.rocketchip.subsystem.WithCoherentBusTopology ++
-  new freechips.rocketchip.system.BaseConfig
+  new chipyard.config.AbstractConfig
 )
 
-// Tutorial Phase 3: Integrate a SHA3 RoCC accelerator
+// // Tutorial Phase 3: Integrate a SHA3 RoCC accelerator
```
alonamid, line 51:
```suggestion
// Tutorial Phase 3: Integrate a SHA3 RoCC accelerator
```

generators/chipyard/src/main/scala/ChipTop.scala
```diff
@@ -50,15 +48,13 @@ class ChipTop(implicit p: Parameters) extends LazyModule with HasTestHarnessFunc
...
-    withClockAndReset(implicit_clock, implicit_reset) {
-      val (_ports, _iocells, _harnessFunctions) = p(IOBinders).values.flatMap(f => f(lSystem) ++ f(lSystem.module)).unzip3
-      // We ignore _ports for now...
-      iocells ++= _iocells.flatten
-      harnessFunctions ++= _harnessFunctions.flatten
-    }
+    val (_ports, _iocells, _portMap) = ApplyIOBinders(lazySystem, p(IOBinders))
+    // We ignore _ports for now...
+    iocells ++= _iocells
+    portMap ++= _portMap
```
abejgonzalez, line 54: Where is the `portMap` defined?

generators/chipyard/src/main/scala/TestHarness.scala
```diff
@@ -39,7 +42,9 @@ class TestHarness(implicit val p: Parameters) extends Module with HasHarnessSign
   // dutReset assignment can be overridden via a harnessFunction, but by default it is just reset
   val dutReset = WireDefault(if (p(GlobalResetSchemeKey).pinIsAsync) reset.asAsyncReset else reset)
 
-  ldut.harnessFunctions.foreach(_(this))
-
+  ldut match { case d: HasTestHarnessFunctions =>
```
abejgonzalez, line 45: A nit: Since we now call `lSystem` -> `lazySystem`, can we do the same to the DUT.

docs/Customization/IOBinders.rst
abejgonzalez: Also the `HasPeripheryUARTModuleIMP` needs to be `IMP`->`Imp`

abejgonzalez (approved):
LGTM barring the comments on the docs.

abejgonzalez:
While this is certainly late, I think it might worth having the `IOBinders` work with a LazyModule `TestHarness`. For example, in the `fpga-shells` case, the `TestHarness` cannot extend the `VCU118Shell` since it is a LazyModule (technically things compile... however the harness functions are determined in the 2nd phase of Diplomacy so only non-diplomatic items can be put in the harness function). Probably out of scope of this PR though...

jerryz123:
> While this is certainly late, I think it might worth having the `IOBinders` work with a LazyModule `TestHarness`. For example, in the `fpga-shells` case, the `TestHarness` cannot extend the `VCU118Shell` since it is a LazyModule (technically things compile... however the harness functions are determined in the 2nd phase of Diplomacy so only non-diplomatic items can be put in the harness function). Probably out of scope of this PR though...

I think out-of-scope for this PR, but I agree this is an important use case to consider. What is diplomatically elaborated in the `VCU118Shell`, and how would `HarnessBinders` interact with them, supposing that `HarnessBinders` supported being evaluated in a `Lazy` context?

abejgonzalez:
> > While this is certainly late, I think it might worth having the `IOBinders` work with a LazyModule `TestHarness`. For example, in the `fpga-shells` case, the `TestHarness` cannot extend the `VCU118Shell` since it is a LazyModule (technically things compile... however the harness functions are determined in the 2nd phase of Diplomacy so only non-diplomatic items can be put in the harness function). Probably out of scope of this PR though...
> 
> I think out-of-scope for this PR, but I agree this is an important use case to consider. What is diplomatically elaborated in the `VCU118Shell`, and how would `HarnessBinders` interact with them, supposing that `HarnessBinders` supported being evaluated in a `Lazy` context?

I need to do more experimenting but it seems like to "place" an `Overlay` which elaborates the top-level IOs of the TestHarness (and adds the `tcl`/`*dc` collateral), you need to create `BundleBridgeSource` of the `ChipTop` IOs and pass that into the `Overlay` "place" function. This requires you to have the harness functions be called in the 1st Diplomacy phase.

In the old version of the `IOBinders`, there was nothing stopping you calling the functions in the `foreach` in the 1st Diplomacy area. The problem was that the `harnessFunctions` array was appended to in the `ChipTop` in the module implementation. I don't know exactly how it works in this PR since that problematic code was deleted... but I assume the functionality is the same as before.

docs/Customization/IOBinders.rst
```diff
@@ -1,41 +1,45 @@
...
-This scheme prevents the need to have two separate binder functions for each ``System`` trait.
-When creating custom ``IOBinders`` it is important to use ``suggestName`` to name ports; otherwise Chisel will raise an exception trying to name the IOs.
-The example ``IOBinders`` demonstrate this.
+The ``HarnessBinder`` functions determine what modules to bind to the IOs of a ``ChipTop`` in the ``TestHarness``. The ``HarnessBinder`` interface is designed to be reused across various simulation/implementation modes, enabling decoupling of the target design from simulation and testing concerns.
 
-As an example, the ``WithGPIOTiedOff`` IOBinder creates IO cells for the GPIO module(s) instantiated in the ``System``, then punches out new ``Analog`` ports for each one.
-The test harness simply ties these off, but additional logic could be inserted to perform some kind of test in the ``TestHarness``.
+ * For SW RTL simulations, the default set of ``HarnessBinders`` instantiate software-simulated models of various devices, for example external memory or UART, and connect those models to the IOs of the ``ChipTop``.
```
colinschmidt, line 26: This is also the strategy for GL simulations.

docs/Customization/IOBinders.rst
```diff
@@ -1,41 +1,45 @@
...
+ * For FireSim simulations, FireSim-specific ``HarnessBinders`` instantiate ``Bridges``, which faciliate cycle-accurate simulation across the simulated chip's IOs. See the FireSim documentation for more details.
+ * In the future, a Chipyard FPGA prototyping flow may use ``HarnessBinders`` to connect ``ChipTop`` IOs to other devices or IOs in the FPGA harness.
 
-.. literalinclude:: ../../generators/chipyard/src/main/scala/IOBinders.scala
-   :language: scala
-   :start-after: DOC include start: WithGPIOTiedOff
-   :end-before: DOC include end: WithGPIOTiedOff
+For FireSim simulations, the ``HarnessBinder`` attaches ``Bridge`` modules (See the FireSim documentation for more details).
```
colinschmidt, line 30: This seems like a duplicate line?

generators/chipyard/src/main/scala/ChipTop.scala
```diff
@@ -50,15 +48,13 @@ class ChipTop(implicit p: Parameters) extends LazyModule with HasTestHarnessFunc
 
     // The implicit clock and reset for the system is also, by convention, used for all the IOBinders
```
colinschmidt, line 49: This comment is no longer useful

generators/chipyard/src/main/scala/Clocks.scala
```diff
@@ -120,6 +120,7 @@ object ClockingSchemeGenerators {
         }
       }}
 
+
```
colinschmidt, line 123: Spurious newline

generators/chipyard/src/main/scala/TestHarness.scala
```diff
@@ -39,7 +42,9 @@ class TestHarness(implicit val p: Parameters) extends Module with HasHarnessSign
   // dutReset assignment can be overridden via a harnessFunction, but by default it is just reset
   val dutReset = WireDefault(if (p(GlobalResetSchemeKey).pinIsAsync) reset.asAsyncReset else reset)
 
-  ldut.harnessFunctions.foreach(_(this))
-
+  lazyDut match { case d: HasTestHarnessFunctions =>
+    d.harnessFunctions.foreach(_(this))
```
colinschmidt, line 46: What are these used for now? Aren't they all applied below?

generators/chipyard/src/main/scala/config/ArianeConfigs.scala
```diff
@@ -13,7 +13,7 @@ class ArianeConfig extends Config(
   new chipyard.config.AbstractConfig)
 
 class dmiArianeConfig extends Config(
-  new chipyard.iobinders.WithTiedOffSerial ++          // Tie off the serial port, override default instantiation of SimSerial
+  new chipyard.harness.WithTiedOffSerial ++          // Tie off the serial port, override default instantiation of SimSerial
```
colinschmidt, line 16: We should be consistent about adding the spaces after the plus. It seems like you missed some.

generators/chipyard/src/main/scala/config/TracegenConfigs.scala
```diff
@@ -42,8 +50,10 @@ class NonBlockingTraceGenL2Config extends Config(
   new freechips.rocketchip.groundtest.GroundTestBaseConfig)
 
 class NonBlockingTraceGenL2RingConfig extends Config(
-  new chipyard.iobinders.WithBlackBoxSimMem ++
-  new chipyard.iobinders.WithTraceGenSuccessBinder ++
+  new chipyard.harness.WithBlackBoxSimMem ++
+  new chipyard.harness.WithTraceGenSuccess ++
+  new chipyard.iobinders.WithAXI4MemPunchthrough ++
+  new chipyard.iobinders.WithTraceGenSuccessPunchthrough ++
```
colinschmidt, line 56: You might consider having a `TraceGenAbstractConfig` to reduce the duplication here.

generators/firechip/src/main/scala/BridgeBinders.scala
```diff
@@ -25,73 +25,116 @@ import tracegen.{TraceGenSystemModuleImp}
...
 }
 
-class WithSerialBridge extends OverrideIOBinder({
-  (system: CanHavePeripherySerialModuleImp) =>
-    system.serial.foreach(s => SerialBridge(system.clock, s, MainMemoryConsts.globalName)(system.p)); Nil
+trait Unsupported {
+  require(false, "We do not support this IOCell type")
+}
+
+class FireSimAnalogIOCell extends RawModule with AnalogIOCell with Unsupported
```
colinschmidt, line 42: Wow this is quite the loaded line. I look forward to seeing this implemented several PhDs from now.

generators/chipyard/src/main/scala/DigitalTop.scala
```diff
@@ -33,7 +33,6 @@ class DigitalTop(implicit p: Parameters) extends ChipyardSystem
 class DigitalTopModule[+L <: DigitalTop](l: L) extends ChipyardSystemModule(l)
   with testchipip.CanHaveTraceIOModuleImp
   with testchipip.CanHavePeripheryBlockDeviceModuleImp
-  with testchipip.CanHavePeripherySerialModuleImp
```
colinschmidt, line 36: Why did this get removed?

generators/chipyard/src/main/scala/HarnessBinders.scala
```diff
@@ -0,0 +1,279 @@
...
+
+case object HarnessBinders extends Field[Map[String, (Any, HasHarnessSignalReferences, Seq[Data]) => Seq[Any]]](
+  Map[String, (Any, HasHarnessSignalReferences, Seq[Data]) => Seq[Any]]().withDefaultValue((t: Any, th: HasHarnessSignalReferences, d: Seq[Data]) => Nil)
+)
+
+
+object ApplyHarnessBinders {
+  def apply(th: HasHarnessSignalReferences, sys: LazyModule, map: Map[String, (Any, HasHarnessSignalReferences, Seq[Data]) => Seq[Any]], portMap: Map[String, Seq[Data]]) = {
+    val pm = portMap.withDefaultValue(Nil)
+    map.map { case (s, f) => f(sys, th, pm(s)) ++ f(sys.module, th, pm(s)) }
```
colinschmidt, line 36: The function here is expected to match on the type and therefore only perform the actual work once?

generators/chipyard/src/main/scala/HarnessBinders.scala
```diff
@@ -0,0 +1,279 @@
...
+  (system: HasPeripheryGPIOModuleImp, th: HasHarnessSignalReferences, ports: Seq[Data]) => {
+    ports.map { case p: GPIOPortIO => p <> AnalogConst(0) }
+    Nil
+  }
+})
+
+// DOC include start: WithUARTAdapter
+class WithUARTAdapter extends OverrideHarnessBinder({
+  (system: HasPeripheryUARTModuleImp, th: HasHarnessSignalReferences, ports: Seq[Data]) => {
+    UARTAdapter.connect(ports.map(_.asInstanceOf[UARTPortIO]))(system.p)
```
colinschmidt, line 72: Nit: I think we should do the type case matching for all of these.
If you do fix this nit it only seems to be UART and SPI that doesn't use the matching.

generators/chipyard/src/main/scala/HarnessBinders.scala
```diff
@@ -0,0 +1,279 @@
...
+    }
+    Nil
+  }
+})
+
+class WithSimAXIMem extends OverrideHarnessBinder({
+  (system: CanHaveMasterAXI4MemPort, th: HasHarnessSignalReferences, ports: Seq[Data]) => {
+    val p: Parameters = chipyard.iobinders.GetSystemParameters(system)
+    val clock = WireInit(false.B.asClock)
+    ports.filter(_.isInstanceOf[Clock]).map { case p: Clock => clock := p }
```
colinschmidt, line 135: isn't the filter redundant if you are matching on the type as well? I think restructuring it as in my above nit would make the duplication clear.

generators/chipyard/src/main/scala/HarnessBinders.scala
```diff
@@ -0,0 +1,279 @@
...
+      }
+      mem.io_axi4.head <> port
+    }
+    Nil
+  }
+})
+
+class WithBlackBoxSimMem extends OverrideHarnessBinder({
+  (system: CanHaveMasterAXI4MemPort, th: HasHarnessSignalReferences, ports: Seq[Data]) => {
+    val p: Parameters = chipyard.iobinders.GetSystemParameters(system)
```
colinschmidt, line 150: Most of my comments from SimAxiMem also apply here.

generators/chipyard/src/main/scala/HarnessBinders.scala
```diff
@@ -0,0 +1,279 @@
...
+      mem.io.clock := clock
+      mem.io.reset := th.harnessReset
+    }
+    Nil
+  }
+})
+
+class WithSimAXIMMIO extends OverrideHarnessBinder({
+  (system: CanHaveMasterAXI4MMIOPort, th: HasHarnessSignalReferences, ports: Seq[Data]) => {
+    val p: Parameters = chipyard.iobinders.GetSystemParameters(system)
```
colinschmidt, line 168: Same comments as the last 2.

generators/chipyard/src/main/scala/HarnessBinders.scala
```diff
@@ -0,0 +1,279 @@
...
+    val clock = WireInit(false.B.asClock)
+    ports.filter(_.isInstanceOf[Clock]).map { case p: Clock => clock := p }
+    (ports zip system.mmioAXI4Node.edges.in).zipWithIndex.map { case ((port: AXI4Bundle, edge), i) =>
+      val mmio_mem = LazyModule(new SimAXIMem(edge, size = 4096)(p))
```
colinschmidt, line 172: Is this hardcoded number really not in the target config at all?

generators/chipyard/src/main/scala/HarnessBinders.scala
```diff
@@ -0,0 +1,279 @@
...
+      withClockAndReset(clock, th.harnessReset) {
+        Module(mmio_mem.module).suggestName(s"mmio_mem_${i}")
```
colinschmidt, line 174: I think we should eliminate the zipwithindex and let chisel/firrtl to the postfixing here as we did in the other suggest names for simaximem et al.

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -39,21 +40,18 @@ import scala.reflect.{ClassTag}
...
-  Map[String, (Any) => Seq[IOBinderTuple]]().withDefaultValue((Any) => Nil)
+case object IOBinders extends Field[Map[String, (Any) => (Seq[Data], Seq[IOCell])]](
+  Map[String, (Any) => (Seq[Data], Seq[IOCell])]().withDefaultValue((Any) => (Nil, Nil))
 )
 
+object ApplyIOBinders {
+  def apply(sys: LazyModule, map: Map[String, (Any) => (Seq[Data], Seq[IOCell])]):
+      (Iterable[Data], Iterable[IOCell], Map[String, Seq[Data]]) = {
+    val r = map.map({ case (s,f) => (f(sys), s) }) ++ map.map({ case (s,f) => (f(sys.module), s) })
+    (r.flatMap(_._1._1), r.flatMap(_._1._2), r.map { t => t._2 -> t._1._1 })
```
colinschmidt, line 51: This is a little hard to read, perhaps you could name `r.map(_._1)`?

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -204,288 +218,130 @@ object AddIOCells {
...
-  (system: CanHavePeripheryIceNICModuleImp) => system.connectNicLoopback(); Nil
-})
-
-class WithSimNIC extends OverrideIOBinder({
-  (system: CanHavePeripheryIceNICModuleImp) => system.connectSimNetwork(system.clock, system.reset.asBool); Nil
+class WithSerialIOCells extends OverrideIOBinder({
+  (system: CanHavePeripherySerial) => system.serial.map({ s =>
+    val sys = system.asInstanceOf[BaseSubsystem]
+    val (port, cells) = IOCell.generateIOFromSignal(s, Some("iocell_serial"), sys.p(IOCellKey))
+    val serial_clock = Wire(Output(Clock())).suggestName("chiptop_serial_clock")
```
colinschmidt, line 245: Aren't you supposed to use your helper object here?

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -204,288 +218,130 @@ object AddIOCells {
...
-      }
-      Nil
-    }
-    Seq((peiTuples.map(_._1), peiTuples.flatMap(_._3), Some(harnessFn)))
+    val port = system.l2_frontend_bus_axi4.zipWithIndex.map({ case (m, i) =>
+      val p = IO(Flipped(DataMirror.internal.chiselTypeClone[AXI4Bundle](m))).suggestName(s"axi4_fbus_${i}")
+      m <> p
+      p
+    })
+    (port, Nil)
```
colinschmidt, line 296: Is this synchronized on chip? Or should it also be supplied a clock to help the harness generate signals that will already be synced when they get on chip?

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -204,288 +218,130 @@ object AddIOCells {
...
-    val harnessFn = (th: HasHarnessSignalReferences) => {
-      peiTuples.zipWithIndex.map { case ((port, edge, ios), i) =>
-        port := DontCare // tieoff doesn't completely tie-off, for some reason
-        port.tieoff()
-      }
-      Nil
-    }
-    Seq((peiTuples.map(_._1), peiTuples.flatMap(_._3), Some(harnessFn)))
+    val port = system.l2_frontend_bus_axi4.zipWithIndex.map({ case (m, i) =>
+      val p = IO(Flipped(DataMirror.internal.chiselTypeClone[AXI4Bundle](m))).suggestName(s"axi4_fbus_${i}")
```
colinschmidt, line 292: Something about this `Flipped` is weird to me. Isn't `m` directioned already so the cloneType should give you the correct direction?

colinschmidt (approved):
I had a lot of comments but nothing show stopping.

generators/chipyard/src/main/scala/TestHarness.scala
jerryz123: The clock/reset of ChipTop is still connected by a `harnessFunction`. I left `harnessFunctions` around for unusual cases which can't be captured by a `HarnessBinder`.

generators/chipyard/src/main/scala/DigitalTop.scala
jerryz123: `CanHavePeripherySerial` exists only as a lazy trait now, I removed the corresponding `ModuleImp` trait.

generators/chipyard/src/main/scala/HarnessBinders.scala
jerryz123: Yep. Like `IOBinders`, it expects that you won't have a case where both the `LazyModule` and the `LazyModuleImp` match some trait.

generators/chipyard/src/main/scala/HarnessBinders.scala
jerryz123: Yeah I think we can read it out of the config. Its strange that rocket-chip's `connectMMIO` hard-codes it as 4096...

generators/chipyard/src/main/scala/IOBinders.scala
jerryz123: `BoreHelper` does not add an IOCell, while we do need an IOCell for the serial clock.

My reasoning was that "fake" IOs would be directly bored out (via `BoreHelper`), while the serial-clock, as a "realistic" chip io, would get a IOCell.

Also in #673 I pull the clock into a `ClockedSerialIO` bundle, so this code will be eliminated anyways.

generators/chipyard/src/main/scala/IOBinders.scala
jerryz123: I think this should supply its clock, like the other axi4 ports. I forgot about this port because we don't have harness widgets which plug into this port.

generators/chipyard/src/main/scala/IOBinders.scala
jerryz123: You are right, this is pretty confusing...
There's actually a bug here, concatenating the maps will drop some values.

generators/chipyard/src/main/scala/IOBinders.scala
jerryz123: Its unclear to me why the `Flipped` is necessary, but there is a directionality error without it,
```
Expression axi4_fbus_0 is used as a SourceFlow but can only be used as a SinkFlow.
```

The `Flipped` was present in the old code, too.

davidbiancolin (commented):
I can't say i like the current implementation as we're losing a bunch of static type safety as we ascend through the layers. Furthermore, dynamic type problems will likely only manifest after the bulk of the dut has been elaborated...

I think a cleaner solution from a type-system perspective would have the Chip-side I/O binder return a container class of a specific type with references to all of the important pieces that a harness-side binder could draw from (as opposed to a Seq[Data]). This is essentially analagous to the digital top trait, but for chip top -- except it does not expect to compose into a larger cake. This class would have members for only the I/O that should exist, and they'll have the right types. If the harness binder wants to introspect on stuff done that can't be inferred from I/O, references to that can be passed out too.   Additionally, it can also provide a reference to the system if that harness-binder needs access to stuff that lives there, collapsing the 3-tuple into a 2-tuple. It also has the potential benefit of decoupling the harness-side binder from the digital-top-side trait if that's desired. 

The cost as i see it is more boiler plate but that might be a wash because it cleans up the harness side binder.

generators/chipyard/src/main/scala/HarnessBinders.scala
```diff
@@ -0,0 +1,279 @@
...
+import testchipip._
+
+import chipyard.HasHarnessSignalReferences
+
+import tracegen.{TraceGenSystemModuleImp}
+import icenet.{CanHavePeripheryIceNICModuleImp, SimNetwork, NicLoopback, NICKey, NICIOvonly}
+
+import scala.reflect.{ClassTag}
+
+case object HarnessBinders extends Field[Map[String, (Any, HasHarnessSignalReferences, Seq[Data]) => Seq[Any]]](
```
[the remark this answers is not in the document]
jerryz123: Hmm, doesn't seem to work. https://scastie.scala-lang.org/yDrjNL1pRc6kfaWRZfjdkw

generators/chipyard/src/main/scala/HarnessBinders.scala
```diff
@@ -0,0 +1,279 @@
...
+    SimSPIFlashModel.connect(ports.map(_.asInstanceOf[SPIChipIO]), th.harnessReset, rdOnly)(system.p)
+    Nil
+  }
+})
+
+class WithSimBlockDevice extends OverrideHarnessBinder({
+  (system: CanHavePeripheryBlockDeviceModuleImp, th: HasHarnessSignalReferences, ports: Seq[Data]) => {
+    val clock = WireInit(false.B.asClock)
+    ports.map {
+      case p: BlockDeviceIO => SimBlockDevice.connect(clock, th.harnessReset.asBool, Some(p))(system.p)
```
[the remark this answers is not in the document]
jerryz123: You mean multiple `BlockDeviceIO`?

generators/chipyard/src/main/scala/HarnessBinders.scala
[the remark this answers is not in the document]
davidbiancolin: Ah, ok. Makes sense.

generators/chipyard/src/main/scala/HarnessBinders.scala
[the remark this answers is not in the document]
davidbiancolin: Yeah. If you want.

jerryz123:
@davidbiancolin The latest commit removes most of the type matching in the `HarnessBinders`. The case where a `HarnessBinder` / `IOBinder` pair match the same `DigitalTop` trait, but expect different `ChipTop` IO bundle types, still manifests as a runtime `ClassCastException`, rather than a compile-tile type error. However, I believe the current implementation minimizes the number of classes/bundles which must be defined to support the binders system for arbitrary digital IOs.

colinschmidt:
Wow the typed version looks so much better thanks for changing that @jerryz123 and thanks @davidbiancolin for holding this to a high standard.

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -71,45 +74,60 @@ object GetSystemParameters {
...
+}
+
+class WithGPIOCells extends OverrideIOBinder({
+  (system: HasPeripheryGPIOModuleImp) => {
+    val (ports2d, cells2d) = system.gpio.zipWithIndex.map({ case (gpio, i) =>
       gpio.pins.zipWithIndex.map({ case (pin, j) =>
         val g = IO(Analog(1.W)).suggestName(s"gpio_${i}_${j}")
-        val iocell = genFn().suggestName(s"iocell_gpio_${i}_${j}")
+        val iocell = system.p(IOCellKey).gpio().suggestName(s"iocell_gpio_${i}_${j}")
         iocell.io.o := pin.o.oval
```
colinschmidt, line 131: We dropped some of the gpio pin signals here. Perhaps we should augment the genericIOCells with `pue` and `ds`?

jwright6323: When we discussed this previously those were deemed "Extra" signals that we'd add along with other non-logical signals (because they don't affect the 1/0 logical state of the IO)

jerryz123:
Firesim bumped and AGFIs regenerated.
Still waiting on approvals for https://github.com/firesim/icenet/pull/27, https://github.com/firesim/firesim/pull/625

PR #672, opened 2020-09-05
DTM only supports HTIF in DMI mode

jerryz123:
**Related issue**: 

**Type of change**: bug fix


a0u (approved):
To be more specific, the other debug transports also offer DMI access (e.g., the JTAG `dmi` register); the lack of support is in fesvr and the emulator harness rather than the hardware.

PR #673, opened 2020-09-07
Serial-tilelink backing memory

jerryz123:
This adds support for "LBWIF" backing memory through serialized tilelink. Default configs still use the AXI4 port, but more realistic "chip-like" configs can set the lbwif memory to be used as main backing memory.
Additionally, the SerialAdapter is moved offchip, as it should be considered a harness-level component that can access the chip through the serial TileLink port.

Adding support for other memory channels ("hbwif"), or using a different protocol for request serialization remains future work.

- [x] https://github.com/ucb-bar/testchipip/pull/98
- [x] https://github.com/ucb-bar/barstools/pull/89

### DIFF
--- a/generators/chipyard/src/main/scala/DigitalTop.scala
+++ b/generators/chipyard/src/main/scala/DigitalTop.scala
@@ -17,7 +17,6 @@ class DigitalTop(implicit p: Parameters) extends ChipyardSystem
   with testchipip.CanHaveBackingScratchpad // Enables optionally adding a backing scratchpad
   with testchipip.CanHavePeripheryBlockDevice // Enables optionally adding the block device
   with testchipip.CanHavePeripheryTLSerial // Enables optionally adding the tilelink-over-serial backing memory
-  with testchipip.CanHavePeripheryTSISerial // Enables optionally adding the TSI serial-adapter and port
   with sifive.blocks.devices.uart.HasPeripheryUART // Enables optionally adding the sifive UART
   with sifive.blocks.devices.gpio.HasPeripheryGPIO // Enables optionally adding the sifive GPIOs
   with sifive.blocks.devices.spi.HasPeripherySPIFlash // Enables optionally adding the sifive SPI flash controller
--- a/generators/chipyard/src/main/scala/DigitalTop.scala
+++ b/generators/chipyard/src/main/scala/DigitalTop.scala
@@ -16,7 +16,7 @@ class DigitalTop(implicit p: Parameters) extends ChipyardSystem
   with testchipip.CanHaveTraceIO // Enables optionally adding trace IO
   with testchipip.CanHaveBackingScratchpad // Enables optionally adding a backing scratchpad
   with testchipip.CanHavePeripheryBlockDevice // Enables optionally adding the block device
-  with testchipip.CanHavePeripheryTLSerial // Enables optionally adding the tilelink-over-serial backing memory
+  with testchipip.CanHavePeripheryTLSerial // Enables optionally adding the backing memory and serial adapter
   with sifive.blocks.devices.uart.HasPeripheryUART // Enables optionally adding the sifive UART
   with sifive.blocks.devices.gpio.HasPeripheryGPIO // Enables optionally adding the sifive GPIOs
   with sifive.blocks.devices.spi.HasPeripherySPIFlash // Enables optionally adding the sifive SPI flash controller
--- a/generators/chipyard/src/main/scala/DigitalTop.scala
+++ b/generators/chipyard/src/main/scala/DigitalTop.scala
@@ -16,7 +16,8 @@ class DigitalTop(implicit p: Parameters) extends ChipyardSystem
   with testchipip.CanHaveTraceIO // Enables optionally adding trace IO
   with testchipip.CanHaveBackingScratchpad // Enables optionally adding a backing scratchpad
   with testchipip.CanHavePeripheryBlockDevice // Enables optionally adding the block device
-  with testchipip.CanHavePeripherySerial // Enables optionally adding the TSI serial-adapter and port
+  with testchipip.CanHavePeripheryTLSerial // Enables optionally adding the tilelink-over-serial backing memory
+  with testchipip.CanHavePeripheryTSISerial // Enables optionally adding the TSI serial-adapter and port
   with sifive.blocks.devices.uart.HasPeripheryUART // Enables optionally adding the sifive UART
   with sifive.blocks.devices.gpio.HasPeripheryGPIO // Enables optionally adding the sifive GPIOs
   with sifive.blocks.devices.spi.HasPeripherySPIFlash // Enables optionally adding the sifive SPI flash controller
--- a/generators/chipyard/src/main/scala/HarnessBinders.scala
+++ b/generators/chipyard/src/main/scala/HarnessBinders.scala
@@ -223,47 +223,23 @@ class WithTiedOffDebug extends OverrideHarnessBinder({
 })
 
 
-class WithTiedOffTSISerial extends OverrideHarnessBinder({
-  (system: CanHavePeripheryTSISerial, th: HasHarnessSignalReferences, ports: Seq[ClockedIO[SerialIO]]) => {
-    ports.map { p => SerialAdapter.tieoff(Some(p.bits)) }
-    Nil
-  }
-})
-
-class WithSimTSISerial extends OverrideHarnessBinder({
-  (system: CanHavePeripheryTSISerial, th: HasHarnessSignalReferences, ports: Seq[ClockedIO[SerialIO]]) => {
-    ports.map { p =>
-      val ser_success = SerialAdapter.connectSimSerial(p.bits, p.clock, th.harnessReset)
-      when (ser_success) { th.success := true.B }
-    }
-    Nil
-  }
-})
-
-class WithSimTLSerial(withHarnessSerialAdapter: Boolean = false) extends OverrideHarnessBinder({
+class WithSerialAdapterTiedOff extends OverrideHarnessBinder({
   (system: CanHavePeripheryTLSerial, th: HasHarnessSignalReferences, ports: Seq[ClockedIO[SerialIO]]) => {
     implicit val p = chipyard.iobinders.GetSystemParameters(system)
     ports.map({ port =>
-      withClockAndReset(port.clock, th.harnessReset) {
-        val lRam = LazyModule(new SerialRAM(
-          p(SerialTLKey).get.width,
-          p(SerialTLKey).get.memParams.master.size,
-          p(SerialTLKey).get.memParams.master.base,
-          managerEdge = system.serdesser.get.managerNode.edges.in(0),
-          clientEdge = system.serdesser.get.clientNode.edges.out(0)
-        ))
-        val ram = Module(lRam.module)
-        ram.io.ser <> port.bits
+      val ram = SerialAdapter.connectHarnessRAM(system.serdesser.get, port, th.harnessReset)
+      SerialAdapter.tieoff(ram.module.io.tsi_ser)
+    })
+  }
+})
 
-        require(lRam.serdesser.module.mergedParams == system.serdesser.get.module.mergedParams,
-          "Mismatch between chip-side diplomatic params and testram diplomatic params")
-        if (withHarnessSerialAdapter) {
-          val success = SerialAdapter.connectSimSerial(Some(ram.io.tsi_ser), port.clock, th.harnessReset.asBool)
-          when (success) { th.success := true.B }
-        } else {
-          SerialAdapter.tieoff(Some(ram.io.tsi_ser))
-        }
-      }
+class WithSimSerial extends OverrideHarnessBinder({
+  (system: CanHavePeripheryTLSerial, th: HasHarnessSignalReferences, ports: Seq[ClockedIO[SerialIO]]) => {
+    implicit val p = chipyard.iobinders.GetSystemParameters(system)
+    ports.map({ port =>
+      val ram = SerialAdapter.connectHarnessRAM(system.serdesser.get, port, th.harnessReset)
+      val success = SerialAdapter.connectSimSerial(ram.module.io.tsi_ser, port.clock, th.harnessReset.asBool)
+      when (success) { th.success := true.B }
     })
   }
 })
--- a/generators/chipyard/src/main/scala/HarnessBinders.scala
+++ b/generators/chipyard/src/main/scala/HarnessBinders.scala
@@ -223,20 +223,48 @@ class WithTiedOffDebug extends OverrideHarnessBinder({
 })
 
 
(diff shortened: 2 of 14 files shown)


generators/chipyard/src/main/scala/HarnessBinders.scala
```diff
@@ -240,30 +244,56 @@ class WithTiedOffDebug extends OverrideHarnessBinder({
...
     Nil
   }
 })
 
+class WithSimTLSerial(withHarnessSerialAdapter: Boolean = false) extends OverrideHarnessBinder({
+  (system: CanHavePeripheryTLSerial, th: HasHarnessSignalReferences, ports: Seq[Data]) => {
+    implicit val p = chipyard.iobinders.GetSystemParameters(system)
+    ports.map({ case s: ClockedSerialIO =>
+      withClockAndReset(s.clock, th.harnessReset) {
+        val lRamslave = LazyModule(new SerialRAMSlave(
```
colinschmidt, line 273: I think we can just call this `SerialRAM` or `SerialHarnessRAM`

generators/chipyard/src/main/scala/HarnessBinders.scala
```diff
@@ -240,6 +240,34 @@ class WithSimSerial extends OverrideHarnessBinder({
...
+          managerEdge = system.serdesser.get.managerNode.edges.in(0),
+          clientEdge = system.serdesser.get.clientNode.edges.out(0)
+        ))
+        val ram = Module(lRam.module)
+        ram.io.ser <> port.bits
+
+        require(lRam.serdesser.module.mergedParams == system.serdesser.get.module.mergedParams,
+          "Mismatch between chip-side diplomatic params and testram diplomatic params")
+        if (withHarnessSerialAdapter) {
+          val success = SerialAdapter.connectSimSerial(Some(ram.io.tsi_ser), port.clock, th.harnessReset.asBool)
```
colinschmidt, line 261: It seems like configs that include both this and `WithSIMTSISerial` will instantiate two `SerialAdapter`s. Not sure if that will cause problems in simulation. I think they will be hooked up correctly but I don't think one of them will be pruned and I'm kind of worried about the VPI functions getting called multiple times.

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -250,11 +250,20 @@ class WithDebugIOCells extends OverrideIOBinder({
...
+    val (port, cells) = IOCell.generateIOFromSignal(s.getWrappedValue, Some("serial_tsi"), sys.p(IOCellKey))
+    port.suggestName("serial_tsi")
+    (Seq(port), cells)
+  }).getOrElse((Nil, Nil))
+})
+
+class WithSerialTLIOCells extends OverrideIOBinder({
+  (system: CanHavePeripheryTLSerial) => system.serial_tl.map({ s =>
+    val sys = system.asInstanceOf[BaseSubsystem]
+    val (port, cells) = IOCell.generateIOFromSignal(s.getWrappedValue, Some("serial_tl"), sys.p(IOCellKey))
```
colinschmidt, line 265: This seems wrong. The whole point of `serial` is to have a single off-chip interface that shares the TSI/TL functionality.

generators/chipyard/src/main/scala/config/RocketConfigs.scala
```diff
@@ -182,4 +182,10 @@ class DividedClockRocketConfig extends Config(
   new freechips.rocketchip.subsystem.WithNBigCores(1) ++
   new chipyard.config.AbstractConfig)
 
-
+class TLSerialRocketConfig extends Config(
+  new chipyard.harness.WithSimTLSerial(withHarnessSerialAdapter = true) ++ // add external TL backing memory, and external serial adapter
+  new testchipip.WithDefaultSerialTL ++                                    // support tilelink-over-serial
+  new testchipip.WithNoSerialTSI ++                                        // remove internal serial adapter
```
colinschmidt, line 188: This is confusing to me. I think we want to streamline these configs.

colinschmidt (changes requested):
Right now this seems a little strange to me. In my opinion there are really 2 choices to make:
Do you have a serial port going off chip?
What protocols does it support (in, out, in+out)?
This implementation often seems to end up with a strange duplication because its making the choice:
Do you have inward serial support?
Do you have outward serial support?
This ends up with duplicated code and sometimes RTL.

N.B. Usually we only build configurations that are in-only (simulation-only mostly), or in+out (almost all silicon).
I could imagine building an out-only system where jtag is used to do the in half and the serial port does the out half, but its not much more hardware to support in+out over the serial interface.

colinschmidt:
Also is there a testchipip PR with the changes here?

jerryz123:
I'll open a testchipip PR after the harnessbinders stuff goes in.

The reason I split the two serial types is because serialized-TSI and serialized-TL are really two mostly orthogonal serial interfaces. 
There are 4 possible configurations.

1. No serialized-TL with no serialized-TSI
2. No serialized-TL with serialized-TSI. This is the normal chipyard design, with an chip-side SerialAdapter talking to a harness-side SimSerial
3. Serialized-TL with no serialized-TSI. This configuration supports an external Tilelink manager for cacheable memory, and an external Tilelink client for the fbus. If the `withHarnessSerialAdapter` argument is set, the harness-side Tilelink client connection is used by a harness-side SerialAdapter.  This is the common "chip" like config.
4. Serialized-TL with serialized-TSI. This configuration supports an external Tilelink manager for cacheable memory, and an internal SerialAdapter talking to a harness-side SimSerial. This config may be bugged if the user explicitly specifies an external SerialAdapter as well. This use case is pretty unrealistic for chips, but I found it useful to support this use-case when I was debugging the external-serial-adapter.

I don't think there is much duplication to support use-case 4. The `CanHavePeripherySerial{TL/TSI}` traits, and their HarnessBinders are very different from each other. 

At a general principle, I think each Trait-IOBinder-HarnessBinder "vertical" should correspond to an interface and its use case, rather than just the interface itself. For the same reason why we have separate traits/IOBinders/HarnessBinders/ for each external AXI4 type, we should have separate traits/IOBinders/HarnessBinders for serialized tilelink and serialized TSI.

jwright6323:
Personally, I've struggled enough with the LBWIF code that I think we should scrap it and implement a simpler protocol that doesn't make the host-side RTL dependent on the target configuration (no source/sink ID bits, no address bitwidth dependency, etc.)

jerryz123:
The solution is to deprecate support for the on-chip serial-adapter, and only support serialized-tilelink.

jwright6323:
> The solution is to deprecate support for the on-chip serial-adapter, and only support serialized-tilelink.

I don't think so- if you are serializing tilelink your host RTL still needs to be aware of the chip's TL parameters. A more generic, higher-level transfer protocol (even one that has TL ports on either end), would get rid of that dependency. Even if you solved the problem of generating the host RTL automatically (as I/we have proposed in the past), you still run into trouble if you want to test multiple chips with the same FPGA bitstream.

jerryz123:
> I don't think so
Sorry, I meant the solution to the duplication between Serial-TL and Serial-TSI.

I agree improvements to the serialization protocol could reduce the host-dependency on the chip. Perhaps chiplink might be a place to look, but I think that might be over-engineered for our use case.

colinschmidt:
I think you won't be able to get around a dependency on the chip RTL (without wasting on-chip hardware). The solution we talked about today is going to do the correct inspection of chip parameters from the harness side to ensure that everything matches up.

colinschmidt:
Maybe we should support different serialization via a parameter? If it is constrained to TL on both ends and Clocked[SerialIO] in the middle people should be able to explore different variants.

jerryz123:
The testchipip PR I will open contains a `require` which verifies that the host-side serdesser and the chip-side serdesser agreed on the same parameters.

jwright6323:
> I think you won't be able to get around a dependency on the chip RTL (without wasting on-chip hardware). The solution we talked about today is going to do the correct inspection of chip parameters from the harness side to ensure that everything matches up.

I disagree, but maybe we should discuss what some reasonable tradeoffs would be in a separate forum

colinschmidt:
Sure. We can take the discussion offline.
I think there is minimal work to integrate the current working code that lives outside of chipyard into chipyard.
I also think most of it won't be thrown away if you were to switch up the serialization format.
Therefore I think this PR should go forward as discussed and we can figure out if we want to change the serialization format and what we would change it to at a later date after in person or more online discussions.

jwright6323:
I think it's fine to merge this in, but my main goal would be to prevent anyone from taping out with this, because that would then require supporting this for another year or two. We have an opportunity to break that cycle now, I think.

jerryz123:
As anyone looked at chiplink for serializing tilelink?

colinschmidt:
It wasn't public last time I needed something like it.

jerryz123:
There seems to be stuff in sifive-blocks: https://github.com/sifive/sifive-blocks/blob/master/src/main/scala/devices/chiplink/ChipLink.scala

And freedom references chiplink in a couple places.

I think there is value in preserving this simpler protocol, as long as it works and the limitations are clear.

jwright6323:
I don't think this is a "Simple" protocol; I would argue it is actually overly complex. It may be simple in concept ("Just take tilelink and serialize it"), but it's extremely fragile and difficult to debug because there's so much context dependence to each 32-bit packet.

jerryz123:
Lets move this discussion to https://github.com/ucb-bar/testchipip/pull/98

generators/chipyard/src/main/scala/DigitalTop.scala
```diff
@@ -16,7 +16,8 @@ class DigitalTop(implicit p: Parameters) extends ChipyardSystem
   with testchipip.CanHaveTraceIO // Enables optionally adding trace IO
   with testchipip.CanHaveBackingScratchpad // Enables optionally adding a backing scratchpad
   with testchipip.CanHavePeripheryBlockDevice // Enables optionally adding the block device
-  with testchipip.CanHavePeripherySerial // Enables optionally adding the TSI serial-adapter and port
+  with testchipip.CanHavePeripheryTLSerial // Enables optionally adding the tilelink-over-serial backing memory
```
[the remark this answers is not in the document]
jerryz123: No longer a concern now that on-chip serialadapter is gone.

jerryz123:
I've removed the on-chip serial-adapter support. Now all designs use an external serial-adapter. 
The external harness RAM is by default small (4KB), and uses an unused physical memory region.
The `LBWIFMemoryRocketConfig` (open to name suggestions) sets the external harness RAM to 256 MB, and sets its base address to `DRAM_BASE`, effectively replacing the functionality of the AXI4 backing memory.

generators/firechip/src/main/scala/TargetConfigs.scala
```diff
@@ -131,7 +131,7 @@ class FireSimSmallSystemConfig extends Config(
   new WithoutClockGating ++
   new WithoutTLMonitors ++
   new freechips.rocketchip.subsystem.WithExtMemSize(1 << 28) ++
-  new testchipip.WithTSI ++
+  new testchipip.WithDefaultSerialTL ++
```
colinschmidt, line 134: This small config seems like a good place to put a FireSim that test LBWIF. (reduces number of off-chip IOs of the modeled system by a lot)

colinschmidt (approved):
A few nitty comments, but LGTM.

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -250,11 +250,11 @@ class WithDebugIOCells extends OverrideIOBinder({
...
 
-class WithSerialIOCells extends OverrideIOBinder({
-  (system: CanHavePeripherySerial) => system.serial.map({ s =>
+class WithSerialTLIOCells extends OverrideIOBinder({
+  (system: CanHavePeripheryTLSerial) => system.serial_tl.map({ s =>
     val sys = system.asInstanceOf[BaseSubsystem]
-    val (port, cells) = IOCell.generateIOFromSignal(s.getWrappedValue, Some("serial"), sys.p(IOCellKey))
-    port.suggestName("serial")
+    val (port, cells) = IOCell.generateIOFromSignal(s.getWrappedValue, Some("serial_tl"), sys.p(IOCellKey))
+    port.suggestName("serial_tl")
```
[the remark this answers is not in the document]
jerryz123: Yeah that's a good point. I think we should fix that in the next barstools bump.

generators/chipyard/src/main/scala/Subsystem.scala
```diff
@@ -25,12 +25,12 @@ import freechips.rocketchip.amba.axi4._
...
 
 
-import testchipip.{DromajoHelper, CanHavePeripherySerial, SerialKey}
+import testchipip.{DromajoHelper, CanHavePeripheryTLSerial, SerialTLKey}
 
 trait CanHaveHTIF { this: BaseSubsystem =>
   // Advertise HTIF if system can communicate with fesvr
   if (this match {
-    case _: CanHavePeripherySerial if p(SerialKey) => true
+    case _: CanHavePeripheryTLSerial if p(SerialTLKey).nonEmpty => true
```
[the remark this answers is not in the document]
jerryz123: WithSerialAdapterTiedOff is a harness-configuration fragment, so it should probably not affect this chip-generator code.

generators/chipyard/src/main/scala/config/AbstractConfig.scala
```diff
@@ -14,8 +14,8 @@ class AbstractConfig extends Config(
   // The HarnessBinders control generation of hardware in the TestHarness
   new chipyard.harness.WithUARTAdapter ++                       // add UART adapter to display UART on stdout, if uart is present
   new chipyard.harness.WithBlackBoxSimMem ++                    // add SimDRAM DRAM model for axi4 backing memory, if axi4 mem is enabled
+  new chipyard.harness.WithSimSerial ++                         // add external serial-adapter and RAM
```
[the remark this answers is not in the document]
jerryz123: They are not order dependent.

generators/firechip/src/main/scala/TargetConfigs.scala
jerryz123: I'm currently uncertain if FireSim configs without the AXI4 backing memory work correctly, since they use a separate channel to write the program binary into DRAM.

jerryz123: They don't work correctly, issue opened here: https://github.com/firesim/firesim/issues/627

generators/chipyard/src/main/scala/IOBinders.scala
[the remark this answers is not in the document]
jerryz123: Wait I forgot this PR already bumps barstools. I'll just fix it here then.

generators/chipyard/src/main/scala/config/AbstractConfig.scala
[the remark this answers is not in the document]
jerryz123: They aren't order dependent

PR #674, opened 2020-09-14
Undo regression in iocells flexibility

jerryz123:
**Related issue**: 

**Type of change**: bug fix

### DIFF
--- a/generators/firechip/src/main/scala/BridgeBinders.scala
+++ b/generators/firechip/src/main/scala/BridgeBinders.scala
@@ -39,10 +39,20 @@ trait Unsupported {
   require(false, "We do not support this IOCell type")
 }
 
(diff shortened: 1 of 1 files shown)


jwright6323 (approved):
This seems fine to me, but I don't really know how the FireSim IO cells work at all. How do they handle the `Analog`s?

jerryz123:
> This seems fine to me, but I don't really know how the FireSim IO cells work at all. How do they handle the `Analog`s?

The don't. If you build a design with analog IOs, the `require` in `trait Unsupported` will complain that its unsupported.

jwright6323:
Oh, I forgot the `pad` signal is not an `Analog` for inputs/outputs. Sounds good.

PR #675, opened 2020-09-15
Improve CI build times by grouping similar builds

jerryz123:
This avoids lots of redundant time copying sources to/from build instances, and rebuilding all the Scala

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -99,11 +99,10 @@ commands:
             tools-version:
                 type: string
                 default: "riscv-tools"
+            group-key:
+                type: string
             project-key:
                 type: string
-            extra-cache-restore:
-                type: string
-                default: ""
             run-script:
                 type: string
                 default: "run-tests.sh"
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -194,219 +187,166 @@ jobs:
...
             - prepare-rtl:
                 tools-version: "esp-tools"
-                project-key: "chipyard-gemmini"
-    prepare-tracegen:
-        executor: main-env
-        steps:
-            - prepare-rtl:
-                project-key: "tracegen"
-    prepare-tracegen-boom:
+                project-key: "group-accels"
```
abejgonzalez, line 206: If we are grouping things together... wouldn't this key/parameter be better named as "group-key"?

.circleci/do-rtl-build.sh
```diff
@@ -58,12 +56,19 @@ run "export RISCV=\"$TOOLS_DIR\"; \
      export PATH=\"$REMOTE_VERILATOR_DIR/bin:\$PATH\"; \
      export VERILATOR_ROOT=\"$REMOTE_VERILATOR_DIR\"; \
      export COURSIER_CACHE=\"$REMOTE_WORK_DIR/.coursier-cache\"; \
-     make -C $REMOTE_SIM_DIR clean; \
-     make -j$REMOTE_MAKE_NPROC -C $REMOTE_SIM_DIR FIRRTL_LOGLEVEL=info JAVA_ARGS=\"$REMOTE_JAVA_ARGS\" ${mapping[$1]}"
-run "rm -rf $REMOTE_CHIPYARD_DIR/project"
-
-# copy back the final build
-
+     make -C $REMOTE_SIM_DIR clean;"
```
abejgonzalez, line 59: Why do we need a bunch of exports before the clean?  Also, why remove the `rm -rf project` line?

abejgonzalez: Nevermind on the `rm -rf project` comment... it seems like there was some duplication?

jerryz123: Technically we just need to export `RISCV` before the clean.

`rm -rf project` was duplicated, I removed one of them.

abejgonzalez (approved):
Ok LGTM

PR #676, opened 2020-09-17
Simple Divider-Only PLL for Multiclock RTL Simulation

davidbiancolin:
This subsumes #659. 

Provides the mechanism to specify clock frequencies by name per interim advise from SiFive, reusing the PLL from #659. (I've applied all the comments from the PR). At a high level, users provide "assigner" functions that accept the name of a clock and return an Option[Double]. Users can register new functions by appending to `ClockFrequencyAssignment. The clock schemes from before have been collapsed one. You request a multiclock design by (1) providing an assigner, and (2) ensuring the rest of the target has the right CDCs. 

TODOs:
- [x] Should harness side clock generation should be driven by the chosen PLL reference frequency?
- [x] Figure out reset handling
  - Handle in follow up PR
- [x] Port FireSim to use the string-based frequency specification.


generators/chipyard/src/main/scala/clocking/ClockGroupNamePrefixer.scala
```diff
@@ -0,0 +1,65 @@
...
+          case None            => Some(s"${s.name}_${idx}")
+      })
+    })})).node
+}
+
+/**
+  * [Word from on high is that Strings are in...]
+  * Overrides the take field of all clocks in a group, by attempting to apply a
+  * series of assignment functions:
+  *   (name: String) => freq-in-MHz: Option[Double]
```
jerryz123, line 45: The assigners shouldn't override any take frequencies already specified by the node, right? 
https://github.com/chipsalliance/rocket-chip/pull/2641 is giving us some control over take frequencies

davidbiancolin: I was meaning to ask about that. It seemed to me that in the short run it made sense to blow away all taken frequencies because it would be less surprising, but it's trivial to not do that.

jerryz123: If the primary means of controlling frequencies is through various `WithTileClockFreq(freq)` fragments, we could just change the underlying implementation of those fragments to use the rocketchip API when its ready.

davidbiancolin: Yeah, that sounds good. 

I'm still a little concerned about specifying absolute frequencies; i do think it would be preferable to be able to specify constraints on derived frequencies but we can punt on that.

colinschmidt: Re: absolute frequencies. I think the more important part is the relative frequencies but it is easy to calculate the ratio with 100 being the default.

generators/chipyard/src/main/scala/clocking/IdealizedPLL.scala
```diff
@@ -0,0 +1,82 @@
...
+    require(u.size == 1)
+    require(!u.head.members.contains(None),
+      "All output clocks in group must set their take parameters. Use a ClockGroupDealiaser")
+    ClockSinkParameters(
+      name = Some(s"${pllName}_reference_input"),
+      take = Some(FrequencyUtils.computeReferenceFrequencyMHz(u.head.members.flatMap(_.take)))) }
+  )
+
+/**
+  * Generates a digttal-divider-only PLL model that verilator can simulate.
```
colinschmidt, line 40: typo digital

generators/chipyard/src/main/scala/clocking/IdealizedPLL.scala
```diff
@@ -0,0 +1,82 @@
...
+class IdealizedPLL(pllName: String)(implicit p: Parameters, valName: ValName) extends LazyModule {
+  val node = IdealizedPLLNode(pllName)
+
+  lazy val module = new LazyRawModuleImp(this) {
+    require(node.out.size == 1, "Idealized PLL expects to generate a single output clock group. Use a ClockGroupAggregator")
+    val (refClock, ClockEdgeParameters(_, refSinkParam, _, _)) = node.in.head
+    val (outClocks, ClockGroupEdgeParameters(_, outSinkParams,  _, _)) = node.out.head
+
+    val referenceFreq = refSinkParam.take.get.freqMHz
+    println(s"Idealized PLL Frequency Summary")
```
colinschmidt, line 56: I think we should also dump this in an artefact, along with the requested freqs and what is being generated for them.

colinschmidt (commented):
Looks good.

generators/chipyard/src/main/resources/vsrc/ClockDividerN.sv
```diff
@@ -0,0 +1,40 @@
...
+    localparam LOW_CYCLES = DIV / 2;
+
+    generate
+        if (DIV == 1) begin
+            // This needs to be procedural because of the assignment on declaration
+            always @(clk_in) begin
+                clk_out = clk_in;
+            end
+        end else begin
+            reg [DIV_COUNTER_WIDTH - 1: 0] count = '0;
```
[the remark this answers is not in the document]
davidbiancolin: The latter. I'm lazy.

generators/chipyard/src/main/scala/Clocks.scala
```diff
@@ -78,101 +80,62 @@ object GenerateReset {
 }
 
 
-case object ClockingSchemeKey extends Field[ChipTop => Unit](ClockingSchemeGenerators.harnessClock)
+case object ClockingSchemeKey extends Field[ChipTop => Unit](ClockingSchemeGenerators.idealizedPLL)
+/**
+  * This is a Seq of assignment functions, that accept a clock name and return an optional frequency.
+  * Functions that appear later in this seq have higher precedence that earlier ones.
+  * If no function returns a non-empty value, the value specified in
+  * [[DefaultClockFrequencyKey]] will be used -- DFU.
```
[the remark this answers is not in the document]
davidbiancolin: :andrewemoji:

generators/chipyard/src/main/scala/ConfigFragments.scala
```diff
@@ -26,9 +26,11 @@ import sifive.blocks.devices.gpio._
 import sifive.blocks.devices.uart._
 import sifive.blocks.devices.spi._
 
-import chipyard.{BuildTop, BuildSystem, ClockingSchemeGenerators, ClockingSchemeKey, TestSuitesKey, TestSuiteHelper}
-
+import chipyard.{BuildTop, BuildSystem, ClockingSchemeGenerators, ClockingSchemeKey, TestSuitesKey, TestSuiteHelper, ClockNameContainsAssignment}
 
+// Imports for multiclock sketch
```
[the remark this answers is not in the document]
davidbiancolin: Yeah this can deleted.

generators/chipyard/src/main/scala/clocking/IdealizedPLL.scala
davidbiancolin: Any thoughts about the format?

generators/firechip/src/main/scala/TargetConfigs.scala
```diff
@@ -194,7 +194,7 @@ class FireSimArianeConfig extends Config(
 //* Multiclock Configurations
 //*********************************************************************************/
 class FireSimMulticlockRocketConfig extends Config(
-  new WithFireSimRationalTileDomain(2, 1) ++
+  new chipyard.config.WithTileFrequency(6400.0) ++ //lol
```
jerryz123, line 197: Wouldn't it make more sense to set pbus freq to 1600, and set this to 3200? Is this to avoid changing UART stuff?

davidbiancolin: UART and NIC stuff, yes.

generators/chipyard/src/main/scala/Clocks.scala
```diff
@@ -77,99 +79,61 @@ object GenerateReset {
...
+  */
+case object ClockFrequencyAssignersKey extends Field[Seq[(String) => Option[Double]]](Seq.empty)
```
jerryz123, line 89: Why not just make this `Map[String, Option[Double]]`

generators/firechip/src/main/scala/FireSim.scala
```diff
@@ -43,107 +34,114 @@ object NodeIdx {
...
-          if (name.contains("core")) {
-            data.clock := tile_clock
-            data.reset := ResetCatchAndSync(tile_clock, reset.asBool)
-          } else {
-            data.clock := uncore_clock
-            data.reset := reset
-          }
+      // Assign resets. The synchronization scheme is still WIP.
+      for ((name, clockBundle) <- clockGroupBundle.member.elements) {
+        if (name.contains("core")) {
```
jerryz123, line 120: Why does this only need to be synchronized for the cores?

generators/chipyard/src/main/scala/Clocks.scala
davidbiancolin: That was the original motivation. I thought about making it a regex, but then i just settled on doing a more general thing. If you wanted to use the core idx as an argument you could do that with this very simply.

generators/firechip/src/main/scala/FireSim.scala
davidbiancolin: This just matches what was done before. I've been punting on it. The general problem is: in what order should these resets be deasserted, and how should the user specify that (if they don't want to provide their own clocking scheme)? Simply synchronizing them will result in the fastest domains coming out of reset first, which may or may not not be desired. In any case, we have the same problem the idealizedPLL. 

As an aside, I have tried synchronizing everything and it seems to just work for FireSim.

Un elated, maybe we should start building our library of PRCI diplomatic widgets and write a synchronizer node. We probably also need a lengthener, as alluded to in some of henry's PRs.

jerryz123 (approved):

PR #677, opened 2020-09-22
Fix smartelf2hex.sh creating files 64x the minimum size

jerryz123:
**Type of change**: bug fix

**Impact**: software change


colinschmidt (approved):
LGTM

colinschmidt:
Can you test this with a riscv-test isa test and benchmark in addition to whatever use case uncovered the bug.

jerryz123:
Good call. The imprecise hex sizing was masking another issue, that the script did not consider the TLS segments, which elf2hex would still read.

PR #681, opened 2020-10-01
Support Placing DRAM Memory Systems In Their Own Clock Domain

davidbiancolin:
Adds an extension of CanHaveMasterAXI4MemPort that allows the user to place the memory controller (and abutted diplomatic widgets) in a separate clock domain for performance modelling purposes. I intend for this to be a placeholder until something is decided upstream.

Other changes:

- Adds a ResetSynchronizer diplonode that acts on ClockGroups. Now uniformly synchronizes all outputs from the divider-only clock generator
- Places the FASED memory model in a 1GHz async domain for FireChip targets. This has the effect of making all default targets in FireSim multiclock. 

Questions:
- Should i add a sychronizer node for ClockNodes too?
- Should we keep the IO binders that matched on the old trait?  
- By default i put everything i could in CY, let me know what i should push into testchipip and where. 
- Should there be a ClockedIO and ClockedAndResetIO? Perhaps just the latter? Maybe reset can be made an `Option`.
- Renaming clocking package -> prci to mirror RC? 

**Impact**: rtl change
**Release Notes**: Add support for putting DRAM memory systems in their own clock domains.

### DIFF
--- a/generators/chipyard/src/main/scala/Clocks.scala
+++ b/generators/chipyard/src/main/scala/Clocks.scala
@@ -8,11 +8,11 @@ import freechips.rocketchip.prci._
 import freechips.rocketchip.subsystem.{BaseSubsystem, SubsystemDriveAsyncClockGroupsKey}
 import freechips.rocketchip.config.{Parameters, Field, Config}
 import freechips.rocketchip.diplomacy.{OutwardNodeHandle, InModuleBody, LazyModule}
-import freechips.rocketchip.util.{ResetCatchAndSync, Pow2ClockDivider}
+import freechips.rocketchip.util.{ResetCatchAndSync}
 
 import barstools.iocell.chisel._
 
-import chipyard.clocking.{DividerOnlyClockGenerator, ClockGroupNamePrefixer, ClockGroupFrequencySpecifier}
+import chipyard.clocking._
 
 /**
   * Chipyard provides three baseline, top-level reset schemes, set using the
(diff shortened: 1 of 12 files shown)


hcook:
> Should i add a sychronizer node for ClockNodes too?

Yes and can you upstream both of them please 😎

davidbiancolin:
> > Should i add a sychronizer node for ClockNodes too?
> 
> Yes and can you upstream both of them please 😎

Sure thing. :)

jerryz123:
I wonder if, instead of putting the crossings in the System-internal diplomacy graph, we can insert them in the harness-side graph instead. The advantage of this would be that we no longer have to fork that rocket-chip trait.

The disadvantage would be that this solution would be more complicated then what you have already implemented.

davidbiancolin:
> I wonder if, instead of putting the crossings in the System-internal diplomacy graph, we can insert them in the harness-side graph instead (like the one defined in SimAXIMem). The advantage of this would be that we no longer have to fork any rocket-chip traits.
> 
> The disadvantage would be that this solution would be more complicated then what you have already implemented.

I think that's reasonable if we are ok with putting the crossings after all of the diplo widgets in the Subsystem trait. Relatedly, henry had some suggestions that i'll play with which might gel with an IOBinder-side approach. If that works out i can push another sketch.

davidbiancolin:
Subsumed by #690.

PR #690, opened 2020-10-09
DRAM In A Separate Domain Take 2

davidbiancolin:
This a follow up PR to #681, proposing an alternate implementation that uses a slightly different bus topology, and requires no changes to the cake trait. 

Also see ucb-bar/testchipip#106.

The difficulty is that it relies on setting the `dtsFrequency` parameter of the mbus, for our own CY clock generation purposes,  which changes the generated DTS to include a new entry for the mbus. This should be innocuous, but may not be acceptable upstream. It's exposed through everyone's favourite `p` so it can be left unset by default.

I also threw in some config fragments for other bus crossings in the future. I figure we probably want to add a pbus <-> cbus or cbus <-> sbus crossing soon.

### DIFF
--- a/generators/chipyard/src/main/scala/CustomBusTopologies.scala
+++ b/generators/chipyard/src/main/scala/CustomBusTopologies.scala
@@ -1,0 +1,100 @@
+
+package chipyard
+
+import freechips.rocketchip.config.{Field, Config, Parameters}
+import freechips.rocketchip.diplomacy._
+import freechips.rocketchip.tilelink._
+import freechips.rocketchip.util.{Location, Symmetric}
+import freechips.rocketchip.subsystem._
+
+// I'm putting this code here temporarily as I think it should be a candidate
+// for upstreaming based on input from Henry Cook, but don't wnat to deal with
+// an RC branch just yet.
+
+// For subsystem/BusTopology.scala
+
+/**
+  * Keys that serve as a means to define crossing types from a Parameters instance
+  */
+case object SubsystemCrossingParamsKey extends Field[SubsystemCrossingParams](SubsystemCrossingParams())
+case object MemoryBusCrossingTypeKey extends Field[ClockCrossingType](NoCrossing)
+
+// Biancolin: This, modified from Henry's email
+/** Parameterization of a topology containing a banked coherence manager and a bus for attaching memory devices. */
+case class CoherentBusTopologyParams(
+  sbus: SystemBusParams, // TODO remove this after better width propagation
+  mbus: MemoryBusParams,
+  l2: BankedL2Params,
+  sbusToMbusXType: ClockCrossingType = NoCrossing
+) extends TLBusWrapperTopology(
+  instantiations = (if (l2.nBanks == 0) Nil else List(
+    (MBUS, mbus),
+    (L2, CoherenceManagerWrapperParams(mbus.blockBytes, mbus.beatBytes, l2.nBanks, L2.name)(l2.coherenceManager)))),
+  connections = if (l2.nBanks == 0) Nil else List(
+    (SBUS, L2,   TLBusWrapperConnection(xType = NoCrossing, driveClockFromMaster = Some(true), nodeBinding = BIND_STAR)()),
+    (L2,  MBUS,  TLBusWrapperConnection.crossTo(
+      xType = sbusToMbusXType,
+      driveClockFromMaster = Some(true),
+      nodeBinding = BIND_QUERY))
+  )
+)
+
+// For subsystem/Configs.scala
+
+class WithCoherentBusTopology extends Config((site, here, up) => {
+  case TLNetworkTopologyLocated(InSubsystem) => List(
+    JustOneBusTopologyParams(sbus = site(SystemBusKey)),
+    HierarchicalBusTopologyParams(
+      pbus = site(PeripheryBusKey),
+      fbus = site(FrontBusKey),
+      cbus = site(ControlBusKey),
+      xTypes = SubsystemCrossingParams()),
+    CoherentBusTopologyParams(
+      sbus = site(SystemBusKey),
+      mbus = site(MemoryBusKey),
+      l2 = site(BankedL2Key),
+      sbusToMbusXType = site(MemoryBusCrossingTypeKey)))
+})
+
+/**
+  * Mixins to specify crossing types between the 5 traditional TL buses
+  *
+  * Note: these presuppose the legacy connections between buses and set
+  * parameters in SubsystemCrossingParams; they may not be resuable in custom
+  * topologies (but you can specify the desired crossings in your topology).
+  *
+  * @param xType The clock crossing type
+  *
+  */
+class WithMemoryBusCrossingType(xType: ClockCrossingType) extends Config((site, here, up) => {
+    case MemoryBusCrossingTypeKey => xType
+})
+
+class WithFrontBusCrossingType(xType: ClockCrossingType) extends Config((site, here, up) => {
+    case SubsystemCrossingParamsKey => up(SubsystemCrossingParamsKey, site)
+      .copy(fbusToSbusXType = xType)
+})
+
+class WithControlBusCrossingType(xType: ClockCrossingType) extends Config((site, here, up) => {
+    case SubsystemCrossingParamsKey => up(SubsystemCrossingParamsKey, site)
+      .copy(sbusToCbusXType = xType)
+})
+
+class WithPeripheryBusCrossingType(xType: ClockCrossingType) extends Config((site, here, up) => {
+    case SubsystemCrossingParamsKey => up(SubsystemCrossingParamsKey, site)
+      .copy(cbusToPbusXType = xType)
+})
+
+/**
+  * Mixins to set the dtsFrequency field of BusParams -- these will percolate it'st way
+  * through the diplomatic clock graph to the clock sources.
+  */
+class WithPeripheryBusFrequency(freq: BigInt) extends Config((site, here, up) => {
+  case PeripheryBusKey => up(PeripheryBusKey).copy(dtsFrequency = Some(freq))
+})
+class WithMemoryBusFrequency(freq: BigInt) extends Config((site, here, up) => {
+  case MemoryBusKey => up(MemoryBusKey).copy(dtsFrequency = Some(freq))
+})
+
+class WithRationalMemoryBusCrossing extends WithMemoryBusCrossingType(RationalCrossing(Symmetric))
+class WithAsynchrousMemoryBusCrossing extends WithMemoryBusCrossingType(AsynchronousCrossing())
--- a/generators/chipyard/src/main/scala/CustomBusTopologies.scala
+++ b/generators/chipyard/src/main/scala/CustomBusTopologies.scala
@@ -16,12 +16,14 @@ import freechips.rocketchip.subsystem._
 /**
   * Keys that serve as a means to define crossing types from a Parameters instance
   */
-case object SubsystemCrossingParamsKey extends Field[SubsystemCrossingParams](SubsystemCrossingParams())
-case object MemoryBusCrossingTypeKey extends Field[ClockCrossingType](NoCrossing)
+case object SbusToMbusXTypeKey extends Field[ClockCrossingType](NoCrossing)
+case object SbusToCbusXTypeKey extends Field[ClockCrossingType](NoCrossing)
+case object CbusToPbusXTypeKey extends Field[ClockCrossingType](SynchronousCrossing())
+case object FbusToSbusXTypeKey extends Field[ClockCrossingType](SynchronousCrossing())
 
 // Biancolin: This, modified from Henry's email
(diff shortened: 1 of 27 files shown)


davidbiancolin:
@hcook If you could take a look at this, that would be dope too.

generators/chipyard/src/main/scala/CustomBusTopologies.scala
```diff
@@ -0,0 +1,100 @@
...
+) extends TLBusWrapperTopology(
+  instantiations = (if (l2.nBanks == 0) Nil else List(
+    (MBUS, mbus),
+    (L2, CoherenceManagerWrapperParams(mbus.blockBytes, mbus.beatBytes, l2.nBanks, L2.name)(l2.coherenceManager)))),
+  connections = if (l2.nBanks == 0) Nil else List(
+    (SBUS, L2,   TLBusWrapperConnection(xType = NoCrossing, driveClockFromMaster = Some(true), nodeBinding = BIND_STAR)()),
+    (L2,  MBUS,  TLBusWrapperConnection.crossTo(
+      xType = sbusToMbusXType,
+      driveClockFromMaster = Some(true),
+      nodeBinding = BIND_QUERY))
```
colinschmidt, line 38: I'm not 100% sure on how this parameter works but I'd like make sure that this results in an MBUS with NBanks inputs.

generators/chipyard/src/main/scala/config/RocketConfigs.scala
```diff
@@ -178,6 +178,8 @@ class DividedClockRocketConfig extends Config(
   new chipyard.config.WithTileFrequency(200.0) ++
   new freechips.rocketchip.subsystem.WithRationalRocketTiles ++   // Add rational crossings between RocketTile and uncore
   new freechips.rocketchip.subsystem.WithNBigCores(1) ++
+  new chipyard.WithMemoryBusFrequency(50 * 1000 * 1000) ++
```
colinschmidt, line 181: The disconnect between this and the tile frequency is annoying. IMO this API should also be in MHz.
Also to clarify this config has tiles: 200, sbus+l2: 100, and mbus: 50?

colinschmidt (commented):
Looks good.
I do agree that it would be good to support independent PBUS and CBUS frequencies+crossings.
I'm not sure what the additional DTS node is but if BBL/linux is still able to read it and do the right thing it should be ok.

Also every time we increase configurability it just makes me wonder why diplomatic clocking can't do even more. In this case it is interesting that diplomacy knows how the buses are clocked but still requires the user to tell it the right kind of crossing. But that is clearly not something we should solve downstream (IMO).

generators/chipyard/src/main/scala/CustomBusTopologies.scala
```diff
@@ -0,0 +1,100 @@
...
+  sbusToMbusXType: ClockCrossingType = NoCrossing
+) extends TLBusWrapperTopology(
+  instantiations = (if (l2.nBanks == 0) Nil else List(
+    (MBUS, mbus),
+    (L2, CoherenceManagerWrapperParams(mbus.blockBytes, mbus.beatBytes, l2.nBanks, L2.name)(l2.coherenceManager)))),
+  connections = if (l2.nBanks == 0) Nil else List(
+    (SBUS, L2,   TLBusWrapperConnection(xType = NoCrossing, driveClockFromMaster = Some(true), nodeBinding = BIND_STAR)()),
+    (L2,  MBUS,  TLBusWrapperConnection.crossTo(
+      xType = sbusToMbusXType,
+      driveClockFromMaster = Some(true),
```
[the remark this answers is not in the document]
davidbiancolin: No, since the slave-side clock is bound to a different diplomatic source in the master based on the crossing type.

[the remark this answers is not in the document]
colinschmidt: So both connections are made in the case where the clocks are different, and we rely on last connect semantics? Or is it somehow explicitly overridden in the diplomacy code.

I trust that its correct and you don't need to change anything but I'm just curious.

generators/chipyard/src/main/scala/CustomBusTopologies.scala
davidbiancolin: This matches what was there before. `BIND_QUERY` is effectively a `slave :=* master` which sounds like it does what you want.

generators/chipyard/src/main/scala/config/RocketConfigs.scala
davidbiancolin: Yeah, it is. The dtsFrequency argument is a `BigInt` in units in Hz, but frequencies used by diplomatic clocks are `Double`s and are in units of MHz. I think my preference would be to rally about the latter.

davidbiancolin: And yes, those are the correct frequencies.

generators/chipyard/src/main/scala/CustomBusTopologies.scala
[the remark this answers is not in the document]
davidbiancolin: My understanding is that there's no reliance on chisel last connect semantics. A single connection between master wrapper and the slave wrapper is generated (because driveClockFromMaster = true). The diplomatic origin of the clock in the master defines what provides the RHS, and that is determined based on the crossing type provided as the first argument in `TLBusWrapperConnection` methods.

Specifically, that happens here: 
https://github.com/chipsalliance/rocket-chip/blob/master/src/main/scala/tilelink/BusWrapper.scala#L181

If we're using an async crossing, the slave bus (MBUS) will show up as its own clock group in the AsyncClockGroups in our Subsystem. Anything else and it shows up as another clock in the master bus's clock group (this is why there are so many "sbus"-prefixed clocks by default -- in a single-clock system everything is eventually coupled to the sbus's clock group).

generators/chipyard/src/main/scala/CustomBusTopologies.scala
```diff
@@ -0,0 +1,100 @@
...
+/**
+  * Keys that serve as a means to define crossing types from a Parameters instance
+  */
+case object SubsystemCrossingParamsKey extends Field[SubsystemCrossingParams](SubsystemCrossingParams())
+case object MemoryBusCrossingTypeKey extends Field[ClockCrossingType](NoCrossing)
+
+// Biancolin: This, modified from Henry's email
+/** Parameterization of a topology containing a banked coherence manager and a bus for attaching memory devices. */
+case class CoherentBusTopologyParams(
+  sbus: SystemBusParams, // TODO remove this after better width propagation
```
jerryz123, line 25: If we have this forked class from RC, can we name it to indicate it as such? `CoherentCrossedBusTopologyParams`?

I'm concerned if we ever have code which does 
```
import freechips.rocketchip.subsystem._
import chipyard._
```

generators/chipyard/src/main/scala/CustomBusTopologies.scala
```diff
@@ -0,0 +1,100 @@
...
+      xType = sbusToMbusXType,
+      driveClockFromMaster = Some(true),
+      nodeBinding = BIND_QUERY))
+  )
+)
+
+// For subsystem/Configs.scala
+
+class WithCoherentBusTopology extends Config((site, here, up) => {
+  case TLNetworkTopologyLocated(InSubsystem) => List(
```
jerryz123, line 45: `WithCoherentMulticlockBusTopology`?

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -108,14 +108,23 @@ class ComposeIOBinder[T, S <: Data](fn: => (T) => (Seq[S], Seq[IOCell]))(implici
...
-    dontTouch(clock_wire)
-    clock_wire := false.B.asClock // necessary for BoringUtils to work properly
-    BoringUtils.bore(source, Seq(clock_wire))
-    clock_io := clock_wire
-    clock_io
+  def apply[T <: Data](name: String, source: T): T = {
+    val (io, wire) = source match {
+      case c: Clock =>
+        val wire =  Wire(Clock())
+        wire := false.B.asClock
```
jerryz123, line 115: Can we preserve the comment indicating why these must have an initial assignment?

generators/chipyard/src/main/scala/CustomBusTopologies.scala
```diff
@@ -0,0 +1,100 @@
...
+
+import freechips.rocketchip.config.{Field, Config, Parameters}
+import freechips.rocketchip.diplomacy._
+import freechips.rocketchip.tilelink._
+import freechips.rocketchip.util.{Location, Symmetric}
+import freechips.rocketchip.subsystem._
+
+// I'm putting this code here temporarily as I think it should be a candidate
+// for upstreaming based on input from Henry Cook, but don't wnat to deal with
+// an RC branch just yet.
```
hcook, line 12: Yup, I would encourage you to upstream this. Though I am thinking it might be best to just bite the bullet of deprecating my old `SubsystemCrossingParams` in favor of the pattern you used here for `sbusToMbusXType` (i.e. make upstream `HierarchicalBusTopologyParams` take three `ClockCrossingType` arguments with no extra case class layer to have to juggle alterations upon.) If you provide the mix-ins too I will get our stuff ported to using them.

generators/chipyard/src/main/scala/CustomBusTopologies.scala
```diff
@@ -0,0 +1,100 @@
...
+// For subsystem/Configs.scala
+
+class WithCoherentBusTopology extends Config((site, here, up) => {
+  case TLNetworkTopologyLocated(InSubsystem) => List(
+    JustOneBusTopologyParams(sbus = site(SystemBusKey)),
+    HierarchicalBusTopologyParams(
+      pbus = site(PeripheryBusKey),
+      fbus = site(FrontBusKey),
+      cbus = site(ControlBusKey),
+      xTypes = SubsystemCrossingParams()),
```
hcook, line 51: you probably want this to be `site(SubsystemCrossingParamsKey)`

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -257,10 +266,11 @@ class WithSerialTLIOCells extends OverrideIOBinder({
 
 class WithAXI4MemPunchthrough extends OverrideIOBinder({
   (system: CanHaveMasterAXI4MemPort) => {
-    val ports: Seq[ClockedIO[AXI4Bundle]] = system.mem_axi4.zipWithIndex.map({ case (m, i) =>
-      val p = IO(new ClockedIO(DataMirror.internal.chiselTypeClone[AXI4Bundle](m))).suggestName(s"axi4_mem_${i}")
+    val ports: Seq[ClockedAndResetIO[AXI4Bundle]] = system.mem_axi4.zipWithIndex.map({ case (m, i) =>
+      val p = IO(new ClockedAndResetIO(DataMirror.internal.chiselTypeClone[AXI4Bundle](m))).suggestName(s"axi4_mem_${i}")
       p.bits <> m
       p.clock := BoreHelper("axi4_mem_clock", system.asInstanceOf[BaseSubsystem].mbus.module.clock)
+      p.reset := BoreHelper("axi4_mem_reset", system.asInstanceOf[BaseSubsystem].mbus.module.reset)
```
hcook, line 273: if you want to future-proof this class against configurable topologies, i recommend making it take `fromWhere: TLBusWrapperLocation = MBUS` as an argument and then doing `system.asInstanceOf[HasTileLinkLocations]. locateTLBusWrapper(fromWhere).module.clock` to find the clock. Lets you punch to something that isn't based on directly on `BaseSubsystem` too 😎

generators/chipyard/src/main/scala/config/RocketConfigs.scala
hcook: My preference is also for the latter

generators/chipyard/src/main/scala/clocking/ResetSynchronizer.scala
```diff
@@ -0,0 +1,30 @@
+
```
hcook, line 1: Upstream plz

generators/chipyard/src/main/scala/IOBinders.scala
davidbiancolin: Hmm, neat.

generators/chipyard/src/main/scala/CustomBusTopologies.scala
davidbiancolin: I'm happy to just enumerate the conventional crossings and give them names in the config mix-ins. 

It would be cool if, in `p` or elsewhere, we had a Map[(Location, Location), ClockCrossingType] or something to that effect.

generators/chipyard/src/main/scala/CustomBusTopologies.scala
davidbiancolin: I've renamed it for now. The scala compiler will throw a warning but you're right to be concerned.

colinschmidt (approved):
LGTM

jerryz123 (approved):

PR #693, opened 2020-10-13
Update smartelf2hex to use MemSiz instead of FileSiz

jerryz123:
elf2hex writes zeros to a segment for which MemSize > FileSize, which adheres to the ELF spec. 
Thus, we should calculate the total size of the file from the MemSize of the last segment, rather than the FileSize.

From `man elf`: 
```
                        The array element specifies a loadable segment,
                        described by p_filesz and p_memsz.  The bytes from
                        the file are mapped to the beginning of the memory
                        segment.  If the segment's memory size p_memsz is
                        larger than the file size p_filesz, the "extra"
                        bytes are defined to hold the value 0 and to follow
                        the segment's initialized area.  The file size may
                        not be larger than the memory size.  Loadable seg‐
                        ment entries in the program header table appear in
                        ascending order, sorted on the p_vaddr member.
```

**Related issue**: 

**Type of change**: bug fix 

**Impact**: software change 

**Release Notes**


colinschmidt (approved):
LGTM

PR #699, opened 2020-10-18
Support diplomatic IOBinders

jerryz123:
**Related issue**: 

- New `OverrideLazyIOBinder` and `ComposeLazyIOBinder` APIs to write IOBinders which modify diplomatic parameters
- Clock/reset boring replaced with ClockSinkNode auto wiring
- Fixes `make TOP=DigitalTop` and adds regression test

**Impact**: software change

**Release Notes**

### DIFF
--- a/.circleci/defaults.sh
+++ b/.circleci/defaults.sh
@@ -47,7 +47,7 @@ LOCAL_FIRESIM_DIR=$LOCAL_CHIPYARD_DIR/sims/firesim/sim
 
 # key value store to get the build groups
 declare -A grouping
-grouping["group-cores"]="chipyard-ariane chipyard-rocket chipyard-hetero chipyard-boom chipyard-sodor"
+grouping["group-cores"]="chipyard-ariane chipyard-rocket chipyard-hetero chipyard-boom chipyard-sodor chipyard-digitaltop"
 grouping["group-peripherals"]="chipyard-dmirocket chipyard-blkdev chipyard-spiflashread chipyard-spiflashwrite chipyard-mmios chipyard-lbwif"
 grouping["group-accels"]="chipyard-nvdla chipyard-sha3 chipyard-hwacha chipyard-gemmini chipyard-streaming-fir chipyard-streaming-passthrough"
 grouping["group-tracegen"]="tracegen tracegen-boom"
@@ -59,6 +59,7 @@ mapping["chipyard-rocket"]=""
 mapping["chipyard-dmirocket"]=" CONFIG=dmiRocketConfig"
 mapping["chipyard-lbwif"]=" CONFIG=LBWIFRocketConfig"
 mapping["chipyard-sha3"]=" CONFIG=Sha3RocketConfig"
+mapping["chipyard-digitaltop"]=" TOP=DigitalTop"
 mapping["chipyard-streaming-fir"]=" CONFIG=StreamingFIRRocketConfig"
 mapping["chipyard-streaming-passthrough"]=" CONFIG=StreamingPassthroughRocketConfig"
 mapping["chipyard-hetero"]=" CONFIG=LargeBoomAndRocketConfig"
--- a/generators/chipyard/src/main/scala/ChipTop.scala
+++ b/generators/chipyard/src/main/scala/ChipTop.scala
@@ -23,12 +23,10 @@ case object BuildSystem extends Field[Parameters => LazyModule]((p: Parameters) 
  * drive clock and reset generation
  */
 
-class ChipTop(implicit p: Parameters) extends LazyModule with HasTestHarnessFunctions {
-  // A publicly accessible list of IO cells (useful for a floorplanning tool, for example)
-  val iocells = ArrayBuffer.empty[IOCell]
-
+class ChipTop(implicit p: Parameters) extends LazyModule
+    with HasTestHarnessFunctions with HasIOBinders {
   // The system module specified by BuildSystem
-  val lazySystem = LazyModule(p(BuildSystem)(p)).suggestName("system")
+  lazy val lazySystem = LazyModule(p(BuildSystem)(p)).suggestName("system")
 
   // The implicitClockSinkNode provides the implicit clock and reset for the System
   val implicitClockSinkNode = ClockSinkNode(Seq(ClockSinkParameters(name = Some("implicit_clock"))))
@@ -44,13 +42,6 @@ class ChipTop(implicit p: Parameters) extends LazyModule with HasTestHarnessFunc
     // These become the implicit clock and reset to the System
     val implicit_clock = implicitClockSinkNode.in.head._1.clock
     val implicit_reset = implicitClockSinkNode.in.head._1.reset
-
-
-    // Note: IOBinders cannot rely on the implicit clock/reset, as this is a LazyRawModuleImp
-    val (_ports, _iocells, _portMap) = ApplyIOBinders(lazySystem, p(IOBinders))
-    // We ignore _ports for now...
-    iocells ++= _iocells
-    portMap ++= _portMap
 
     // Connect the implicit clock/reset, if present
     lazySystem.module match { case l: LazyModuleImp => {
--- a/generators/chipyard/src/main/scala/HarnessBinders.scala
+++ b/generators/chipyard/src/main/scala/HarnessBinders.scala
@@ -1,7 +1,7 @@
 package chipyard.harness
 
 import chisel3._
-import chisel3.experimental.{Analog}
+import chisel3.experimental.{Analog, BaseModule}
 
 import freechips.rocketchip.config.{Field, Config, Parameters}
 import freechips.rocketchip.diplomacy.{LazyModule, LazyModuleImpLike}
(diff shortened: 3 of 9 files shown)


colinschmidt (commented):
Looks reasonable to me. I like that we can still have non-lazy iobinders.

.circleci/defaults.sh
```diff
@@ -47,7 +47,7 @@ LOCAL_FIRESIM_DIR=$LOCAL_CHIPYARD_DIR/sims/firesim/sim
 
 # key value store to get the build groups
 declare -A grouping
-grouping["group-cores"]="chipyard-ariane chipyard-rocket chipyard-hetero chipyard-boom chipyard-sodor"
+grouping["group-cores"]="chipyard-ariane chipyard-rocket chipyard-hetero chipyard-boom chipyard-sodor chipyard-digitaltop"
```
abejgonzalez, line 50: Is this tested? I don't see a corresponding `.circleci` file change?

generators/chipyard/src/main/scala/ChipTop.scala
```diff
@@ -23,12 +23,10 @@ case object BuildSystem extends Field[Parameters => LazyModule]((p: Parameters)
  * drive clock and reset generation
  */
 
-class ChipTop(implicit p: Parameters) extends LazyModule with HasTestHarnessFunctions {
-  // A publicly accessible list of IO cells (useful for a floorplanning tool, for example)
-  val iocells = ArrayBuffer.empty[IOCell]
-
+class ChipTop(implicit p: Parameters) extends LazyModule
```
abejgonzalez, line 26: Can this extend `BindingScope` so that we can use the lazy binders to attach different resources.

.circleci/defaults.sh
jerryz123: I don't want to test if the simulator runs, I only want to test if it compiles. 
Thus the config file does not need to be changed, since every config in group-cores is built.

generators/chipyard/src/main/scala/ChipTop.scala
jerryz123: Can you do this in your FPGA PR?

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -28,6 +28,12 @@ import chipyard.GlobalResetSchemeKey
 
 import scala.reflect.{ClassTag}
 
+object IOBinderTypes {
+  type IOBinderTuple = (Seq[Data], Seq[IOCell])
+  type IOBinderFunction = (Boolean, =>Any) => ModuleValue[IOBinderTuple]
```
abejgonzalez, line 33: Nit: Have a space after the `=>`

generators/chipyard/src/main/scala/ChipTop.scala
abejgonzalez: Sure. I figured this would be innocuous here but I can move it to the other PR.

generators/chipyard/src/main/scala/HarnessBinders.scala
```diff
@@ -31,81 +31,81 @@ case object HarnessBinders extends Field[Map[String, (Any, HasHarnessSignalRefer
...
-  def apply(th: HasHarnessSignalReferences, sys: LazyModule, map: Map[String, (Any, HasHarnessSignalReferences, Seq[Data]) => Seq[Any]], portMap: Map[String, Seq[Data]]) = {
+  def apply(th: HasHarnessSignalReferences, sys: LazyModule, portMap: Map[String, Seq[Data]])(implicit p: Parameters) = {
     val pm = portMap.withDefaultValue(Nil)
-    map.map { case (s, f) => f(sys, th, pm(s)) ++ f(sys.module, th, pm(s)) }
+    p(HarnessBinders).map { case (s, f) => f(sys, th, pm(s)) ++ f(sys.module, th, pm(s)) }
   }
 }
 
-class OverrideHarnessBinder[T, S <: Data](fn: => (T, HasHarnessSignalReferences, Seq[S]) => Seq[Any])(implicit tag: ClassTag[T], ptag: ClassTag[S]) extends Config((site, here, up) => {
+class HarnessBinder[T, S <: HasHarnessSignalReferences, U <: Data](composer: ((T, S, Seq[U]) => Seq[Any]) => (T, S, Seq[U]) => Seq[Any])(implicit tag: ClassTag[T], thtag: ClassTag[S], ptag: ClassTag[U]) extends Config((site, here, up) => {
```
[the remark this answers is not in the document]
jerryz123: Updated with better names

generators/chipyard/src/main/scala/IOBinders.scala
jerryz123: Fixed

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -205,55 +202,55 @@ class WithExtInterruptIOCells extends OverrideIOBinder({
...
+class WithDebugIOCells extends OverrideLazyIOBinder({
+  (system: HasPeripheryDebug) => {
+    implicit val p = GetSystemParameters(system)
+    val tlbus = system.asInstanceOf[BaseSubsystem].locateTLBusWrapper(p(ExportDebug).slaveWhere)
+    val clockSinkNode = system.debugOpt.map(_ => ClockSinkNode(Seq(ClockSinkParameters())))
+    clockSinkNode.map(_ := tlbus.fixedClockNode)
+    def clockBundle = clockSinkNode.get.in.head._1
+
+
+    InModuleBody { system.asInstanceOf[BaseSubsystem].module match { case system: HasPeripheryDebugModuleImp => {
```
abejgonzalez, line 214: Why do you need to cast this as a `BaseSybsystem`? Shouldn't you just be able to access this as `system.module`?

generators/chipyard/src/main/scala/HarnessBinders.scala
```diff
@@ -31,81 +31,82 @@ case object HarnessBinders extends Field[Map[String, (Any, HasHarnessSignalRefer
...
+  (system: CanHavePeripheryBlockDevice, th: BaseModule with HasHarnessSignalReferences, ports: Seq[ClockedIO[BlockDeviceIO]]) => {
     implicit val p: Parameters = GetSystemParameters(system)
     ports.map { b => SimBlockDevice.connect(b.clock, th.harnessReset.asBool, Some(b.bits)) }
     Nil
   }
 })
 
 class WithBlockDeviceModel extends OverrideHarnessBinder({
-  (system: CanHavePeripheryBlockDevice, th: HasHarnessSignalReferences, ports: Seq[ClockedIO[BlockDeviceIO]]) => {
+  (system: CanHavePeripheryBlockDevice, th: BaseModule with HasHarnessSignalReferences, ports: Seq[ClockedIO[BlockDeviceIO]]) => {
```
abejgonzalez, line 101: Why does it matter that we have `BaseModule` instead of just `HasHarness`...?

generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -40,21 +46,30 @@ import scala.reflect.{ClassTag}
...
-
-    val ports: Iterable[Data] = unzipped._1.flatten
-    val cells: Iterable[IOCell] = unzipped._2.flatten
-    val portMap: Map[String, Seq[Data]] = map.keys.map(k => k -> (lzy(k)._1 ++ imp(k)._1)).toMap
-    (ports, cells, portMap)
-  }
+
+abstract trait HasIOBinders { this: LazyModule =>
+  val lazySystem: LazyModule
+  private val iobinders = p(IOBinders)
```
abejgonzalez, line 55: There is no "ApplyIOBinders" function right. So where do the IOBinder functions get called? I'm not exactly sure here.

abejgonzalez (approved):
Other than the questions I have this LGTM. This (from an initial port) works with my FPGA work (can generate Verilog and a proper DTS).

generators/chipyard/src/main/scala/HarnessBinders.scala
jerryz123: Technically `HasHarnessSignalReferences` can be added to something which is not a `BaseModule` (although it probably shouldn't be). This provides the additional safety of checking this is evaluated in the context of a Module

generators/chipyard/src/main/scala/IOBinders.scala
jerryz123: The evaluation of `lzy` and `imp` below evaluates the IOBinders, or wraps them in `InModuleBody` if they are not lazyiobinders. The implementation is a bit hairy because I wanted to avoid having to modify all of the existing concrete iobinders,

lsteveol:
I hope this doesn't sound too pushy, but do you guys have an ETA on this being merged? I would like to try out some of these new features. Thanks!

generators/chipyard/src/main/scala/HarnessBinders.scala
```diff
@@ -31,81 +31,82 @@ case object HarnessBinders extends Field[Map[String, (Any, HasHarnessSignalRefer
...
-        val pts = ports.collect({case p: S => p})
-        require (pts.length == ports.length, s"Port type mismatch between IOBinder and HarnessBinder: ${ptag}")
-        t match {
-          case system: T => up(HarnessBinders, site)(tag.runtimeClass.toString)(system, th, pts) ++ fn(system, th, pts)
-          case _ => Nil
-        }
-      })
-  )
-})
+class OverrideHarnessBinder[T, S <: HasHarnessSignalReferences, U <: Data](fn: => (T, S, Seq[U]) => Seq[Any])
```
davidbiancolin, line 60: I see we're approaching diplomacy-level type genericization.

generators/chipyard/src/main/scala/HarnessBinders.scala
```diff
@@ -31,81 +31,82 @@ case object HarnessBinders extends Field[Map[String, (Any, HasHarnessSignalRefer
   Map[String, (Any, HasHarnessSignalReferences, Seq[Data]) => Seq[Any]]().withDefaultValue((t: Any, th: HasHarnessSignalReferences, d: Seq[Data]) => Nil)
 )
 
-
 object ApplyHarnessBinders {
-  def apply(th: HasHarnessSignalReferences, sys: LazyModule, map: Map[String, (Any, HasHarnessSignalReferences, Seq[Data]) => Seq[Any]], portMap: Map[String, Seq[Data]]) = {
+  def apply(th: HasHarnessSignalReferences, sys: LazyModule, portMap: Map[String, Seq[Data]])(implicit p: Parameters) = {
```
davidbiancolin, line 35: Label the return type.

davidbiancolin (commented):
Cool! It'd be nice to see the boring utils stuff ripped out given we can use diplomacy to plumb out the clock now.

generators/chipyard/src/main/scala/HarnessBinders.scala
```diff
@@ -27,80 +27,82 @@ import icenet.{CanHavePeripheryIceNIC, SimNetwork, NicLoopback, NICKey, NICIOvon
...
+        val pts = ports.collect({case p: U => p})
+        require (pts.length == ports.length, s"Port type mismatch between IOBinder and HarnessBinder: ${portTag}")
+        val upfn = up(HarnessBinders, site)(systemTag.runtimeClass.toString)
+        th match {
+          case th: S =>
+            t match {
+              case system: T => composer(upfn)(system, th, pts)
+              case _ =>
+            }
+          case _ =>
```
timsnyder-siv, line 57: @jerryz123 I'm bumping to start using this code.  Can you help me understand why you need the empty default matching cases on lines 55 and 57 of the new file here (i.e. the `case _ =>` lines)?

jerryz123: I believe this was done to avoid MatchErrors. In the case where the type of the actual elaborated TestHarness does not match the expected Harness type, or in the case where type of the actual elaborated System does not match the expected System type, this expression just becomes a no-op. 
This is useful because it lets us build up a collection of default HarnessBinders through Configs without thinking about exactly what hardware design is being requested.

PR #703, opened 2020-10-26
Make the ChipTop reset pin always async

jerryz123:

**Impact**: rtl change

### DIFF
--- a/generators/chipyard/src/main/scala/Clocks.scala
+++ b/generators/chipyard/src/main/scala/Clocks.scala
@@ -16,40 +16,6 @@ import testchipip.{TLTileResetCtrl}
 import chipyard.clocking._
 
 /**
-  * Chipyard provides three baseline, top-level reset schemes, set using the
-  * [[GlobalResetSchemeKey]] in a Parameters instance. These are:
-  *
-  * 1) Synchronous: The input coming to the chip is synchronous to the provided
-  *    clocks and will be used without modification as a synchronous reset.
-  *    This is safe only for use in FireSim and SW simulation.
-  *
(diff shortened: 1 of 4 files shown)


davidbiancolin (commented):
I'm good with this if Colin is.

colinschmidt (approved):
LGTM

generators/chipyard/src/main/scala/Clocks.scala
```diff
@@ -58,18 +24,12 @@ object GenerateReset {
...
-        val async_reset_wire = Wire(Input(AsyncReset()))
-        reset_wire := ResetCatchAndSync(clock, async_reset_wire.asBool())
-        IOCell.generateIOFromSignal(async_reset_wire, "reset", abstractResetAsAsync = true)
-      }
-    }
+    val async_reset_wire = Wire(Input(AsyncReset()))
+    val (reset_io, resetIOCell) = IOCell.generateIOFromSignal(async_reset_wire, "reset",
+      abstractResetAsAsync = true)
+
+    val reset_wire = ResetCatchAndSync(clock, async_reset_wire.asBool())
```
jerryz123, line 31: @davidbiancolin We don't need this reset synchronizer anymore now that we have the diplomatic synchronizer, right?

colinschmidt: Can you just double check in the generated verilog that the only fanout of this async reset is to the catch and sync?

jerryz123: Yeah. The async reset coming out of the IOCell drives only the the catchandsyncs in the ResetSynchronizer node.

PR #704, opened 2020-10-29
Fix broken example-vlsi path

colinschmidt:
cc @alonamid


alonamid:
I was planning to push this fix as part of my bigger docs PR with a couple more example conf files, but we can also do this quick fix for now

colinschmidt:
Ah ok.

PR #705, opened 2020-11-01
Fix env.sh zsh compatibility

davidbiancolin:
Fix a regression introduced by d402825e that breaks env.sh compatibility for zsh. The RHS of this expression provides the zsh-approximation of `BASH_SOURCE` (which is not defined).

**Related issue**:  #435 

**Type of change**: bug fix

**Impact**: software change


abejgonzalez (approved):
LGTM

PR #707, opened 2020-11-03
Update Reference Frequency Selection for Divider-Only Clock Generator

davidbiancolin:
This modifies how the reference clock is selected for the divider-only clock generator and FireSim Rational clock bridges, to produce a good result when no exact solution exists. Without introducing too much complexity, this algorithm does a brute force search over multiples of the fastest requested clock and picks the smallest that satisfies an RMS error bound. 

There a few constants here that i've mostly pulled out of thin air. The maximumAllowableFrequency is larger than you'd expect to be able to generate with typical VFOs (afaik) but was picked to work with FireSim's high tile frequencies.  

- [ ] Add scalatest to CI?

**Related issue**: #659 

**Type of change**: bug fix

**Impact**: rtl change

**Release Notes**


colinschmidt (approved):
LGTM.

I think the high max frequency is fine since you can control the maxFreq. You can then either configure it lower and accept the error or create two separate PLL configs.

jerryz123 (approved):

PR #709, opened 2020-11-04
Bump Dromajo | Optional ignore QEMU toolchain flag

abejgonzalez:
**Related issue**: 

**Type of change**: new feature + bug fix

**Impact**: software change

**Release Notes**
Small changes to allow Dromajo to be built with an older version of `glibc` and adding an "ignore QEMU" flag to the build toolchains script (default is to still build QEMU).

### DIFF
--- a/scripts/build-toolchains.sh
+++ b/scripts/build-toolchains.sh
@@ -107,14 +107,18 @@ if [ "${EC2FASTINSTALL}" = true ] ; then
     git submodule deinit "${module}" || :
 
 else
-    "${MAKE}" --version | (
-        read -r makever
-        case ${makever} in
-        'GNU Make '[4-9]\.*|'GNU Make '[1-9][0-9]) ;;
-        *) false ;;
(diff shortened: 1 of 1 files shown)


scripts/build-toolchains.sh
```diff
@@ -34,6 +35,7 @@ die() {
 
 TOOLCHAIN="riscv-tools"
 EC2FASTINSTALL="false"
+IGNOREQEMU="false"
```
jerryz123, line 38: Nit: avoid the negation, and just have a `INSTALLQEMU` variable.

abejgonzalez: I'm following what the `EC2FASTINSTALL` did... let me double-check that there wasn't a specific reason to do that. I'm not opposed to just checking if the variable exists (assuming it doesn't break things).

abejgonzalez: Hmm. This looks like this was just coding preference. Ill change it.

alonamid (approved):
LGTM

Issue #708, opened 2020-11-04
Freq Summary has some incorrect frequencies


lsteveol:
**Impact**: software

**Tell us about your environment:**
*Chipyard Version:* branch: `dev`  hash: 37415157d61ec9ccdaeb11da75028afdf990461e
*OS:* 
*Other:* 
First, mad gratitude for the separate clock domains!

**What is the current behavior?**
I have the current custom Config (I have removed some non-related pieces)
``` scala
class WH440RocketConfig extends Config(

  // Clocking
  
  new chipyard.config.WithTileFrequency(1500.0) ++                              //CPUs at 1.5GHz
  new freechips.rocketchip.subsystem.WithRationalRocketTiles ++
  
  // CPUs
  
  new freechips.rocketchip.subsystem.WithNBanks(4) ++
  new freechips.rocketchip.subsystem.WithNSmallCores(1, Some(0)) ++             //Force to HART0
  new freechips.rocketchip.subsystem.WithNBigCores(4, Some(1)) ++
    
  new freechips.rocketchip.subsystem.WithNExtTopInterrupts(64) ++
  
  // Memory Ports
  new chipyard.config.WithMemoryBusFrequency(750.0) ++
  new chipyard.config.WithRationalMemoryBusCrossing ++                          //double check this
  new freechips.rocketchip.subsystem.WithNMemoryChannels(2) ++                  //
  
  // Other Busses
  // SBUS Connections
  new chipyard.config.WithSystemBusFrequency(750.0) ++
  new chipyard.config.WithSbusToCbusCrossingType(AsynchronousCrossing()) ++
  
    
  // PBUS Connections
  new chipyard.config.WithControlBusFrequency(100.0) ++
  new chipyard.config.WithPeripheryBusFrequency(100.0) ++
  //new chipyard.config.WithCbusToPbusCrossingType(SynchronousCrossing()) ++      //Why are these connected like this?
  new chipyard.config.WithCbusToPbusCrossingType(AsynchronousCrossing()) ++
  
  // FBUS Connections
  new chipyard.config.WithFbusToSbusCrossingType(SynchronousCrossing()) ++      //FBUS is same as SBUS
  //removed

  new chipyard.WithMulticlockCoherentBusTopology ++                             //Here for now until upstreamed to RocketChip

  new MyAbstractConfig)
```

[...]

PR #713, opened 2020-11-08
Better Bus Frequency Specification

davidbiancolin:
We're still working through how we want to specify frequencies in multiclock systems -- H/T @lsteveol for shedding light on some of the current limitations. 

This PR modifies how bus clock groups are connected to better preserve the clock names presented to the top-level. This is critical since, for the foreseeable future, we are going to be relying on string matching to provide non-default frequencies. Under this change, all 5 canonical TL buses will have clocks whose names reflect the name of the bus, independent of the choice of crossing type. Here's an example:

```
dividerOnlyClockGenerator Frequency Summary
  Input Reference Frequency: 8000.0 MHz                                                        
  Output clock implicit_clock, requested: 800.0 MHz, actual: 800.0 MHz (division of 10)
  Output clock subsystem_sbus_0, requested: 800.0 MHz, actual: 800.0 MHz (division of 10)
  Output clock subsystem_sbus_core_0, requested: 1600.0 MHz, actual: 1600.0 MHz (division of 5)                                                            
  Output clock subsystem_sbus_2, requested: 800.0 MHz, actual: 800.0 MHz (division of 10) (This is the L2)
  Output clock subsystem_pbus_0, requested: 100.0 MHz, actual: 100.0 MHz (division of 80)
  Output clock subsystem_fbus_0, requested: 800.0 MHz, actual: 800.0 MHz (division of 10)
  Output clock subsystem_mbus_0, requested: 1000.0 MHz, actual: 1000.0 MHz (division of 8)
  Output clock subsystem_cbus_0, requested: 800.0 MHz, actual: 800.0 MHz (division of 10) 
```

`WithCommuteBusFrequencyAssignments` ensures that synchronous clock groups attached to a bus receive the same frequency their bus, instead of the global default. Practically, this means the L2 will always receive the sbus frequency, if it's set (and so long as the L2 continues to not name its clock); this addresses one primary issue outlined in #708.

There are still a few outstanding problems that will need to be addressed in future PRs that make the current multiclock handling unintuitive.
- use of the implicit clock in subsystem mixins (notably the bootrom, as discussed in #708). I'm not sure how many instances of this remain, but presently the CBUS frequency must match the implicit clock frequency for things to work correctly.
  - Perhaps we should assert this is the case?
- devices with multiple attachments assume the attachment points are co-resident in a clock domain (e.g., block device) 
  - Fix in an upcoming FireSim-related PR
- Relying on a global fallback frequency is dangerous
  - We should consider croaking instead and encourage the use of good names in clock sinks.

**Related issue**: #708 

**Type of change**: bug fix 

**Impact**: rtl change

### DIFF
--- a/generators/chipyard/src/main/scala/ConfigFragments.scala
+++ b/generators/chipyard/src/main/scala/ConfigFragments.scala
@@ -242,16 +242,19 @@ class WithFbusToSbusCrossingType(xType: ClockCrossingType) extends Config((site,
   * up the diplomatic graph to the clock sources.
   */
 class WithPeripheryBusFrequency(freqMHz: Double) extends Config((site, here, up) => {
-  case PeripheryBusKey => up(PeripheryBusKey).copy(dtsFrequency = Some(BigInt((freqMHz * 1e6).toLong)))
+  case PeripheryBusKey => up(PeripheryBusKey, site).copy(dtsFrequency = Some(BigInt((freqMHz * 1e6).toLong)))
 })
 class WithMemoryBusFrequency(freqMHz: Double) extends Config((site, here, up) => {
-  case MemoryBusKey => up(MemoryBusKey).copy(dtsFrequency = Some(BigInt((freqMHz * 1e6).toLong)))
+  case MemoryBusKey => up(MemoryBusKey, site).copy(dtsFrequency = Some(BigInt((freqMHz * 1e6).toLong)))
 })
 class WithSystemBusFrequency(freqMHz: Double) extends Config((site, here, up) => {
-  case SystemBusKey => up(SystemBusKey).copy(dtsFrequency = Some(BigInt((freqMHz * 1e6).toLong)))
+  case SystemBusKey => up(SystemBusKey, site).copy(dtsFrequency = Some(BigInt((freqMHz * 1e6).toLong)))
+})
+class WithFrontBusFrequency(freqMHz: Double) extends Config((site, here, up) => {
+  case FrontBusKey => up(FrontBusKey, site).copy(dtsFrequency = Some(BigInt((freqMHz * 1e6).toLong)))
 })
 class WithControlBusFrequency(freqMHz: Double) extends Config((site, here, up) => {
-  case ControlBusKey => up(ControlBusKey).copy(dtsFrequency = Some(BigInt((freqMHz * 1e6).toLong)))
+  case ControlBusKey => up(ControlBusKey, site).copy(dtsFrequency = Some(BigInt((freqMHz * 1e6).toLong)))
 })
 
 class WithRationalMemoryBusCrossing extends WithSbusToMbusCrossingType(RationalCrossing(Symmetric))
--- a/generators/chipyard/src/main/scala/ConfigFragments.scala
+++ b/generators/chipyard/src/main/scala/ConfigFragments.scala
@@ -1,5 +1,6 @@
 package chipyard.config
 
+import scala.util.matching.Regex
 import chisel3._
 import chisel3.util.{log2Up}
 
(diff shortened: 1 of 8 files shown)


generators/chipyard/src/main/scala/ConfigFragments.scala
```diff
@@ -183,6 +185,34 @@ class WithPeripheryBusFrequencyAsDefault extends Config((site, here, up) => {
...
+  * Provides a diplomatic frequency for all clock sinks with an unspecified
+  * frequency bound to each bus.
+  *
+  * For example, the L2 cache, when bound to the sbus, receives a separate
+  * clock that appears as "subsystem_sbus_<num>".  This fragment ensures that
+  * clock requests the same frequency as the sbus itself.
+  */
+
+class WithCommuteBusFrequencyAssignments extends Config(
+  new BusFrequencyAssignment("subsystem_sbus_\\d+".r, SystemBusKey) ++
```
jerryz123, line 209: Commute?

generators/chipyard/src/main/scala/CustomBusTopologies.scala
```diff
@@ -36,17 +36,35 @@ case class CoherentMulticlockBusTopologyParams(
...
-      driveClockFromMaster = Some(true),
+      driveClockFromMaster = None,
       nodeBinding = BIND_QUERY))
   )
 )
 
+// This differs from upstream only in that it does not use the legacy crossTo
+// and crossFrom functions, and it ensures driveClockFromMaster = None
+case class HierarchicalMulticlockBusTopologyParams(
+  pbus: PeripheryBusParams,
```
jerryz123, line 47: We should try to upstream these changes too, right?

generators/chipyard/src/main/scala/Subsystem.scala
```diff
@@ -56,6 +56,23 @@ class ChipyardSubsystem(implicit p: Parameters) extends BaseSubsystem
...
+  // names are always preserved in the top-level clock names.
+  //
+  // For example, using a RationalCrossing between the Sbus and Cbus, and
+  // driveClockFromMaster = Some(true) results in all cbus-attached device and
+  // bus clocks to be given names of the form "subsystem_sbus_[0-9]*".
+  // Conversly, if an async crossing is used, they instead receive names of the
+  // form "subsystem_cbus_[0-9]*". The assignment below the latter names in all cases.
+  Seq(PBUS, FBUS, MBUS, CBUS).foreach { loc =>
+    tlBusWrapperLocationMap.lift(loc).foreach { _.clockGroupNode := asyncClockGroupsNode }
+  }
```
jerryz123, line 75: Do you think changing this behavior could ever backfire on us? It seems safe to me.

lsteveol:
Sweet, thanks for the improvements. I will try them out once they have been merged and I have some time.

generators/chipyard/src/main/scala/Subsystem.scala
davidbiancolin: I don't think so. The regex matching is the gnarliest part of this PR i think.

generators/chipyard/src/main/scala/CustomBusTopologies.scala
davidbiancolin: I will try. The driveClockFromMaster part will need to be a on switch i think

generators/chipyard/src/main/scala/ConfigFragments.scala
davidbiancolin: I dunno the right verb for what this is doing... Share? Passon?

alonamid: Inherit?

colinschmidt: AdoptBusFrequencyAssignments?

jerryz123 (approved):

lsteveol:
Hey @davidbiancolin I checked this out and it appears that the Asynchronous crossings aren't being properly inserted into the couplers between the busses. In the `MultiClockRocketConfig` you have `new chipyard.config.WithCbusToPbusCrossingType(AsynchronousCrossing()) ++` but this is what I see in the graphml.

image

Here is a snapshot from one of my older versions. And you can see the `asource/asink`.

image

I've done a little debug, but don't see anything at the moment.

davidbiancolin:
Ah, shit. Ok, if you want to fix this you should change the `TLBusWrapperConnection`s in `HierarchicalBusTopologyParams` to use `crossTo` and `crossFrom` again. That provides the views that actually insert the crossings. I'll PR a fix momentarily.

I need to look in to why the test passed.

lsteveol:
Initial testing is looking good for that change.

PR #718, opened 2020-11-15
Generalizing VLSI Flows docs

alonamid:
**Related issue**: 

**Impact**: other

**Release Notes**

The goal of this PR is to generalize the VLSI flow docs to be applicable to flows other than just ASAP7 (in contrast to the tutorial), including various nuances that may seem trivial to frequent users, but less trivial to occasional user.

Also includes the occasional makefile bug fix.

It's still a WIP, but open to comments


docs/VLSI/Basic-Flow.rst
```diff
@@ -0,0 +1,217 @@
...
+
+Initialize the Hammer Plug-ins
+----------------------------------
+In the Chipyard root, run:
+
+.. code-block:: shell
+
+    ./scripts/init-vlsi.sh <tech-plugin-name>
+    
+This will pull the Hammer & CAD tool plugin submodules, assuming the technology plugins are available on github.
```
harrisonliew, line 16: The tech plugin doesn't have to be in github. If it's not one of the default plugins in Hammer (asap7, nangate45 <- you need to add this to the vlsi Makefile), then it needs to be added as a submodule in the `vlsi/` folder first before you run the `init-vlsi.sh` script.

harrisonliew: Also I'm a bit confused about where this fits into the existing docs.

vlsi/example-tools.yml
```diff
@@ -0,0 +1,24 @@
...
+# Genus options
+vlsi.core.synthesis_tool: "genus"
+vlsi.core.synthesis_tool_path: ["hammer-cadence-plugins/synthesis"]
+vlsi.core.synthesis_tool_path_meta: "append"
+synthesis.genus.version: "1813"
+# Innovus options
+vlsi.core.par_tool: "innovus"
+vlsi.core.par_tool_path: ["hammer-cadence-plugins/par"]
+vlsi.core.par_tool_path_meta: "append"
+par.innovus.version: "191_ISR3"
```
harrisonliew, line 17: Is this used in the asap7 demo? If so it needs to be 181.

vlsi/example-tools.yml
```diff
@@ -0,0 +1,24 @@
+# SRAM Compiler compiler options
+vlsi.core.sram_generator_tool: "sram_compiler"
+# You should specify a location for the SRAM generator in the tech plugin
+vlsi.core.sram_generator_tool_path: []
```
harrisonliew, line 4: If this is used in the asap7 demo, a more explicit (but technically redundant, since hammer knows about it already) path would be:
`vlsi.core.sram_generator_tool_path: ["/home/ff/eecs151/hammer/src/hammer-vlsi/technology/asap7"]`

I don't know if you want to give this option for users who don't desire to use barstools' macrocompiler:
```
# Uncomment these if you are not using macrocompiler
# vlsi.inputs.sram_parameters: "/home/ff/eecs151/hammer/src/hammer-vlsi/technology/asap7/sram-cache.json"
# vlsi.inputs.sram_parameters_meta: ["transclude", "json2list"] 
```

vlsi/example-tech.yml
```diff
@@ -0,0 +1,10 @@
+# Technology Setup
```
jerryz123, line 1: Are we sure this needs to be specified here? Don't the hammer-tech default.ymls supply these?

docs/VLSI/Basic-Flow.rst
alonamid: I'm not sure I understand this comment. I know that the tech plugin doesn't have to be in github. That's the reason the following lines explain how to clone a private tech plug-in.

Why does the submodule need to be added before `init-vlsi.sh`. The script only initializes the submodules (rescursively). If you're cloning the submodule independently it will already be initialized, wouldn't it? And AFAIK, tech-plugins don't include submodules within them

vlsi/example-tools.yml
alonamid: The intention of this file is to be used as template for people using Hammer.
I did not want to start testing the ASAP7 demo again, and therefore I did not re-factor the ASAP7 tutorial yml (example-asap7.yml) and it is left as it previously was in it's monolithic form.

vlsi/example-tools.yml
alonamid: As previously noted, it is meant to be used as a template for a general case, rather than solving a single-case problem. 
Given that assumption, I think any path that includes the Berkeley instructional machines eecs151 installation directory would be confusing.

vlsi/example-tech.yml
alonamid: I'm going to defer this question to @colinschmidt @harrisonliew

docs/VLSI/Basic-Flow.rst
harrisonliew: Don't you need to register the submodule by doing a `git submodule add` first?

vlsi/example-tools.yml
harrisonliew: Oh sorry, I forgot to take out `/home/ff/eecs151/`! Edited above.

vlsi/example-tech.yml
harrisonliew: Generally, the plugin will supply `vlsi.core.node` only. However, the rest of the keys before this imo should be explicit (otherwise, how do you know which tech you're even using?). `technology.<tech_name>.install_dir` is also end environment-specific, and in the case of ASAP7, comes as a tarball instead of a directory.

vlsi/example-tools.yml
alonamid: Why would I want to specify the explicit path if Hammer is already aware if it through the tech plugin?

vlsi/example-tech.yml
alonamid: I agree that specifying the tech you're using, the path to the plugin and the path to the PDK is important. However, do we really need `vlsi.core.technology_path_meta` and `vlsi.core.node` to be manually specified rather than just using default values from the tech plugin? 
Default values are a good thing!

dpgrubb13: I don't think the `vlsi.core.node` needs to be specified here since it should be specified in the tech plugin.

vlsi/example-tools.yml
harrisonliew: Because it needs to be specified for tech plugins that are separate from main Hammer. Since this is `example-tools.yml` you could just add these fields with filler text instead of ASAP7.

docs/VLSI/Basic-Flow.rst
```diff
@@ -0,0 +1,217 @@
...
+
+Hence, if we want to monolitically place-and-route the entire SoC with the default tech plug-in parameters for power-straps and corners, the relevant command would be
+.. code-block:: shell
+
+    make par CONFIG=<chipyard_config_name> tech_name=<tech_name> INPUT_CONFS="example-design.yml example-tools.yml example-tech.yml"
+
+In a more typical scenario of working on a single module, for example the Gemmini accelerator within the GemminiRocketConfig Chipyard SoC configuration,
+
+.. code-block:: shell
+  vlsi.inputs.placement_constraints:
```
colinschmidt, line 119: I think you should add a sentence about how the width and height here could vary a lot based on the block you are building (is 300x300 even right for gemmini? I thought it was bigger).

Also this is a hammer comment but I think we might be able to support a different mode for this that is even more automatic. But that would need some manual reading, so its probably best just to be noted as an issue (in hammer).

docs/VLSI/Basic-Flow.rst
```diff
@@ -0,0 +1,217 @@
...
+        bottom: 0
+
+The relevant ``make`` command would then be
+.. code-block:: shell
+
+    make par CONFIG=GemminiRocketConfig VLSI_TOP=Gemmini tech_name=tsmintel3 INPUT_CONFS="example-design.yml example-tools.yml example-tech.yml"
+
+
+Place-and-route generally requires more fine-grained input specifications regarding power nets, clock nets, pin assignments and floorplanning. While the template configuration files provide defaults for automatic tool defaults, these will usually result in very bad QoR, and therefore it is recommended to specify better-informed floorplans, pin assignments and power nets. For more information about cutomizing theses parameters, please refer to the :ref:`Customizing Your VLSI Flow in Hammer` sections or to the Hammer documentation. 
+Additionally, some Hammer process technology plugins do not provide sufficient default values for requires settings such as power nets and pin assignments (for example, ASAP7). In those cases, these contraint will need to be specified manually in the top-level configuration yml files, as is the case in the ``example-asap7.yml`` configuration file.
```
colinschmidt, line 139:
```suggestion
Additionally, some Hammer process technology plugins do not provide sufficient default values for requires settings such as power nets and pin assignments (for example, ASAP7). In those cases, these constraints will need to be specified manually in the top-level configuration yml files, as is the case in the ``example-asap7.yml`` configuration file.
```

docs/VLSI/Basic-Flow.rst
```diff
@@ -0,0 +1,217 @@
...
+
+Place-and-route tools are very sensitive to process technologes (significantly more sensitive than synthesis tools), and different process technologies may work only on specific tool versions. It is recommended to check what is the appropriate tool version for the specific process technology you are working with.
+
+
+.. Note:: If you edit the yml configuration files in between synthesis and place-and-route, the `make par` command will automatically re-run synthesis. If you would like to avoid that and are confident that your configuration file changes do not affect synthesis results, you may use the `make redo-par` instead.
```
colinschmidt, line 144: I think this note should point to documentation on the hammer chipyard makefile integration (which I don't think exists), but would describe the ability to set some fancy makefile variables. Setting `INPUT_CONFS` makes all the yamls requirements for all actions, but the fancy variables let you set which yamls are necessary for which actions useful for things like changing floorplans or the like that only affect PnR.

docs/VLSI/Basic-Flow.rst
```diff
@@ -0,0 +1,217 @@
...
+Power Estimation
+^^^^^^^^^^^^^^^^^^^^
+
+
+
+Signoff
+^^^^^^^^^
+
+During chip tapeout, you will need to perform sign-off check to make sure the generated GDSII can be fabricated as intended. This is done using dedicated signoff tools that perform design rule checking (DRC) and layout versus schematic (LVS) verification. 
+In most cases, placed-and-routed designs will not pass DRC and LVS on first attempts due to nuanced design rules and silent failures of the place-and-route tools. Passing DRC and LVS will often requires adding manual placement constraints to "force" the EDA tools into certain patterns.
```
colinschmidt, line 157: The PnR failures aren't silent usually(it reports many DRCs during routing in my experience). It's just that it wasn't able to fully solve the problem. Perhaps just call them "subtle failures"?

docs/VLSI/Basic-Flow.rst
```diff
@@ -0,0 +1,217 @@
...
+^^^^^^^^^^^^^^^^^^^^
+
+
+
+Signoff
+^^^^^^^^^
+
+During chip tapeout, you will need to perform sign-off check to make sure the generated GDSII can be fabricated as intended. This is done using dedicated signoff tools that perform design rule checking (DRC) and layout versus schematic (LVS) verification. 
+In most cases, placed-and-routed designs will not pass DRC and LVS on first attempts due to nuanced design rules and silent failures of the place-and-route tools. Passing DRC and LVS will often requires adding manual placement constraints to "force" the EDA tools into certain patterns. 
+If you have placed-and-routed a design with the goal of getting area and power estimates, DRC and LVS are not strictly neccessary and the results will likely be quite similar. If you are intending to tapeout and fabricate a chip, DRC and LVS are mandatory and will likely requires multiple-iterations of refining manual placement constraints.
```
colinschmidt, line 158: I think we should mention two things here:
1) Having lots of DRC/LVS violations can be an indication of an unrealistic design and fixing it would significantly change area/power.
2) Having lots of DRC/LVS violations can slow down the PnR tool as it will try to fix each one several times. In my experience this can save several hours of runtime when moving from 10s of thousands of violations to a few hundred.

docs/VLSI/Basic-Flow.rst
```diff
@@ -0,0 +1,217 @@
...
+
+
+Since signoff checks are required only for a complete chip tapeout, they are currently not fully automated in Hammer, and often require some additional manual inclusion of custom Makefiles associated with specific process technologies. However, the general steps from running signoff within Hammer (under the assumption of a fully automated tech plug-in) are Make commands similar to the previous steps.
```
colinschmidt, line 161: Can you flip these two sentences? 
```
Generally with a mature tech plugin its like above. However a new tech plugin generally requires more work to get running DRC than PnR and vastly more than synth.
```

docs/VLSI/Basic-Flow.rst
```diff
@@ -0,0 +1,217 @@
...
+
+In order to run DRC, the relevant ``make`` command is ``make drc``. As in the previous stages, the make command should be accompanied by the relevant configuration Make variables:
+
+.. code-block:: shell
+
+    make drc CONFIG=GemminiRocketConfig VLSI_TOP=Gemmini tech_name=tsmintel3 INPUT_CONFS="example-design.yml example-tools.yml example-tech.yml"
+
+
+DRC does not emit autmated reports, but rather scripts that enable opening the DRC error database within the appropriate tool. These generated scripts can be called from ``./build/<config-name>/drc-rundir/generated-scripts/view_drc``.
```
colinschmidt, line 170:
```suggestion
DRC does not emit easily audited reports, as the rule names violated can be quite esoteric. It is often more productive to use the rather scripts generated by hammer to open the DRC error database within the appropriate tool. These generated scripts can be called from ``./build/<config-name>/drc-rundir/generated-scripts/view_drc``.
```

docs/VLSI/Basic-Flow.rst
```diff
@@ -0,0 +1,217 @@
...
+
+
+In order to run LVS, the relevant ``make`` command is ``make lvs``. As in the previous stages, the make command should be accompanied by the relevant configuration Make variables:
+ 
+.. code-block:: shell
+
+    make lvs CONFIG=GemminiRocketConfig VLSI_TOP=Gemmini tech_name=tsmintel3 INPUT_CONFS="example-design.yml example-tools.yml example-tech.yml"
+
+LVS does not emit autmated reports, but rather scripts that enable opening the LVS error database within the appropriate tool. These generated scripts can be called from ``./build/<config-name>/lvs-rundir/generated-scripts/view_lvs``.
```
colinschmidt, line 179:
```suggestion
LVS does not emit easily audited reports, as the violations are often cryptic when seen textually. As a result it is often more productive to visually see the LVS issues using the generated scripts that enable opening the LVS error database within the appropriate tool. These generated scripts can be called from ``./build/<config-name>/lvs-rundir/generated-scripts/view_lvs``.
```

docs/VLSI/Basic-Flow.rst
```diff
@@ -0,0 +1,217 @@
...
+
+
+Composing a Hierarchical Design
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+
+
+Customizing Generated Tcl Scripts
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+The ``example-vlsi`` python script is the Hammer entry script with placeholders for hooks. Hooks are additional snippets of python and TCL (via ``x.append()``) to extend the Hammer APIs. Hooks can be inserted using the ``make_pre/post/replacement_hook`` methods as shown in the ``example-vlsi`` entry script example. In this particular example, a list of hooks is paased in the ``get_extra_par_hooks`` function in the ``ExampleDriver`` class. Refer to the Hammer documentation on hooks for a detailed description of how these are injected into the VLSI flow.
```
colinschmidt, line 217: Link to hammer docs on this topic?

vlsi/example-design.yml
```diff
@@ -0,0 +1,38 @@
...
+# Default pin layer assignments can be found in some tech plug-ins
+vlsi.inputs.pin_mode: generated
+vlsi.inputs.pin.generate_mode: semi_auto
+
+# Specify the floorplan
+# Default floor plan can be set by the tool
+# The path name should match the VLSI_TOP makefile parameter if it is set
+par.innovus.floorplan_mode: "auto"
+vlsi.inputs.placement_constraints:
+#  - path: "ChipTop"
```
colinschmidt, line 27: I think we should leave the default as ChipTop.

vlsi/example-tech.yml
colinschmidt: Re the path_meta variable. Yes we need that so that hammer knows how to handle conflicting assignments of the tech_path key. With append it will add the path in this file to a list of paths to look for tech files.

colinschmidt (commented):
Looks promising.

docs/VLSI/Basic-Flow.rst
```diff
@@ -0,0 +1,217 @@
...
+
+We call the ``make buildfile`` command while also specifying the name of the process technology we are working with (same ``tech_name`` for the configuration files and plugin name) and the configuration files we created. Note, in the ASAP7 tutorial ((:ref:`tutorial`)) these configuration files are merged into a single file called ``example-asap7.yml``.
+
+Hence, if we want to monolitically place and route the entire SoC, the relevant command would be
+.. code-block:: shell
+
+    make buildfile CONFIG=<chipyard_config_name> tech_name=<tech_name> INPUT_CONFS="example-design.yml example-tools.yml example-tech.yml"
+
+In a more typical scenario of working on a single module, for example the Gemmini accelerator within the GemminiRocketConfig Chipyard SoC configuration, the relevant command would be
```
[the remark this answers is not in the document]
alonamid: I have a whole section further down I want to write about hierarchical support. But I don't know what to write in it....

docs/VLSI/Basic-Flow.rst
alonamid: But that page doesn't exist

colinschmidt: It seems I thought right. I can probably draft something with or for you offline.

docs/VLSI/Basic-Flow.rst
```diff
@@ -218,14 +218,89 @@ If you have access to a shared LSF cluster and you would like Hammer to submit i
...
+While hierarchical physical design can be performed in multiple ways (top-down, bottom-up, abutment etc.), Hammer currently supports only the bottom-up approach.
+The bottom-up approach traverses a tree representing the hierarchy starting from the leaves and towards the direction of the root (the "top level"), and runs the physical design flow on each node of the hierarchy tree using the previously layed-out children nodes.
+As nodes get closer to the root (or "top level") of the hierarchy, largers sections of the design get layed-out.
+
+The Hammer hierarchical flow relies on a manually-specified descrition of the desired heirarchy tree. The specification of the heirarchy tree is defined based on the instance names in the generated Verilog, which sometime make this specification challenging due to inconsisent instance names. Additionally, the specification of the heirarchy tree is intertwined with the manual specification of a floorplan for the design.
```
colinschmidt, line 231: description

colinschmidt (approved):
One typo in the new text but I think as you said its better than nothing and we can iterate after the release.

docs/VLSI/Basic-Flow.rst
jerryz123: Don't you need to add stuff to `HAMMER_EXTRA_ARGS` when running `redo-par` to use your updated ymls? Otherwise Hammer will just read cached results from `par-input.json` without regenerating them.

jerryz123 (approved):
A much-needed addition to our documentation. Thanks @alonamid

PR #717, opened 2020-11-11
Bump RC, firesim and barstools for chisel3.4 updates

timsnyder-siv:
Note: firesim and barstools point to commits in the sifive forks of those repos
I didn't update the URL in .gitmodules because I'm not sure how that works in a PR
(because you wouldn't really want to merge sync'ing to the sifive repo).

Requires: ucb-bar/barstools#87 and firesim/firesim#658

The version of rocket-chip, chisel3 and firrtl is chosen here because it is
the latest known to pass my tests.  You will likely want to bump further.

**Related issue**: 

**Type of change**: other enhancement

**Impact**: rtl change | software change | unknown | other

**Release Notes**

PR #719, opened 2020-11-16
Chisel 3.4, FIRRTL 1.4 and Rocket Chip November Bump

abejgonzalez:
**Related PR**: #717 , https://github.com/ucb-bar/barstools/pull/87, https://github.com/firesim/firesim/pull/658 https://github.com/ucb-bar/cva6-wrapper/pull/10 https://github.com/ucb-bar/gemmini/pull/37 https://github.com/ucb-bar/hwacha/pull/23 https://github.com/ucb-bar/riscv-sodor/pull/59 https://github.com/ucb-bar/sha3/pull/27 https://github.com/ucb-bar/dsptools/pull/217 https://github.com/ucb-bar/testchipip/pull/107 https://github.com/ucb-bar/barstools/pull/87

**Impact**: all RTL generation

**Release Notes**
Thanks @timsnyder-siv for helping out with this and the FireSim side of this.

This is a local version of #717 that bumps multiple repos to use Chisel 3.4 / FIRRTL 1.4 / RC Nov.

**Changes**
- Switch `JAVA_ARGS` to `JAVA_OPTS` (done so that people could use the `sbt` bash script instead of just the launcher and it still works with Chipyard)
- Split `SBT_OPTS` from `JAVA_OPTS` (clearer distinction between the two and also supports the `sbt` bash script)
- Modifications to CI for `*_OPTS` env. vars
- Bump virtually every sub-module
- Deprecate `bloop` support
- Support SBT thin client with `export ENABLE_SBT_THIN_CLIENT=1`
- `build.sbt` refactoring including RC changes
- Helper `make` targets to launch console (`sbt`) and shutdown/start thin server (`<start/shutdown>-sbt-server`)
- Bump to SBT 1.4.4 and bump plugins
- Use `; X; Y; Z;` syntax to run multiple SBT commands


abejgonzalez:
I'm not exactly sure how FIRRTL is built since the JAR isn't built 1st... However, these changes seem to make it decently far into the build until `chisel-testers` fails.

timsnyder-siv:
> I'm not exactly sure how FIRRTL is built *yet since the JAR isn't built 1st... However, these changes seem to make it decently far into the build until `chisel-testers` fails.

Looks like you've started figuring this out because I see [sbt-sriracha](https://github.com/sbt/sbt-sriracha) added to the `project/plugins.sbt`.  However, I don't see you using the `.sourceDependency()` method that sbt-sriracha adds to sbtProject.  See https://github.com/sbt/sbt-sriracha#usage 

Let me know if you'd like some help getting the sbt-sriracha stuff ironed out.  I'd be happy to submit another PR to merge into your local-chisel34 branch.  I talked to @davidbiancolin about this yesterday afternoon.  Wasn't sure if you guys would want to switch over to using it and didn't include it in what I had in #717

abejgonzalez:
> > I'm not exactly sure how FIRRTL is built *yet since the JAR isn't built 1st... However, these changes seem to make it decently far into the build until `chisel-testers` fails.
> 
> Looks like you've started figuring this out because I see [sbt-sriracha](https://github.com/sbt/sbt-sriracha) added to the `project/plugins.sbt`. However, I don't see you using the `.sourceDependency()` method that sbt-sriracha adds to sbtProject. See https://github.com/sbt/sbt-sriracha#usage
> 
> Let me know if you'd like some help getting the sbt-sriracha stuff ironed out. I'd be happy to submit another PR to merge into your local-chisel34 branch. I talked to @davidbiancolin about this yesterday afternoon. Wasn't sure if you guys would want to switch over to using it and didn't include it in what I had in #717

Thanks for the look. I was just quickly trying to hack stuff up so this may not be the cleanest bump at the moment. 

Funny enough, I wasn't able to duplicate the sbt-sriracha stuff that Rocket does in CY. I would get errors stating that `chisel` couldn't be found in the main CY directory. This was after pointing the `-Dsbt.workspace` to `$PWD/tools/`.  If you revert to https://github.com/ucb-bar/chipyard/pull/719/commits/9d9813fe0abf8146aaade002ef06203b8065c491 you can replicate the issue.

In any case, even though CY is not using sbt-sriracha, it needs to be added in CY's plugins since it is used in FIRRTL/Chisel/RC (I believe that SBT doesn't inherit sub-project plugins... though my SBT knowledge is lacking).

For now this most recent PR seems to build Chisel/FIRRTL properly from source so maybe the issue is using `(project in` vs `ProjectRef`? I'm definitely happy to take an external PR if you have a nicer solution!

jackkoenig:
> Funny enough, I wasn't able to duplicate the sbt-sriracha stuff that Rocket does in CY. I would get errors stating that chisel couldn't be found in the main CY directory. This was after pointing the -Dsbt.workspace to $PWD/tools/. 

How do you invoke SBT? SBT only picks up `.sbtopts` if you invoke it via the `sbt` shell script. Rocket Chip has logic to handle this in the Makefiles since it invokes SBT via `java -jar sbt-launch.jar`: https://github.com/chipsalliance/rocket-chip/blob/b5d797a435e04400281a886c72ee1cfb09e90d88/Makefrag#L28

> In any case, even though CY is not using sbt-sriracha, it needs to be added in CY's plugins since it is used in FIRRTL/Chisel/RC (I believe that SBT doesn't inherit sub-project plugins... though my SBT knowledge is lacking).

This is correct

> For now this most recent PR seems to build Chisel/FIRRTL properly from source so maybe the issue is using (project in vs ProjectRef? 

Make sure to check your classpaths to see if you're getting the Chisel you think you're getting. eg.
```
$ sbt
> projects
# Will list all projects
> show <project> / Compile / dependencyClasspath
```

abejgonzalez:
> How do you invoke SBT? SBT only picks up `.sbtopts` if you invoke it via the `sbt` shell script. Rocket Chip has logic to handle this in the Makefiles since it invokes SBT via `java -jar sbt-launch.jar`: https://github.com/chipsalliance/rocket-chip/blob/b5d797a435e04400281a886c72ee1cfb09e90d88/Makefrag#L28

You can see the SBT invocation in the `variables.mk` file. I follow what RC does for know by checking on if `.sbtopts` exists and adding it to the cmdline invocation. Eventually, I need to check that bloop still works, but Ill get there eventually. 

> Make sure to check your classpaths to see if you're getting the Chisel you think you're getting. eg.
> 
> ```
> $ sbt
> > projects
> # Will list all projects
> > show <project> / Compile / dependencyClasspath
> ```

I will look into this to double-check... but I'm pretty sure it's building the right FIRRTL/Chisel since it is giving the compile warnings for the `tools/{chisel3,firrtl}` repositories. Here is a quick screenshot:

image

.gitmodules
```diff
@@ -76,7 +76,7 @@
 	url = https://github.com/ucb-bar/dsptools.git
 [submodule "tools/chisel-testers"]
 	path = tools/chisel-testers
-	url = https://github.com/freechipsproject/chisel-testers.git
+	url = https://github.com/abejgonzalez/chisel-testers.git
```
abejgonzalez, line 79: Switched to my own fork until `chisel-testers` is updated.

abejgonzalez:
This is able to at least elaborate a the default `RocketConfig`. However, this errors during the FIRRTL compilation with:

image

Ill continue trying to look into why things fail when I get the chance.

timsnyder-siv:
@abejgonzalez can you share the commands you're using in Chipyard so that I can reproduce the error you're running into?  Are you using 'make' commands or just doing things at the SBT prompt right now?

abejgonzalez:
> @abejgonzalez can you share the commands you're using in Chipyard so that I can reproduce the error you're running into? Are you using 'make' commands or just doing things at the SBT prompt right now?

Things are a bit in flux right now. Looks like some things were getting cached so building from a fresh Chipyard repo would get different errors. Im trying to track down some things as we speak. However, in the meantime, the basic steps are:

1. Clone Chipyard
2. Checkout this branch
3. `./script/init-submodules-no-riscv-tools.sh`
4. Add RISCV toolchain to PATH/LD_LIBRARY_PATH/RISCV
5. `make -C sims/vcs`

abejgonzalez:
Ok, I think I have gotten back to the `barstools` annotation error consistently on a fresh CY clone.

In the meantime, I also tried out SBT 1.4.X's `--client` feature with no success. It looks like it just hangs for now. Since this isn't a priority we can look into that in a separate PR.

**For the SBT gods:**
For some reason, `antlr-runtime` was not added to the classpath causing a runtime exception. This is resolved when "hackily" adding it to Chipyards `libraryDependencies`. I think this problem stems from the fact that I can't use a `ProjectRef` when referring to Chisel3 (see <>). I was checking around RC and it looks like the `antlr-runtime` is present on the classpath when Chisel3 uses `ProjectRef` hence FIRRTL works. Does anybody know if my reasoning is correct?

If I use `ProjectRef(file("tools/chisel3"), "chisel")` I get an error saying:

```
No project 'chisel' in 'file:/scratch/abejgonza/chipyard-work/cy-bump-test/'
```

I presume that it is trying to refer to the `chisel` project in `chisel3/build.sbt` but is failing because it cannot get the right directory with `"."`. Is there a way to avoid this?

build.sbt
```diff
@@ -72,43 +67,53 @@ def freshProject(name: String, dir: File): Project = {
...
+val chiselVersion = "3.4.0"
+lazy val chisel = (project in file("tools/chisel3"))
+lazy val chiselLib = "edu.berkeley.cs" %% "chisel3" % chiselVersion
+// While not built from source, *must* be in sync with the chisel3 git submodule
+// Building from source requires extending sbt-sriracha or a similar plugin and
+//   keeping scalaVersion in sync with chisel3 to the minor version
+lazy val chiselPluginLib = "edu.berkeley.cs" % "chisel3-plugin" % chiselVersion cross CrossVersion.full
+
+lazy val firrtlRef = ProjectRef(file("tools/firrtl"), "firrtl")
+lazy val firrtlLib = "edu.berkeley.cs" %% "firrtl" % "1.4-SNAPSHOT"
```
jackkoenig, line 85:
```suggestion
lazy val firrtlLib = "edu.berkeley.cs" %% "firrtl" % "1.4.0"
```
I'd stick to stable versions when using published dependencies. Note that so long as you set `-Dsbt.sourcemode=true`, it's going to be using the source dependency so the version isn't used.

The `chiselVersion` matters a lot more because we can't build the plugin from source and use it here (limitation of SBT, not fundamental), so it's always going to be using the published plugin.

build.sbt
```diff
@@ -72,43 +67,53 @@ def freshProject(name: String, dir: File): Project = {
...
 // Subproject definitions begin
-//
-// FIRRTL is handled as an unmanaged dependency. Make will build the firrtl jar
-// before launching sbt if any of the firrtl source files has been updated
-// The jar is dropped in chipyard's lib/ directory, which is used as the unmanagedBase
-// for all subprojects
-lazy val chisel  = (project in file("tools/chisel3"))
+
+val chiselVersion = "3.4.0"
+lazy val chisel = (project in file("tools/chisel3"))
```
jackkoenig, line 77:
```suggestion
lazy val chisel = ProjectRef(workspaceDirectory / "chisel3", "chisel")
```
@abejgonzalez try doing exactly what RocketChip does here.

This should work with your setting of `-Dsbt.workspace=$PWD/tools` as you do in `.sbtopts`.

build.sbt
```diff
@@ -72,43 +67,53 @@ def freshProject(name: String, dir: File): Project = {
...
+lazy val chiselLib = "edu.berkeley.cs" %% "chisel3" % chiselVersion
+// While not built from source, *must* be in sync with the chisel3 git submodule
+// Building from source requires extending sbt-sriracha or a similar plugin and
+//   keeping scalaVersion in sync with chisel3 to the minor version
+lazy val chiselPluginLib = "edu.berkeley.cs" % "chisel3-plugin" % chiselVersion cross CrossVersion.full
+
+lazy val firrtlRef = ProjectRef(file("tools/firrtl"), "firrtl")
```
jackkoenig, line 84:
```suggestion
lazy val firrtlRef = ProjectRef(workspaceDirectory / "firrtl", "firrtl")
```

Same thing as `chisel`.

build.sbt
abejgonzalez: Sounds good. I've been changing alot of `X.X.SNAPSHOT` versions to `X.X.+` which should be the same thing (iirc)

build.sbt
abejgonzalez: Yes. I found the issue here. It was due to FireSim referencing the `chisel` project in Chipyard. However, if you make it a `ProjectRef` the project doesn't exist in Chipyard (just in the `tools/chisel3/`). I'll push up a change that addresses this soon.

build.sbt
abejgonzalez: Sure.

abejgonzalez:
Ok bumped alot of repos. Please take a look if you have a repo that is bumped. I basically tried to do 1 of 2 things with `build.sbt` files:

1. Simplify the `build.sbt` so that Chipyard (or another top-level `build.sbt` file) specifies the necessary dependencies. This is similar to what Hwacha does.
2. Fix current `build.sbt` files so that it can build independently of Chipyard... which the caveat that Chipyard can override its settings.

I also tried to experiment around with Chipyards `build.sbt`. I drop the dependencies of Chisel, FIRRTL, RC, Treadle, Chisel-Testers, Firrtl-Interpreter from all sub-projects so that they can depend on the source built versions. However, when a project is dropped, all lib. deps. that might be brought in by that project will also be dropped. Thus, I added a bit of extra stuff to add dropped lib. deps. (that aren't Chisel, FIRRTL, ....) back in.

abejgonzalez:
As https://github.com/firesim/firesim/pull/651#issuecomment-716857208 mentioned, SBT now supports a thin client similar to `bloop`s current functionality. I've gone ahead and added support for it through `export ENABLE_SBT_THIN_CLIENT=1`. Unlike `bloop` if you update the `build.sbt` you should 1st shutdown the existing server before continuing other work.

abejgonzalez:
**Bloop v SBT Thin Performance Numbers on `dev` version of CY:**

After running `make -C sims/vcs verilog` once. Compiling/Downloading everything.

SBT (normal): 10m45
SBT (thin client): 9m
Bloop: 6m31

After running `make -C sims/vcs verilog` once (everything already compiled) then changing a single file in Chipyard (adding a small println to `IOBinders.scala`).

SBT (normal): 4m38
SBT (thin client): 2m54
Bloop: 2m47

After running `make -C sims/vcs verilog` once (everything already compiled) then changing a single file in Rocket Chip (adding a single newline to the file). Fairly similar to the last test since it doesn't trigger recompilation of downstream items.

SBT (normal): 4m37
SBT (thin client): 3m04
Bloop: 2m52

On a side note, I can also confirm that you can have multiple thin clients working in different repos.

davidbiancolin:
Very cool -- go ahead and strip out the Bloop support then.

jerryz123:
> Very cool -- go ahead and strip out the Bloop support then.

RIP Bloop (2020-2020), a feature we never actually released into master.

common.mk
```diff
@@ -63,37 +63,23 @@ SCALA_SOURCES = $(call lookup_srcs,$(SOURCE_DIRS),scala)
...
-	$(MAKE) -C $(CHIPYARD_FIRRTL_DIR) SBT="$(SBT)" root_dir=$(CHIPYARD_FIRRTL_DIR) build-scala
-	mkdir -p $(@D)
-	cp -p $(CHIPYARD_FIRRTL_DIR)/utils/bin/firrtl.jar $@
-	touch $@
-
-$(FIRRTL_TEST_JAR): $(call lookup_srcs,$(CHIPYARD_FIRRTL_DIR),scala)
-	cd $(CHIPYARD_FIRRTL_DIR) && $(SBT) "test:assembly"
-	mkdir -p $(@D)
-	cp -p $(CHIPYARD_FIRRTL_DIR)/utils/bin/firrtl-test.jar $@
+$(SBT_THIN_CLIENT_TIMESTAMP): $(SBT_SOURCES)
```
abejgonzalez, line 71: This is a bit hacky.

If you don't do this, changing the `build.sbt` would require you to `$(SBT) shutdown` the server then continue with the normal workflow (i.e. run `make verilog` or whatever afterwards). Doing this would re-spin up the thin server and reload all the project definitions. 

Once the https://github.com/sbt/sbt/issues/4638 issue in SBT is resolved, we can avoid this `ifneq` logic by just doing:

```
cd $(base_dir) && $(SBT) "shutdown"
touch $@
```

The reason I don't do this now is that doing `$(SBT) shutdown` the 1st time using the thin client will spin-up the server, then kill it resulting in the next `$(SBT)` call spinning-up the server again (unnecessary loading project def.s twice on both spinups).

davidbiancolin: Is there no means to send a `reload` command to the server?

abejgonzalez: Actually, you don't want to send a `reload` in every case that the `build.sbt` is changed. For the same reason mentioned in the last paragraph. Sending a `reload` before the server is started will read the proj. def.s twice. Hence I "hack" around this by checking if the `SBT_TIMESTAMP` file exists to avoid sending an extra `reload` when the server is down (which is most likely when the `SBT_TIMESTAMP` file doesn't exist... though not always the case).

Here is timeline of events in the **good** case (what I do):
1. TIMESTAMP file doesn't exist
2. `make ...`
3. Create TIMESTAMP file WITHOUT running `reload`
4. First SBT call is done (to `GenSimFiles`)... which reads the proj. def. and starts the server
5. Finished Verilog
6. Touch `build.sbt` and re-run `make ...`
7.  Update TIMESTAMP file and run `reload` to update the server

Here is timeline of events in the **bad** case (what I don't do):
1. TIMESTAMP file doesn't exist
2. `make ...`
3. Create TIMESTAMP file and run `reload` (This reads the proj. def.s twice, once for starting the server, once for `reload`). This would be the 1st SBT call.
4. 2nd SBT call is done (to `GenSimFiles`)... with server already started
5. Finished Verilog
6. Touch `build.sbt` and re-run `make ...`
7.  Update TIMESTAMP file and run `reload` to update the server

davidbiancolin: I don't think this exactly what it was intended for but you could use the presence of the named-pipe
for the sbt server (https://www.scala-sbt.org/1.x/docs/sbt-server.html) to detect if it's running, You could instead touch `project/target/active.json` instead of creating and managing your own timestamp.

abejgonzalez: That looks good to me and seems to pass a couple of quick scenarios. Good catch.

abejgonzalez: Ok for future reference this is the event sequence when the server isn't started:

1. `active.json` doesn't exist
2. `make ...`
3. Run `sbt --client exit` to start server and quickly exit. (1st call)
4. Normal `make ...` sequence (`GenSimFiles`..., etc)
5. Touch `build.sbt`
6. Touch `active.json` and `sbt reload` to update the server.

build.sbt
```diff
@@ -1,30 +1,34 @@
...
   scalacOptions ++= Seq("-deprecation","-unchecked","-Xsource:2.11"),
-  libraryDependencies += "org.scalatest" %% "scalatest" % "3.0.8" % "test",
-  libraryDependencies += "org.json4s" %% "json4s-jackson" % "3.6.1",
-  libraryDependencies += "org.scala-lang" % "scala-reflect" % scalaVersion.value,
-  libraryDependencies += "com.github.scopt" %% "scopt" % "3.7.0",
-  libraryDependencies += "org.scala-lang.modules" % "scala-jline" % "2.12.1",
-  libraryDependencies += "com.typesafe.play" %% "play-json" % "2.6.10",
-  libraryDependencies += "org.typelevel" %% "spire" % "0.16.2",
-  libraryDependencies += "org.scalanlp" %% "breeze" % "1.0",
   addCompilerPlugin("org.scalamacros" % "paradise" % "2.1.0" cross CrossVersion.full),
```
timsnyder-siv, line 16:
```suggestion
  addCompilerPlugin("org.scalamacros" % "paradise" % "2.1.1" cross CrossVersion.full),
```
The recent bump of scalaVersion to 2.12.12 for mdf causes me to hit:
```
[error] stack trace is suppressed; run last mdf / update for the full output
[error] (mdf / update) sbt.librarymanagement.ResolveException: Error downloading org.scalamacros:paradise_2.12.12:2.1.0
[error]   Not found
[error]   Not found
[error]   not found: /home/centos/.ivy2/local/org.scalamacros/paradise_2.12.12/2.1.0/ivys/ivy.xml
[error]   not found: https://repo1.maven.org/maven2/org/scalamacros/paradise_2.12.12/2.1.0/paradise_2.12.12-2.1.0.pom
[error]   not found: https://oss.sonatype.org/content/repositories/snapshots/org/scalamacros/paradise_2.12.12/2.1.0/paradise_2.12.12-2.1.0.pom
[error]   not found: https://oss.sonatype.org/content/repositories/releases/org/scalamacros/paradise_2.12.12/2.1.0/paradise_2.12.12-2.1.0.pom
[error]   not found: /home/centos/.m2/repository/org/scalamacros/paradise_2.12.12/2.1.0/paradise_2.12.12-2.1.0.pom
```

Bumping paradise to version 2.1.1 in CY fixes the issue.  We may want to bump to 2.1.1 here as well.

common.mk
```diff
@@ -236,6 +225,24 @@ dramsim_lib = $(dramsim_dir)/libdramsim.a
...
+# Helper to run SBT or manage the SBT server
+################################################
+
+SBT_COMMAND ?= shell
+.PHONY: launch-sbt
+launch-sbt:
+	cd $(base_dir) && $(SBT_NON_THIN) "$(SBT_COMMAND)"
+
+.PHONY: shutdown-sbt-server
+shutdown-sbt-server:
```
davidbiancolin, line 238: We should probably use the same commands across both firesim / CY.

build.sbt
abejgonzalez: Do you know if `paradise` is needed outside of Rocket Chip? Can we just remove that plugin entirely? If not we can bump this.

common.mk
abejgonzalez: Sure.

generators/firechip/src/test/scala/ScalaTestSuite.scala
```diff
@@ -42,10 +42,9 @@ abstract class FireSimTestSuite(
   }
 
   def runTest(backend: String, name: String, debug: Boolean, additionalArgs: Seq[String] = Nil) = {
-    behavior of s"${name} running on ${backend} in MIDAS-level simulation"
     compileMlSimulator(backend, debug)
     if (isCmdAvailable(backend)) {
-      it should s"pass" in {
+      it should s"pass in ML simualtion on ${backend}" in {
```
abejgonzalez, line 47:
```suggestion
      it should s"pass in ML simulation on ${backend}" in {
```

abejgonzalez: If CI passes you can just add a `[ci skip]` on the commit message to skip running CI on this. CI will run when everything is merged.

davidbiancolin (commented):
I don't really understand the rationale for some of the SBT changes -- some comments should be added to explain intent, particularly regarding the ANTLR change, and the key-lookup for library dependencies.

davidbiancolin:
Other than that, this LGTM.

abejgonzalez:
> I don't really understand the rationale for some of the SBT changes -- some comments should be added to explain intent, particularly regarding the ANTLR change, and the key-lookup for library dependencies.

Brief Explanations / Reasoning on `build.sbt` changes:
- Removing `commonSettings` library dependencies: Most of these libraries are included in the sub-projects now so this is now unneeded. Additionally, some of these libdeps were needed since RC/FIRRTL/Chisel was dropped using the `allDependencies` in the past so their libdeps would not be inherited.
- Change to `allDependencies`: This is just a more selective version of the prior `allDependencies` so that only the major projects are dropped instead of every project (since all Berkeley projects have a `edu.berkeley.cs`).
- Removing `conditionalDependsOn`: Replaced in favor of `sbt-sriracha` which is what RC uses. Instead of `-DROCKET_USE_MAVEN` you now use the `-Dsbt.sourcemode=` flag.
- Chisel/FIRRTL projects: This uses `sbt-sriracha` to choose between the `ProjectRef` version of the project or the libdep from Maven. Right now Chipyard uses the `ProjectRef`. Additionally, both projects store their libdeps so that they can be added to downstream projects (important because the `allDependencies` task drops these projects from the dependencies and thus their libdeps are also dropped). FIRRTL's libdeps are a bit weird, since it uses the ANTLR plugin, if the `antlr4` libdep is included without setting up the plugin then it will break.
- FIRRTLInterpreter/Treadle/ChiselTesters/RC: They use the Chisel/FIRRTL libdeps and also store their own libdeps.
- Tapeout: The `handlebars` libdep isn't used.
- Rest: Hopefully straightforward. Uses `sourceDependency` for `sbt-sriracha`, adds libdeps for major repos that are dropped, if was a `freshProject` then move the libdeps from the subproject into this file.

davidbiancolin (approved):
LGTM let's ship this

PR #721, opened 2020-11-17
Add config fragment for replacing L2 with broadcastManager

jerryz123:
**Related issue**: 
Surprisingly, there exists no predefined config fragment in rocketchip to set this. 

### DIFF
--- a/generators/chipyard/src/main/scala/ConfigFragments.scala
+++ b/generators/chipyard/src/main/scala/ConfigFragments.scala
@@ -151,6 +151,11 @@ class WithRocketDCacheScratchpad extends Config((site, here, up) => {
   }
 })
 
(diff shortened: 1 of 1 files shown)


colinschmidt (approved):
One doesn't exist upstream because it is the default value of this key. This makes sense for chipyard which assumes the default is an L2.

colinschmidt (commented):
Speaking of that default do we need to force nBanks = 1?

colinschmidt:
https://github.com/chipsalliance/rocket-chip/blob/c7d201f10eb14437d0303237a07b6890a0182309/src/main/scala/subsystem/BankedL2Params.scala#L29

jerryz123:
I think its orthogonal to `nBanks`.

Issue #735, opened 2020-12-08
Release 1.3.0 libgloss installation error


SaltedFishLZ:
**Impact**: unknown | other
Hi, I am following the tutorial to set up chipyard (https://chipyard.readthedocs.io/en/latest/Chipyard-Basics/Initial-Repo-Setup.html), but it seems `libgloss` cannot be configured correctly and my installation is aborted.

**Tell us about your environment:**
*Chipyard Version:* 1.3.0, Hash: c576a7e76717376b7a2cbbc0a531f51382ff87f1
*OS:* `Ubuntu 18.04 Docker Image` 
*Other:* 
https://chipyard.readthedocs.io/en/latest/Chipyard-Basics/Initial-Repo-Setup.html

**What is the current behavior?**

After running `./scripts/build-toolchains.sh riscv-tools`, the compilation process will be aborted due to an error, which shows "configure: error: The target architecture of the C compiler does not appear to be RISC-V. If cross-compiling, specify a target triplet such as '--host=riscv64-unknown-elf'."

**What is the expected behavior?**

**Other information**

Here is my log
```
riscv64-unknown-elf-objdump --disassemble-all --disassemble-zeroes --section=.text --section=.text.startup --section=.text.init --section=.data rv64mi-p-sbreak > rv64mi-p-sbreak.dump
make[1]: Leaving directory '/chipyard/toolchains/riscv-tools/riscv-tests/build/isa'
==>  Installing riscv-tests
mkdir -p benchmarks
mkdir -p isa
/usr/bin/make -C benchmarks -f /chipyard/toolchains/riscv-tools/riscv-tests/build/../benchmarks/Makefile src_dir=/chipyard/toolchains/riscv-tools/riscv-tests/build/../benchmarks XLEN=64
/usr/bin/make -C isa -f /chipyard/toolchains/riscv-tools/riscv-tests/build/../isa/Makefile src_dir=/chipyard/toolchains/riscv-tools/riscv-tests/build/../isa XLEN=64
make[1]: Entering directory '/chipyard/toolchains/riscv-tools/riscv-tests/build/benchmarks'
make[1]: Nothing to be done for 'default'.
make[1]: Leaving directory '/chipyard/toolchains/riscv-tools/riscv-tests/build/benchmarks'
make[1]: Entering directory '/chipyard/toolchains/riscv-tools/riscv-tests/build/isa'
make[1]: Nothing to be done for 'default'.
[...]

PR #739, opened 2020-12-14
Disable CC and CXX overrides for libgloss build

a0u:
**Related issue**: #735

**Type of change**: bug fix

**Impact**: other

**Release Notes**

This permits users to specify `CC` and `CXX` for building the host software without impacting cross-compilation of libgloss-htif, same as how riscv-pk is handled.

### DIFF
--- a/scripts/build-toolchains.sh
+++ b/scripts/build-toolchains.sh
@@ -135,7 +135,7 @@ module_all riscv-tests --prefix="${RISCV}/riscv64-unknown-elf"
 
 # Common tools (not in any particular toolchain dir)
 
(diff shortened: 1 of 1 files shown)


alonamid (approved):
LGTM. Can We make sure that CI didn't cache the tools when it ran the tests for this? (otherwise we find out build script failures only several PRs later...)

abejgonzalez:
To do that you should update this to `v6`: https://github.com/ucb-bar/chipyard/blob/cca3cd087c1c5a54f98867bc4a89129c1c299b94/.circleci/config.yml#L6-L9

a0u:
This has no effect on CI as it does not set either `CC` or `CXX` in the environment.  It was manually tested.

PR #740, opened 2020-12-15
[vlsi] Add Makefile flag to use memory compiler defined in tech library

jerryz123:


colinschmidt (approved):
Looks fine to me. I think we should maybe change the name to `USE_SRAM_COMPILER` since "compiler" is a relatively generic name.
Also make sure to add documentation for this to the vlsi docs PR.

harrisonliew:
Just want to also mention that nangate45 dummy SRAMs only have timing models for a typical corner, but no simulation models or geometry. Perhaps we should also add it to the default synflops only list.

Issue #555, opened 2020-05-19
Invalid build URI


ConstantPark:
This is my error symptom.
I don't know "Invalid build URI"

[info] Loading settings for project chipyard-build from plugins.sbt ...
[info] Loading project definition from /home/esoc/1.Workspace_1/constant/0.RISCV/mmio/temp/chipyard/project
[info] Updating ProjectRef(uri("file:/home/esoc/1.Workspace_1/constant/0.RISCV/mmio/temp/chipyard/project/"), "chipyard-build")...
[info] Done updating.
[warn] There may be incompatibilities among your library dependencies; run 'evicted' to see detailed eviction warnings.
[info] Loading settings for project dsptools from build.sbt ...
[info] Loading settings for project barstoolsMacros from build.sbt ...
[info] Loading settings for project mdf from build.sbt ...
[info] Loading settings for project boom from build.sbt ...
[info] Loading settings for project hwacha from build.sbt ...
[info] Loading settings for project icenet from build.sbt ...
[info] Loading settings for project testchipip from build.sbt ...
[info] Loading settings for project hardfloat from build.sbt ...
[info] Loading settings for project chisel_testers from build.sbt ...
[info] Loading settings for project treadle from build.sbt ...
[info] Loading settings for project firrtl_interpreter from build.sbt ...
[info] Loading settings for project chisel from build.sbt ...
[info] Loading settings for project chipyard from build.sbt ...
[error] java.lang.RuntimeException: Invalid build URI (no handler available): file:/home/esoc/1.Workspace_1/constant/0.RISCV/mmio/temp/chipyard/sims/firesim/sim/
[error]         at scala.sys.package$.error(package.scala:26)
[error]         at sbt.internal.Load$.$anonfun$builtinLoader$1(Load.scala:481)
[error]         at sbt.internal.BuildLoader.$anonfun$apply$3(BuildLoader.scala:240)
[error]         at scala.Option.getOrElse(Option.scala:121)
[error]         at sbt.internal.BuildLoader.apply(BuildLoader.scala:240)
[error]         at sbt.internal.Load$.loadURI$1(Load.scala:546)
[error]         at sbt.internal.Load$.loadAll(Load.scala:562)
[error]         at sbt.internal.Load$.loadURI(Load.scala:492)
[error]         at sbt.internal.Load$.load(Load.scala:471)
[error]         at sbt.internal.Load$.$anonfun$apply$1(Load.scala:251)
[...]

PR #745, opened 2020-12-23
Update MINGIT version to 1.8.5

jerryz123:
1.8.5 is necessary for `git -C` to work.

**Related issue**: https://github.com/ucb-bar/chipyard/issues/555

**Type of change**: bug fix

**Impact**:  software change


alonamid (approved):
True

PR #750, opened 2020-12-29
Small CI Cleanup + CI Fix

abejgonzalez:
**Related issue**: 

**Impact**: other

**Release Notes**
Small changes to improve CI a bit. First, when looking at the `check-commit` printout, split it into all 0's (valid) and all 1's (invalid, not on `master` branch) so that it is less annoying trying to find the 1 in the sea of 0's. Second, don't transfer `.git` folder on copies since we don't do `git` operations on the build server. Additionally, it adds a new SSH key to login to the build server.

### DIFF
--- a/.circleci/check-commit.sh
+++ b/.circleci/check-commit.sh
@@ -124,19 +124,25 @@ search
 # turn off verbose printing to make this easier to read
 set +x
 
-# print all result strings
+# print 0's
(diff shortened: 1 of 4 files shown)


jerryz123 (approved):

abejgonzalez:
Note: CI is currently down due to build server issues. I will restart when all is resolved.

Issue #595, opened 2020-06-05
init submodule script leaves gitconfig in modified state if submodule update fails


colinschmidt:
**Impact**: software

**Tell us about your environment:**
*Chipyard Version:* 1.3

**What is the current behavior?**
If I have modifications in a submodule that would be updated by `init-submodules-no-riscv-tools.sh` the `git submodule update` command will fail and cause the script to exit early, leaving some of the `git config *.update none` commands to remain in effect.

**What is the expected behavior?**
The script should catch this error and return the `git config` back to the initial state before exiting

PR #761, opened 2021-01-14
Ensure git config changes are reverted on failure

a0u:
**Related issue**: #595

**Type of change**: bug fix

**Impact**: other

**Release Notes**
This ensures that the temporary `git config` settings for excluding certain submodules from the recursive update are properly reverted if the submodule update fails or is interrupted by SIGINT or SIGTERM.

The git commands have been refactored to loop over a centralized blacklist, which should reduce the potential for human errors when updating this script in the future.

### DIFF
--- a/scripts/init-submodules-no-riscv-tools-nolog.sh
+++ b/scripts/init-submodules-no-riscv-tools-nolog.sh
@@ -20,43 +20,41 @@ fi
 DIR="$(dirname "$(readlink -f "${BASH_SOURCE[0]}")")"
 CHIPYARD_DIR="$(dirname "$DIR")"
 
-# Ignore toolchain submodules
 cd "$CHIPYARD_DIR"
-for name in toolchains/*-tools/*/ ; do
-    git config submodule."${name%/}".update none
-done
-git config submodule.toolchains/libgloss.update none
-git config submodule.toolchains/qemu.update none
 
-# Don't automatically initialize generators with big submodules (e.g. linux source)
-git config submodule.generators/sha3.update none
-git config submodule.generators/gemmini.update none
+(
+    # Blocklist of submodules to initially skip:
+    # - Toolchain submodules
+    # - Generators with huge submodules (e.g., linux sources)
+    # - FireSim until explicitly requested
+    # - Hammer tool plugins
+    git_submodule_exclude() {
+        # Call the given subcommand (shell function) on each submodule
+        # path to temporarily exclude during the recursive update
+        for name in \
+            toolchains/*-tools/*/ \
+            toolchains/libgloss \
+            toolchains/qemu \
+            generators/sha3 \
+            generators/gemmini \
+            sims/firesim \
+            vlsi/hammer-cadence-plugins \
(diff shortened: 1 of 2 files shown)


scripts/init-submodules-no-riscv-tools-nolog.sh
```diff
@@ -65,10 +63,9 @@ git submodule update --init generators/sha3
 git submodule update --init generators/gemmini
 git -C generators/gemmini/ submodule update --init --recursive software/gemmini-rocc-tests
 
-git config --unset submodule.sims/firesim.update
 # Minimal non-recursive clone to initialize sbt dependencies
 git submodule update --init sims/firesim
-git config submodule.sims/firesim.update none
+git config --local submodule.sims/firesim.update none
```
a0u, line 68: Is there a reason to keep the firesim submodule permanently masked?  I'm inclined to omit this to be consistent with the treatment of other submodules.

davidbiancolin: I think it was mostly to stop people from having an out-of-control recursive update (given the circular dependency between CY <-> FireSim) if they accidentally called `git submodule update` themselves instead of using the script.

a0u: Ah, yeah, that makes a lot of sense.

As an additional measure, `update = none` could be added directly to the firesim section in `.gitmodules` so it applies by default, not just after the init-submodules script runs.  Then only explicitly running `git submodule update --checkout sims/firesim` would bypass the setting.

scripts/init-submodules-no-riscv-tools-nolog.sh
```diff
@@ -20,43 +20,41 @@ fi
...
-    git config submodule."${name%/}".update none
-done
-git config submodule.toolchains/libgloss.update none
-git config submodule.toolchains/qemu.update none
 
-# Don't automatically initialize generators with big submodules (e.g. linux source)
-git config submodule.generators/sha3.update none
-git config submodule.generators/gemmini.update none
+(
+    # Blacklist of submodules to initially skip:
```
colinschmidt, line 26: Can we change this to skiplist or blocklist?

colinschmidt (approved):
LGTM.

scripts/init-submodules-no-riscv-tools-nolog.sh
davidbiancolin: Ah, i didn't know you could do that. That does seems cleaner.

scripts/init-submodules-no-riscv-tools-nolog.sh
a0u: Sure, I went with "blocklist".

scripts/init-submodules-no-riscv-tools-nolog.sh
a0u: This works in testing, but I'll leave that for a future PR.

jerryz123:
Merging to move forwards with release

Issue #732, opened 2020-12-03
Prebuilt Toolchain Does Not Match Submodules


davidbiancolin:
**Impact**: software 

**What is the current behavior?**
Uses a significantly older version of the gnu-toolchain that has the gcc bug present. This breaks some SPEC2017 benchmarks. 

**What is the expected behavior?**
Use the submoduled version.

**Other information**
This should be done before the next release. Opening this mostly as a reminder.

alonamid:
Fixed in release 1.4

PR #763, opened 2021-01-18
Bump Prebuilt Toolchain

alonamid:
**Related issue**: #732 

**Type of change**: bug fix

**Impact**: other

**Release Notes**


davidbiancolin (approved):
LGTM

davidbiancolin:
This is out-of-scope for this PR, but I'm slightly concerned about having the tars in the main repo as it might lead to considerable bloating.  It seems to me using a submodule or an S3 bucket, that we only pull in on running `ec2fast` would be desirable. Or @a0u, do think these tarsballs are small enough it shouldn't matter

a0u:
@davidbiancolin:  Yeah, git LFS or S3 would be the proper solution, although I don't think the problem is that urgent just yet.
Compared to the old FireSim prebuilt toolchain, there are two measures that improve the efficiency of storing the tarballs in git:
* This repo is shallow-cloned by Chipyard during the install, so the blobs corresponding to the older builds aren't actually fetched:
https://github.com/ucb-bar/chipyard/blob/cca3cd087c1c5a54f98867bc4a89129c1c299b94/scripts/build-toolchains.sh#L81-L83
(Note: This means we have to be diligent about tagging commits so they remain reachable by past Chipyard releases.)
* All x86 binaries are also stripped, which saves considerable space.

However, the git history bloat will eventually make it unwieldy to maintain, although at our release cadence (every ~6 months) we still have some headroom.

davidbiancolin:
Oh, i'm stupid. It's already a submodule. + Shallow clone, opts -- yeah no reason to change this anytime soon. Sorry for the confusion.

jerryz123 (approved):

a0u:
@alonamid: Did the riscv-gnu-toolchain commit hash not change?  What is different about this version?

davidbiancolin:
For some reason, the commited tarballs didn't match the hash / checked in sources. When I built an SPEC with the prebuilt toolchain I'd see failures  consistent with an old riscv-gcc bug, whereas if i built teh toolchain from scratch they ran cleanly. I checked the submodule pointers and they looked good. So Alon kindly rebuilt the toolchain.

davidbiancolin:
I think @NathanTP tried running spec2017 recently and had problems. This is probably why.

a0u (approved):
Ah ok, that was probably my fault somehow.

PR #771, opened 2021-01-22
Update dockerfiles

schwarz-em:
**Type of change**: new feature + other enhancement

**Impact**: other

**Release Notes**
This adds a dockerfile with a complete chipyard set-up and its corresponding entrypoint script. The updates to the CI dockerfile use a portion of the build stages from the main dockerfile.


abejgonzalez:
Looks like this is pretty close to running all of the CI tests. I think we just need to add `python2` to the requirements scripts for tracegen tests to run.

Dockerfiles/CI-Dockerfile
```diff
@@ -0,0 +1,43 @@
+### Note: This DockerFile is adapted from https://github.com/ucb-bar/chipyard/blob/master/.circleci/images/Dockerfile which was adapted from: https://github.com/CircleCI-Public/example-images/openjdk
```
abejgonzalez, line 1: Nit(s): Lets change the folder to `dockerfiles` since everything else in the repo is lowercase. Also by the next Chipyard release (when this gets put on `master`) this comment will be inaccurate. Maybe just delete the reference to `master` and add a hash or just delete the reference to CY `master`.

Dockerfiles/CI-Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+# This is a chipyard setup for CI, no toolchains are initialized
```
abejgonzalez, line 2: Now that we know this mainly works. Can we dedup this Dockerfile with the other one so that the only difference is the `init-submodules` and `build-toolchains` lines? Something like this: https://github.com/QuarantineGemmini/Dockerfiles

Dockerfiles/Dockerfile
```diff
@@ -0,0 +1,55 @@
+### Note: This DockerFile is adapted from https://github.com/ucb-bar/chipyard/blob/master/.circleci/images/Dockerfile which was adapted from: https://github.com/CircleCI-Public/example-images/openjdk
+# This is a full chipyard setup
+
+FROM ubuntu:18.04
+ARG CHIPYARD_HASH
```
abejgonzalez, line 5: Is there any way to specify a "default" argument?

abejgonzalez:
Some other thoughts:

- Can you also move the `.circleci/images/README.md` to the new `dockerfile` folder and write a bit about the work done?
- It would be nice to have the CI automatically build the Dockerfile and push it up to Dockerhub (we can move that to a separate PR but it would be a nice QoL change).

Dockerfiles/Dockerfile
schwarz-em: There is a way, but it also just stays on the most recent master branch commit by default even when no argument is given. Would we want the default to be a different commit?

abejgonzalez: Actually, I think the main Dockerfile (the one with the toolchains) doesn't need a default argument. If an external user tries to use the Dockerfile (modify and build it themselves) I think pointing them to the most recent Chipyard is fine. 

As for the CI Dockerfile, I think knowing what commit was used is valuable (unless we somehow bake in that information into the version - which is another option).

scripts/ubuntu-req.sh
```diff
@@ -19,6 +19,7 @@ sudo apt-get install -y libgtk-3-dev gettext
 sudo apt-get install -y python3-pip python3.6-dev rsync libguestfs-tools expat ctags
 # install DTC
 sudo apt-get install -y device-tree-compiler
+sudo apt-get install -y python
```
abejgonzalez, line 22: This also needs to be updated in the `centos-req.sh`

dockerfiles/README.md
```diff
@@ -0,0 +1,20 @@
+General
+-------
+This DockerFile contains the necessary steps to build a Docker container that can run
```
abejgonzalez, line 3: Nice. I think in addition to this, we can write some small thing on accessing this Dockerfile in this area (so people can use it if they don't want to install it manually): https://chipyard.readthedocs.io/en/latest/Chipyard-Basics/Initial-Repo-Setup.html#initial-repository-setup. Basically something like: "Another option to use Chipyard is to use one of Chipyard's pre-build Docker images. It can be located here <LINK>. Where the image tag is the major version of Chipyard".

dockerfiles/CI-Dockerfile
```diff
@@ -0,0 +1,43 @@
+### Note: This DockerFile is adapted from: https://github.com/CircleCI-Public/example-images/openjdk
```
abejgonzalez, line 1: Reminder to remove this.

docs/Chipyard-Basics/Initial-Repo-Setup.rst
```diff
@@ -66,6 +66,22 @@ You can put this in your ``.bashrc`` or equivalent environment setup file to get
 
 These variables need to be set for the ``make`` system to work properly.
 
+Pre-built Docker Image
+-------------------------------------------
+
+An alternative to setting up the Chipyard repository locally is to pull the pre-built Docker image from Docker Hub. The image comes with all necessary dependencies and toolchains installed.
```
abejgonzalez, line 72: I would mention that this docker image has all toolchains included, all dependencies installed, and Chipyard cloned.

abejgonzalez: I might also mention that the tag of the Docker image corresponds to the Chipyard version that is cloned/setup there.

abejgonzalez:
Looks like some tests are still failing. I would try to still increase the timeout.

docs/Chipyard-Basics/Initial-Repo-Setup.rst
```diff
@@ -66,6 +66,22 @@ You can put this in your ``.bashrc`` or equivalent environment setup file to get
 
 These variables need to be set for the ``make`` system to work properly.
 
+Pre-built Docker Image
+-------------------------------------------
+
+An alternative to setting up the Chipyard repository locally is to pull the pre-built Docker image from Docker Hub. The image comes with all dependencies installed, chipyard cloned, and toolchains initialized. Each image comes with a tag that corresponds to the version of Chipyard cloned/set-up in that image. Not including a tag during the pull will pull the image with the latest version of chipyard.
```
abejgonzalez, line 72:
```suggestion
An alternative to setting up the Chipyard repository locally is to pull the pre-built Docker image from Docker Hub. The image comes with all dependencies installed, Chipyard cloned, and toolchains initialized. Each image comes with a tag that corresponds to the version of Chipyard cloned/set-up in that image. Not including a tag during the pull will pull the image with the latest version of Chipyard.
```

abejgonzalez:
Note: We will release the 1st version of this Docker image (with all deps, toolchains, etc) with 1.5.0 to Dockerhub.

abejgonzalez (approved):
LGTM

scripts/centos-req.sh
```diff
@@ -18,6 +18,7 @@ sudo yum install -y centos-release-scl
 sudo yum install -y devtoolset-8-make
 # install DTC
 sudo yum install -y dtc
+sudo yum install -y python
```
alonamid, line 21: A few lines above there is a python36, python36-pip and python36-devel installation. Is this python installation for an earlier or later version?

alonamid (commented):
LGTM overall.
I had one question about the Python requirement, and I'm also wondering if it might be worth mentioning in the docs section that this docker image only sets up baseline Chipyard (not including FireMarshal, FireSim and Hammer initializations)

dockerfiles/Dockerfile
```diff
@@ -0,0 +1,68 @@
...
+# Update PATH for RISCV toolchain (note: hardcoded for CircleCI)
+ENV RISCV="/home/riscvuser/riscv-tools-install"
+ENV LD_LIBRARY_PATH="$RISCV/lib"
+ENV PATH="$RISCV/bin:$PATH"
+
+# Install Chipyard and run ubuntu-req.sh to install necessary dependencies
+RUN git clone https://github.com/ucb-bar/chipyard.git && \
+        cd chipyard && \
+        git checkout $CHIPYARD_HASH && \
+        ./scripts/ubuntu-req.sh 1>/dev/null
```
tymcauley, line 40: Since this runs `apt-get ...` internally, you might want to add `rm -rf /var/lib/apt/lists/*` at the conclusion. In general though, [this page](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run) mentions that running `apt-get update` from two separate steps (since you run it from within this script, and also from the very first `RUN` command) can cause weird issues, since build caching can mean that different `apt install` commands are working with different `apt` source lists.

tymcauley (commented):
Looks great! Mostly I have some nitpicks from the `apt-get` section of [this page](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run), but those could be fixed in a subsequent PR.

scripts/centos-req.sh
schwarz-em: It would ideally be for a later version. I added this because there was a bug I ran into with ubuntu-req.sh where one of the tests was looking for python2 instead of python3. I don't know a way to test on centos, so I'm not entirely sure if the bug would still occur there, so I added it just in case.

alonamid: which test was looking for python2?

schwarz-em: It was both of the tracegen tests.

dockerfiles/Dockerfile
```diff
@@ -0,0 +1,67 @@
...
+ENV RISCV="/home/riscvuser/riscv-tools-install"
+ENV LD_LIBRARY_PATH="$RISCV/lib"
+ENV PATH="$RISCV/bin:$PATH"
+
+# Install Chipyard and run ubuntu-req.sh to install necessary dependencies
+RUN git clone https://github.com/ucb-bar/chipyard.git && \
+        cd chipyard && \
+        git checkout $CHIPYARD_HASH && \
+        ./scripts/ubuntu-req.sh 1>/dev/null && \
+        sudo rm -rf /var/lib/apt/lists/*
```
abejgonzalez, line 39: Aren't you already super user here? Can you remove the `sudo`?

dockerfiles/Dockerfile
```diff
@@ -0,0 +1,67 @@
...
+
+# Install dependencies for ubuntu-req.sh
+RUN apt-get update && \
+    DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
+               curl \
+               git \
+               sudo \
+               ca-certificates \
+               keyboard-configuration \
+               console-setup
```
abejgonzalez, line 18: Should this be finished with a `rm -rf /var/lib/apt/lists/*` as well?

schwarz-em: I tried that, but then when ubuntu-req.sh was run during the build, it wasn't able to find any of the packages.

dockerfiles/Dockerfile
```diff
@@ -0,0 +1,68 @@
...
+# BUILD BASE FOR CI
+
+FROM ubuntu:18.04 as base
+ARG CHIPYARD_HASH
+
+MAINTAINER https://groups.google.com/forum/#!forum/chipyard
+
+# Install dependencies for ubuntu-req.sh
+RUN apt-get update && \
+    apt-get upgrade -y && \
```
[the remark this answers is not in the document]
schwarz-em: Thanks for the tip. I was able to remove `apt-get upgrade` without any problems.

dockerfiles/Dockerfile
schwarz-em: It feels like I should be able to, but the build runs into permissions issues without the `sudo` for some reason.

abejgonzalez:
Merging (followup PR's can clean it up some more if need be... but this is currently much cleaner than the current approach).

abejgonzalez (approved):

PR #774, opened 2021-01-23
Avoid permissions conflict on shared protocjar.webcache

jerryz123:
The Java protoc plugin (part of FIRRTL) by default caches protoc in /tmp, and does not delete it afterwards.
This causes permissions failures when multiple users try to install Chipyard on the same machine, due to permissions failures over accessing the same cache in /tmp.

A solution is to place the tmpdir within the Chipyard repo. No other files besides protocjar.webcache currently use that directory, so I believe this should be safe.

**Type of change**: bug fix

**Impact**: software change

### DIFF
--- a/.java_tmp/.gitignore
+++ b/.java_tmp/.gitignore
@@ -1,0 +1,4 @@
+# Ignore everything in this directory
+*
+# Except this file
+!.gitignore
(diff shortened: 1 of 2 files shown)


abejgonzalez:
Doesn't this require you to make the folder beforehand (before the first time `GenerateSimFiles` is run)?

jerryz123:
Oh yeah, I need to check in that directory as well

abejgonzalez (approved):
I'm not sure we want to enforce this on every user since this seems like a very UCB problem. Maybe we should add some documentation/issue pointing to this fix? With that being said, I think for now we can use this and if someone wants this removed we can revisit it (#agile-dev).

PR #778, opened 2021-01-31
Update comment on GenerateReset

### LINKED ISSUE #741: Update comment for chipyard.GenerateReset
https://github.com/ucb-bar/chipyard/blob/8836f84c79772e2cf9b8fc9118fb862ed2cce9a7/generators/chipyard/src/main/scala/Clocks.scala#L18-L22

@jerryz123 after #703 , the comment above is really confusing.  I just spent awhile looking around for what happened to `GlobalResetScheme` and it doesn't exist anymore.  We should update this comment or delete it.

jerryz123:
ChipTop reset was standardized to be async for 1.4.0
Resolves #741

### DIFF
--- a/generators/chipyard/src/main/scala/Clocks.scala
+++ b/generators/chipyard/src/main/scala/Clocks.scala
@@ -17,8 +17,8 @@ import chipyard.clocking._
 
 /**
  * A simple reset implementation that punches out reset ports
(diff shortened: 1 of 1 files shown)


colinschmidt (approved):
LGTM

PR #784, opened 2021-02-02
Add support for Digilent Genesys 2 board

yqszxx:
Depends on sifive/fpga-shells#159

**Related issue**: N/A

**Impact**: new fpga target

**Release Notes**

Add support for Digilent Genesys 2 board

### DIFF
--- a/fpga/src/main/scala/genesys2/IOBinders.scala
+++ b/fpga/src/main/scala/genesys2/IOBinders.scala
@@ -1,0 +1,74 @@
+package chipyard.fpga.genesys2
+
+import chipyard.CanHaveMasterTLMemPort
+import chipyard.iobinders.{GetSystemParameters, OverrideIOBinder, OverrideLazyIOBinder}
+import chisel3._
+import chisel3.experimental.{DataMirror, IO}
+import freechips.rocketchip.devices.debug._
+import freechips.rocketchip.diplomacy.InModuleBody
+import freechips.rocketchip.prci.{ClockSinkNode, ClockSinkParameters}
+import freechips.rocketchip.subsystem.BaseSubsystem
+import freechips.rocketchip.tilelink.TLBundle
+import freechips.rocketchip.util.{HeterogeneousBag, PSDTestMode}
+import sifive.blocks.devices.uart.HasPeripheryUARTModuleImp
+
+class WithUARTIOPassthrough extends OverrideIOBinder({
+  (system: HasPeripheryUARTModuleImp) => {
+    val io_uart_pins_temp = system.uart.zipWithIndex.map { case (dio, i) => IO(dio.cloneType).suggestName(s"uart_$i") }
+    (io_uart_pins_temp zip system.uart).foreach { case (io, sysio) =>
+      io <> sysio
+    }
+    (io_uart_pins_temp, Nil)
+  }
+})
+
+class WithJTAGIOPassthrough extends OverrideLazyIOBinder({
+  (system: HasPeripheryDebug) => {
+    implicit val p = GetSystemParameters(system)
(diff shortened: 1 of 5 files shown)


fpga/src/main/scala/genesys2/IOBinders.scala
```diff
@@ -0,0 +1,74 @@
...
+import chisel3.experimental.{DataMirror, IO}
+import freechips.rocketchip.devices.debug._
+import freechips.rocketchip.diplomacy.InModuleBody
+import freechips.rocketchip.prci.{ClockSinkNode, ClockSinkParameters}
+import freechips.rocketchip.subsystem.BaseSubsystem
+import freechips.rocketchip.tilelink.TLBundle
+import freechips.rocketchip.util.{HeterogeneousBag, PSDTestMode}
+import sifive.blocks.devices.uart.HasPeripheryUARTModuleImp
+
+class WithUARTIOPassthrough extends OverrideIOBinder({
```
abejgonzalez, line 15: This code seems like a copy of the VCU118 code. Maybe we can dedup this by creating a folder like: `fpga/src/main/scala/common` and putting both sources there (with a new package name `chipyard.fpga.common`). Then we just reference that.

abejgonzalez:
Awesome, thanks so much for the contribution! I added one comment to maybe reduce the amount of duplication between this and the VCU118 work. I think before we consider merging this into Chipyard, you should try to get your PR merged into `fpga-shells`.

With that being said, the Chipyard team is discussing how best to deal with a PR like this... mainly because we have no mechanisms for testing that this works on the Genesys2 board without having one ourselves. We will get back to you on that once we have a decision.

fpga/src/main/scala/genesys2/IOBinders.scala
yqszxx: Good idea, I noticed I was just copying and pasting code when porting too 😂. Should I try to do some dedup work now or after this PR get merged?

yqszxx:
> Awesome, thanks so much for the contribution! I added one comment to maybe reduce the amount of duplication between this and the VCU118 work. I think before we consider merging this into Chipyard, you should try to get your PR merged into `fpga-shells`.
> 
> With that being said, the Chipyard team is discussing how best to deal with a PR like this... mainly because we have no mechanisms for testing that this works on the Genesys2 board without having one ourselves. We will get back to you on that once we have a decision.

Testing is really a problem, looking forward to your discussion results~

fpga/src/main/scala/genesys2/IOBinders.scala
abejgonzalez: I would wait until we get back to you on the testing side and `fpga-shells` gets merged.

jerryz123:
Do we need the actual boards to do testing? Wouldn't it be sufficient to test that Vivado can elaborate the design?

It may be a unreasonable goal for us to maintain regressions for many FPGA boards anyways. It may be better to have the board support checked in anyways, with the caveat that it might require more tinkering to make work compared to other flows.

abejgonzalez:
> Do we need the actual boards to do testing? Wouldn't it be sufficient to test that Vivado can elaborate the design?
> 
> It may be a unreasonable goal for us to maintain regressions for many FPGA boards anyways. It may be better to have the board support checked in anyways, with the caveat that it might require more tinkering to make work compared to other flows.

Correct, we want to avoid the need for having a board. However, we want to have some sort of simulation flow setup that can run the tests with FPGA collateral provided by Xilinx (through Vivado etc). An example of this would be https://github.com/ucb-bar/chipyard/tree/arty-sim/fpga or FPGA-level sim in FireSim

PR #790, opened 2021-02-08
Passing MBus clock frequency to SimDRAM

jerryz123:
**Related issue**: 

**Type of change**: bug fix


generators/chipyard/src/main/scala/config/AbstractConfig.scala
```diff
@@ -45,6 +45,8 @@ class AbstractConfig extends Config(
   new chipyard.config.WithNoSubsystemDrivenClocks ++             // drive the subsystem diplomatic clocks from ChipTop instead of using implicit clocks
   new chipyard.config.WithInheritBusFrequencyAssignments ++      // Unspecified clocks within a bus will receive the bus frequency if set
   new chipyard.config.WithPeripheryBusFrequencyAsDefault ++      // Unspecified frequencies with match the pbus frequency (which is always set)
+  new chipyard.config.WithMemoryBusFrequency(100.0) ++           // Default 100 MHz mbus
```
colinschmidt, line 48: Does this affect the default config's simulation speed?

jerryz123: No. The default was already 100 MHz here, this just makes it more explicit.

jerryz123: The explicitness is necessary for the WithSimDRAM HarnessBinder to pick it up

davidbiancolin (approved):
Ideally there'd be a clock sink in the port you could query, instead of grabbing from p(MemoryBusKey), but i know that's easier said than done. 

LGTM

PR #791, opened 2021-02-09
Fix parsing of --ignore-qemu option

a0u:
**Related issue**: 

**Type of change**: bug fix

**Impact**: other

**Release Notes**

Remove an extra `shift` that breaks parsing of subsequent arguments.  The positional parameters are already shifted once at the end of the loop body.


abejgonzalez (approved):
Oops. LGTM pending CI

PR #807, opened 2021-02-26
Use "tile" instead of "core" to assign freq's in WithTileFrequency config. fragment

abejgonzalez:
**Related issue**: 

**Type of change**: bug fix

**Impact**: other

**Release Notes**
Fixes an issue where `WithTileFrequency` wouldn't modify the frequency of all tiles since clock names are based on the `TileParams` name (which defaults to `tile` and `boom_tile` for Rocket/BOOM respectively).

Tested with 2 configs:

1. `MulticlockRocketConfig`:
image

2. `DoubleRocketAndBoomConfig` (custom config that only used `WithTileFrequency(600)` - no other freq's changed):
image

### DIFF
--- a/generators/chipyard/src/main/scala/ConfigFragments.scala
+++ b/generators/chipyard/src/main/scala/ConfigFragments.scala
@@ -193,7 +193,7 @@ class WithTLBackingMemory extends Config((site, here, up) => {
   case ExtTLMem => up(ExtMem, site) // enable TL backing memory
 })
 
-class WithTileFrequency(fMHz: Double) extends ClockNameContainsAssignment("core", fMHz)
+class WithTileFrequency(fMHz: Double) extends ClockNameContainsAssignment("tile", fMHz)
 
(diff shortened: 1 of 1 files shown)


jerryz123:
I remember this was done because it defaults naming everything to `core_{hartid}`, done here: https://github.com/chipsalliance/rocket-chip/blob/8f9149faf01fec9725d08c31c9c0bb5341f5e999/src/main/scala/subsystem/HasTiles.scala#L251

Was there a change somewhere that changed the naming to `tile`?

abejgonzalez:
I saw that... The `getOrElse` is getting what `<Rocket/Boom>TileParams` gives: https://github.com/chipsalliance/rocket-chip/blob/8f9149faf01fec9725d08c31c9c0bb5341f5e999/src/main/scala/tile/RocketTile.scala#L25. I'm not sure how this worked before...

jerryz123 (approved):
Approved, but @davidbiancolin should take a look as well

davidbiancolin:
It was named core before. This LGTM.

PR #809, opened 2021-03-01
Add WithMultiRoCCFromBuildRoCC to make heterogeneous accelerator configs easier

jerryz123:
This fragment basically lets you copy the contents of the BuildRoCC key into a MultiRoCC hart mapping, so you no longer have to define custom `WithMultiRoCCX` fragments for each accelerator you want to use with MultiRoCC


abejgonzalez (approved):
LGTM. Do we want to remove the other approach and re-work the examples?

jerryz123:
Let's leave that stuff in for now.

PR #812, opened 2021-03-03
AXI Memory Over TL Serial Link

abejgonzalez:
**Related issue**: https://github.com/ucb-bar/testchipip/pull/121

**Impact**: rtl change

**Release Notes**
This PR adds support for simulating an AXI memory interface over the default TL serial link in Chipyard/FireSim.

- Add `WithSerialTLBackingMemory` config frag. to move `ExtMem` params to `SerialTLKey`
- New harness binders to connect the Serial Link + Passthrough Clock (for AXI clock domain)
- New example configs to demonstrate setup (`MulticlockAXIOverSerialConfig`, `FireSimMulticlockAXIOverSerialConfig`)
- Modified config. fragment to specify tile clocks (overall or just by hart id)
- Split FireSim "tweaks" into 2 fragments... without clocking information and overall. This lets users ignore FireSim clock defaults if they want.
- Add support in both Chipyard and FireSim to request a clock in the harness (and harness binder). Chipyard provides both a clock/reset while FireSim just provides a clock. Note that (https://github.com/ucb-bar/chipyard/pull/812#issuecomment-791815160) applies to both FireSim and Chipyard's test harnesses.

This was tested in FireSim @ 30MHz with the added config.


generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -260,6 +260,48 @@ class WithSerialTLIOCells extends OverrideIOBinder({
...
+    InModuleBody {
+      val port = IO(new SerialAndPassthroughClockResetIO(sVal.width)).suggestName(s"serial_tl_passthrough_clk")
+      port.clocked_serial <> serial_io
+      port.passthrough_clock_reset <> clockBundle
+
+      // return the ports and no IO cells
+      (Seq(port), Nil)
+    }
+  }).getOrElse(InModuleBody{(Nil, Nil)}).asInstanceOf[ModuleValue[IOBinderTuple]]
+})
```
jerryz123, line 304: Can we avoid punching out the clock here? This breaks the "Chip Config" vs "Platform Config" abstraction, since this IOBinder affects chiptop hardware, but should never appear on an actual chip. 

I realize there's no other easy way to pass the appropriately divided clock out to the harness aximem, but for SW RTL simulation, timing accuracy was never a primary goal anyways. I would much prefer that the RTL simulations use the Harness clock for the sake of preserving this abstraction, even if that is totally unrealistic compared to the chip clock.

generators/chipyard/src/main/scala/config/RocketConfigs.scala
```diff
@@ -212,3 +213,26 @@ class LBWIFRocketConfig extends Config(
...
+  new chipyard.config.WithFrontBusFrequency(4000 / 2) ++
+
+  new chipyard.config.WithFbusToSbusCrossingType(RationalCrossing(SlowToFast)) ++
+  new testchipip.WithAsynchronousSerialSlaveCrossing ++
+
+  new chipyard.harness.WithSimAXIMemOverSerialTL ++ // add SimDRAM DRAM model for axi4 backing memory over the SerDes link, if axi4 mem is enabled
+  new chipyard.iobinders.WithSerialTLAndPassthroughClockPunchthrough ++ // add new clock for axi domain over serdes and passthrough ios
+  new chipyard.config.WithSerialTLBackingMemory ++ // remove axi4 mem port in favor of SerialTL memory
+  new testchipip.WithBlockDeviceLocations(
+    freechips.rocketchip.subsystem.PBUS,
```
jerryz123, line 234: Does this work? The block device can't master memory if both edges connect to the pbus

generators/chipyard/src/main/scala/config/RocketConfigs.scala
```diff
@@ -212,3 +213,26 @@ class LBWIFRocketConfig extends Config(
   new freechips.rocketchip.subsystem.WithNoMemPort ++       // remove AXI4 backing memory
   new freechips.rocketchip.subsystem.WithNBigCores(1) ++
   new chipyard.config.AbstractConfig)
+
+class MulticlockAXIOverSerialConfig extends Config(
+  new chipyard.config.WithFbusToSbusCrossingType(RationalCrossing(SlowToFast)) ++
+
+  new chipyard.config.WithSystemBusFrequencyAsDefault ++
+  new chipyard.config.WithSystemBusFrequency(4000) ++
+  new chipyard.config.WithPeripheryBusFrequency(4000) ++
```
jerryz123, line 222: Why 4000?

generators/firechip/src/main/scala/TargetConfigs.scala
```diff
@@ -99,6 +82,23 @@ class WithFireSimConfigTweaks extends Config(
   new chipyard.config.WithNoDebug
 )
 
+// Tweaks that are generally applied to all firesim configs
+class WithFireSimConfigTweaks extends Config(
+  // Optional*: Removing this will require adjusting the UART baud rate and
```
jerryz123, line 87: Like this change.
Maybe call this `WithFireSimClockingTweaks`, and keep the name of the old one?

generators/chipyard/src/main/scala/ConfigFragments.scala
```diff
@@ -193,7 +193,20 @@ class WithTLBackingMemory extends Config((site, here, up) => {
...
+    memParams = {
+      val memPortParams = up(ExtMem, site).get
+      require(memPortParams.nMemoryChannels == 1)
+      memPortParams.master
+    },
+    isMemoryDevice = true
+  )}
+})
+
+class WithTileFrequency(fMHz: Double) extends ClockNameContainsAssignment("tile", fMHz)
```
jerryz123, line 208: Instead of adding another fragment, maybe add an optional parameter which, if set, specifies a hart, if not set, specifies all harts?

generators/firechip/src/main/scala/TargetConfigs.scala
```diff
@@ -204,6 +204,14 @@ class FireSimMulticlockRocketConfig extends Config(
   new freechips.rocketchip.subsystem.WithRationalRocketTiles ++   // Add rational crossings between RocketTile and uncore
   new FireSimRocketConfig)
 
+class FireSimMulticlockAXIOverSerialConfig extends Config(
+  new WithAXIOverSerialTLCombinedBridges ++ // use combined bridge to connect to axi mem over serial
+  new WithDefaultFireSimBridges ++
```
jerryz123, line 209: Probably need to remove blockdevice from this config.

jerryz123 (changes requested):

generators/firechip/src/main/scala/TargetConfigs.scala
jerryz123: Actually, its probably safer to rename the old one as well, `WithFireSimDesignTweak`, perhaps?

generators/chipyard/src/main/scala/config/RocketConfigs.scala
abejgonzalez: This doesn't work so far. I hadn't really looked into why the block device is failing past this though. As you mentioned, I just decided to remove the block device in an unpushed commit (more closely matches a chip config where there would instead be a "SPIBridge" of some sorts).

generators/chipyard/src/main/scala/IOBinders.scala
abejgonzalez: I'm trying some workarounds for this.

generators/chipyard/src/main/scala/config/RocketConfigs.scala
abejgonzalez: Arbitrary. I just did this so things were nice multiples of 1000.

generators/chipyard/src/main/scala/IOBinders.scala
abejgonzalez: Ok sketched out a new implementation. Things to note:

- I still need to fix FireSim to work in a similar fashion
- Is there a nicer way to tell the harness the reference clock frequency of the chip (instead of using `ReferenceClockTracker`)
- Right now there is a difference between the `harness<Clock/Reset>` and the implicit harness clock and reset. I need to fix this...

generators/chipyard/src/main/scala/Clocks.scala
```diff
@@ -47,6 +47,12 @@ case object ClockingSchemeKey extends Field[ChipTop => Unit](ClockingSchemeGener
   */
 case object ClockFrequencyAssignersKey extends Field[Seq[(String) => Option[Double]]](Seq.empty)
 case object DefaultClockFrequencyKey extends Field[Double]()
+case object ReferenceClockTrackerKey extends Field[ReferenceClockTracker](new ReferenceClockTracker)
+class ReferenceClockTracker {
```
jerryz123, line 51: Using a `Field` to pass values between Module and Harness feels wrong.
Perhaps just set harnessFreq as a property of ChipTop, and then in Harness, do
```
val harnessFreq = lazyDut match { 
  case d: ChipTop => d.harnessFreq
  case _ => // some default
}
```

generators/chipyard/src/main/scala/Clocks.scala
```diff
@@ -86,20 +92,22 @@ object ClockingSchemeGenerators {
...
 
-
     InModuleBody {
       val clock_wire = Wire(Input(Clock()))
       val reset_wire = GenerateReset(chiptop, clock_wire)
       val (clock_io, clockIOCell) = IOCell.generateIOFromSignal(clock_wire, "clock")
       chiptop.iocells ++= clockIOCell
 
+      p(ReferenceClockTrackerKey).set(dividerOnlyClkGenerator.module.referenceFreq)
+
```
jerryz123, line 110: It might be reasonable to make the ClockingSchemeGenerator function signature return the expected frequency of the harness clock, then it can be set directly in ChipTop, where the ClockingSchemeGenerator is called.

generators/chipyard/src/main/scala/HarnessBinders.scala
```diff
@@ -83,7 +83,9 @@ class WithGPIOTiedOff extends OverrideHarnessBinder({
 // DOC include start: WithUARTAdapter
 class WithUARTAdapter extends OverrideHarnessBinder({
   (system: HasPeripheryUARTModuleImp, th: HasHarnessSignalReferences, ports: Seq[UARTPortIO]) => {
-    UARTAdapter.connect(ports)(system.p)
+    withClockAndReset(th.harnessClock, th.harnessReset) {
+      UARTAdapter.connect(ports)(system.p)
```
jerryz123, line 87: Was this necessary for correct behavior? Curious

generators/chipyard/src/main/scala/TestHarness.scala
```diff
@@ -25,6 +29,47 @@ trait HasHarnessSignalReferences {
   def success: Bool
 }
 
+class HarnessClockInstantiator {
+  private var _clockMap: HashMap[String, (Double, ClockBundle)] = HashMap.empty
+
```
jerryz123, line 34: In the future we might have multiple clocks in the harness as well, so this should generalize? Pretty cool

generators/chipyard/src/main/scala/config/RocketConfigs.scala
jerryz123: If its arbitrary, can we choose realistic frequencies for these?

generators/chipyard/src/main/scala/IOBinders.scala
jerryz123: I suggested a way to avoid `ReferenceClockTracker` in another comment.

generators/chipyard/src/main/scala/TestHarness.scala
abejgonzalez: Yup. This in theory should be able to provide clocks to anything in the harness that wants it (in theory because I haven't tested that use case past the current AXI domain stuff)

generators/chipyard/src/main/scala/HarnessBinders.scala
abejgonzalez: Technically no... but in a previous commit there was a difference between the implicit harness clock/reset and the `th.harness<Clock/Reset>`. This should be fixed now so I reverted this.

generators/chipyard/src/main/scala/Clocks.scala
abejgonzalez: Done

generators/chipyard/src/main/scala/Clocks.scala
abejgonzalez: I just used a `var` variable that I could reference.

abejgonzalez:
There is a distinction that needs to be made in the harness with these changes:

`harnessClock/harnessReset` =/= `clock/reset` in the TestHarness

In the case that the reference clock coming from `ChipTop` isn't the overall reference clock of the entire simulation (i.e. isn't the clock/reset coming into the TestHarness module), the `harnessClock/harnessReset` can differ from the implicit `TestHarness` `clock/reset`. Most of the time this won't be the case... but the current `MulticlockAXIOverSerialConfig` exposes this behavior. Here is a snippet of the clock summaries:

image

Here you can see that the reference clock for `ChipTop` is 500MHz but the reference clock for the entire system is actually 1GHz.

generators/firechip/src/main/scala/TargetConfigs.scala
abejgonzalez: This `WithFireSimConfigTweaks` combines the clocking with the "design" tweaks. In order to avoid changing FireSim itself (changing the build recipes), I will at least keep this name `WithFireSimConfigTweaks`.

jerryz123:
In the example you provide, the `mem_over_serial_tl_clock` is the clock for the AXI memory?

abejgonzalez:
> In the example you provide, the `mem_over_serial_tl_clock` is the clock for the AXI memory?

Yup that's correct.

alonamid:
Will this setup come with a docs section under the "Advanced Concepts" section?

abejgonzalez:
> Will this setup come with a docs section under the "Advanced Concepts" section?

Sure. Though I think I might re-work some of the "Communicating with the DUT" section a bit instead of adding a completely new page.

docs/Advanced-Concepts/Chip-Communication.rst
```diff
@@ -130,13 +118,81 @@ Using the JTAG Interface
...
+Most, if not all, Chipyard configurations are tethered using TSI (over a serial link) and have access
+to external memory through an AXI port (backing AXI memory).
+The following image shows the DUT with these set of default signals:
+
+.. image:: ../_static/images/default-chipyard-config-communication.png
+
+In this setup, the serial-link is connected to the TSI/FESVR peripherals while the AXI port is connected
+to a simulated AXI memory.
+However, AXI ports tend to have many signals associated with them so instead of creating an AXI port off the DUT,
+one can send the memory transactions over the bi-directional serial-link (``TLSerdesser``) so that main
```
alonamid, line 156:
```suggestion
one can send the memory transactions over the bi-directional serial-link (``TLSerdesser``) so that the main
```

docs/Advanced-Concepts/Chip-Communication.rst
```diff
@@ -130,13 +118,81 @@ Using the JTAG Interface
...
+interface to the DUT is the serial-link (which as comparatively less signals than an AXI port).
```
alonamid, line 157:
```suggestion
interface to the DUT is the serial-link (which has comparatively less signals than an AXI port).
```

docs/Advanced-Concepts/Chip-Communication.rst
```diff
@@ -130,13 +118,81 @@ Using the JTAG Interface
...
 
 Debugging with JTAG
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+~~~~~~~~~~~~~~~~~~~
 
-Please refer to the following resources on how to debug with JTAG.
+Please refer to the following resources on how to debug a Rocket Chip system with JTAG.
 
 * https://github.com/chipsalliance/rocket-chip#-debugging-with-gdb
 * https://github.com/riscv/riscv-isa-sim#debugging-with-gdb
```
alonamid, line 131: Should we move these lines to be after step 2, so people don't jump directly to rocketchip and try to go by the full rocketchip instructions, but rather start by building the chipyard configs and running the chipyard simulation, and then using the rocketchip instructions only for OpenOCD and GDB instructions?

abejgonzalez: I reworked this a bit. Check now.

generators/firechip/src/main/scala/TargetConfigs.scala
```diff
@@ -87,10 +74,6 @@ class WithFireSimConfigTweaks extends Config(
   new chipyard.config.WithTraceIO ++
   // Optional: Request 16 GiB of target-DRAM by default (can safely request up to 32 GiB on F1)
   new freechips.rocketchip.subsystem.WithExtMemSize((1 << 30) * 16L) ++
-  // Required: Adds IO to attach SerialBridge. The SerialBridges is responsible
-  // for signalling simulation termination under simulation success. This fragment can
-  // be removed if you supply an auxiliary bridge that signals simulation termination
-  new testchipip.WithDefaultSerialTL ++
```
abejgonzalez, line 93: Removed since this already is default in Chipyard configs

generators/chipyard/src/main/scala/TestHarness.scala
```diff
@@ -19,32 +23,96 @@ trait HasTestHarnessFunctions {
...
 trait HasHarnessSignalReferences {
+  // clock/reset of the chiptop reference clock (can be different than the implicit harness clock/reset)
   def harnessClock: Clock
   def harnessReset: Reset
   def dutReset: Reset
   def success: Bool
 }
 
+class HarnessClockInstantiator {
+  private var _clockMap: HashMap[String, (Double, ClockBundle)] = HashMap.empty
```
colinschmidt, line 34: I think in general it is better to use a SortedMap that keeps orderings when building Chisel hardware to keep names stable.
See: https://github.com/chipsalliance/rocket-chip/commit/f86489b59e82c4ae4e39b6ba0caf9fd279d4aa16 for example

I can't quite tell if this will be an issue in this case but I don't think it will hurt.

generators/chipyard/src/main/scala/TestHarness.scala
```diff
@@ -19,32 +23,96 @@ trait HasTestHarnessFunctions {
...
+
+  // request a clock bundle at a particular frequency
+  def getClockBundle(name: String, freqRequested: Double): ClockBundle = {
```
colinschmidt, line 37: I think this should be named `requestClockBundle` as it looks much more like a setter than a getter which the `getClockBundle` name suggested to me at first. I had to read the comment to understand but if you change the name the comment basically isn't needed as its now obvious from the name.

generators/chipyard/src/main/scala/TestHarness.scala
```diff
@@ -19,32 +23,96 @@ trait HasTestHarnessFunctions {
...
+  // connect all clock wires specified to a divider only PLL
+  def instantiateHarnessDividerPLL(refClock: ClockBundle): Unit = {
+    val sinks = _clockMap.map({ case (name, (freq, bundle)) =>
+      ClockSinkParameters(take=Some(ClockParameters(freqMHz=freq/1000000)),name=Some(name))
+    }).toSeq
+
+    val pllConfig = new SimplePllConfiguration("harnessDividerOnlyClockGenerator", sinks)
+    pllConfig.emitSummaries()
+
+    val dividedClocks = HashMap[Int, Clock]()
```
colinschmidt, line 52: I think this HashMap is ok because we don't iterate over it to create hardware.

generators/chipyard/src/main/scala/TestHarness.scala
```diff
@@ -19,32 +23,96 @@ trait HasTestHarnessFunctions {
...
     val success = Output(Bool())
   })
 
+  val harnessClock = Wire(Clock())
+  val harnessReset = Wire(Reset())
+
   val lazyDut = LazyModule(p(BuildTop)(p)).suggestName("chiptop")
-  val dut = Module(lazyDut.module)
+  withClockAndReset(harnessClock, harnessReset) {
+    val dut = Module(lazyDut.module)
```
colinschmidt, line 90: If this is just to avoid having an implicit clock we can change the class to be a `BareModule`.
Otherwise its confusing to me to be passing harnessClock to the dut.

generators/firechip/src/main/scala/BridgeBinders.scala
```diff
@@ -98,7 +100,59 @@ class WithUARTBridge extends OverrideHarnessBinder({
 class WithBlockDeviceBridge extends OverrideHarnessBinder({
   (system: CanHavePeripheryBlockDevice, th: FireSim, ports: Seq[ClockedIO[BlockDeviceIO]]) => {
     implicit val p: Parameters = GetSystemParameters(system)
-    ports.map { b => BlockDevBridge(b.clock, b.bits, th.harnessReset.toBool) }
+    ports.map { b => BlockDevBridge(b.clock, b.bits, th.harnessReset.asBool) }
+    Nil
```
colinschmidt, line 104: What was the effect of leaving this off? Was this code just not exercised before?

generators/firechip/src/main/scala/BridgeBinders.scala
```diff
@@ -98,7 +100,59 @@ class WithUARTBridge extends OverrideHarnessBinder({
...
+
+        val harnessMultiClockAXIRAM = SerialAdapter.connectHarnessMultiClockAXIRAM(
+          system.serdesser.get,
+          port,
+          axiClockBundle,
+          th.harnessReset)
+        SerialBridge(port.clock, harnessMultiClockAXIRAM.module.io.tsi_ser, Some(MainMemoryConsts.globalName))
+
+        // connect SimAxiMem
+        (harnessMultiClockAXIRAM.mem_axi4 zip harnessMultiClockAXIRAM.memNode.edges.in).map { case (axi4, edge) =>
```
colinschmidt, line 140: I don't like how the first 30 lines of this class is pure duplication with the non-firesim version.

generators/firechip/src/main/scala/FireSim.scala
```diff
@@ -38,44 +40,113 @@ object NodeIdx {
 /**
   * Under FireSim's current multiclock implementation there can be only a
   * single clock bridge. This requires, therefore, that it  be instantiated in
-  * the harness and reused across all supernode instances. This class attempts to 
+  * the harness and reused across all supernode instances. This class attempts to
   * memoize its instantiation such that it can be referenced from within a ClockScheme function.
   */
 class ClockBridgeInstantiator {
-  private var _clockRecord: Option[RecordMap[Clock]] = None
+  private var _harnessClockMap: HashMap[String, (Double, Clock)] = HashMap.empty
```
colinschmidt, line 47: I think you iterate over these two hashmaps later but I can't tell if that will affect the verilog naming or not.

colinschmidt (commented):
The idea of this PR is to test something closer to the bringup infrastructure? This will let us model the performance of those systems better but not build them easier right?
Like overriding `WithSimAXIMemOverSerialTL` with something that actually connects to an FPGA DRAM system wouldn't help because you are still building a chip and not the FPGA component.

generators/chipyard/src/main/scala/TestHarness.scala
jerryz123: Yeah don't do this. The dut already creates a harnessFunction to wire its clock/reset, we don't want to make it seem like this relies on implicit behavior.

generators/chipyard/src/main/scala/clocking/DividerOnlyClockGenerator.scala
```diff
@@ -146,5 +147,5 @@ class DividerOnlyClockGenerator(pllName: String)(implicit p: Parameters, valName
 }
 
 object DividerOnlyClockGenerator {
-  def apply()(implicit p: Parameters, valName: ValName) = LazyModule(new DividerOnlyClockGenerator(valName.name)).node
+  def apply()(implicit p: Parameters, valName: ValName) = LazyModule(new DividerOnlyClockGenerator(valName.name))
```
jerryz123, line 150: Just delete this helper object if we have to reference the module after construction

abejgonzalez:
> The idea of this PR is to test something closer to the bring-up infrastructure? This will let us model the performance of those systems better but not build them easier right?

That's correct, this PR is trying to model a setup where backing memory is over the serial link which is really only used in test chips. This is mainly for performance reasons... as for building a test chip this doesn't help too much (the test chip builder still has to properly setup the FPGA bringup infrastructure, remove the AXI port, etc).

generators/chipyard/src/main/scala/ConfigFragments.scala
```diff
@@ -205,7 +205,25 @@ class WithTLBackingMemory extends Config((site, here, up) => {
   case ExtTLMem => up(ExtMem, site) // enable TL backing memory
 })
 
-class WithTileFrequency(fMHz: Double) extends ClockNameContainsAssignment("tile", fMHz)
+class WithSerialTLBackingMemory extends Config((site, here, up) => {
+  case ExtMem => None
+  case SerialTLKey => up(SerialTLKey, site).map { k => k.copy(
+    memParams = {
+      val memPortParams = up(ExtMem, site).get
+      require(memPortParams.nMemoryChannels == 1)
```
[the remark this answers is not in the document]
abejgonzalez: Nope, this was the original functionality. I just made this clearer by adding a `require`.

generators/firechip/src/main/scala/BridgeBinders.scala
jerryz123: One solution is to make a generic harness binder accept a function as a parameter which connects the AXI IO in the harness to wherever it needs to go.

generators/firechip/src/main/scala/BridgeBinders.scala
abejgonzalez: I'm not sure I follow. I think the only thing that changed here was that I converted `toBool` to `asBool` (the diff makes this a bit hard to see). If you are referring to removing the block device... there is no block device on the test chip so I removed it (forcing people to use the initramfs linux version).

colinschmidt: I was referring to adding `Nil`.

generators/firechip/src/main/scala/FireSim.scala
abejgonzalez: I never iterate over the hashmaps to determine Verilog signals. I only use it to "store" requests (clock wires) and later connect them to the clock bridge.

generators/firechip/src/main/scala/BridgeBinders.scala
abejgonzalez: `Nil` was already added before. The diff makes this hard to see. https://github.com/ucb-bar/chipyard/pull/812/files#diff-1cd9711add00014e504e37e13d8aa69c28654c356c53e6b84bddd104c11a4763L102

colinschmidt: oh wow. that is hard to see. Thanks

davidbiancolin (commented):
I'll look through the FireSim stuff later.

generators/firechip/src/main/scala/TargetConfigs.scala
```diff
@@ -59,26 +59,13 @@ class WithNIC extends icenet.WithIceNIC(inBufFlits = 8192, ctrlQueueDepth = 64)
 class WithNVDLALarge extends nvidia.blocks.dla.WithNVDLA("large")
 class WithNVDLASmall extends nvidia.blocks.dla.WithNVDLA("small")
 
-
-// Tweaks that are generally applied to all firesim configs
-class WithFireSimConfigTweaks extends Config(
+class WithFireSimDesignTweaks extends Config(
```
davidbiancolin, line 62: Add a comment here explaining in greater detail what set of tweaks this contains.

generators/firechip/src/main/scala/TargetConfigs.scala
davidbiancolin: Could  you make a `WithFireSimDefaultFrequencies` or something and compose the two of them to produce the old tweaks.

generators/firechip/src/main/scala/FireSim.scala
```diff
@@ -38,44 +40,113 @@ object NodeIdx {
...
+
+  // Assumes that the supernode implementation results in duplicated clocks
+  //   (i.e. only 1 set of clocks is generated for all BuildTop designs)
+  private var _ratClockMap: HashMap[String, (RationalClock, Clock)] = HashMap.empty
```
davidbiancolin, line 51: This can and should be a val since the data structure itself mutable. I think i'd also prefer if you used a `mutable.LinkedHashMap` just to future proof it against someone iterating on it to generate HW.

davidbiancolin: Same for the structure above.

davidbiancolin: Also these two structures are analogs for harness and chip clocks right? If so, can you make the names reflect that by renaming _ratClockMap to _chipClockMap or something?

davidbiancolin (commented):
Mostly just nits.

generators/chipyard/src/main/scala/TestHarness.scala
abejgonzalez: I think `Double` is fine. However, the DTS frequency associated with the buses uses a `BigInt` (https://github.com/chipsalliance/rocket-chip/blob/ca2567ebc200e7027cbbe8c2d3e70378c6bab445/src/main/scala/tilelink/BusWrapper.scala#L26-L27). IIRC there are other places that use `Int` and `BigInt` as well... I can try going through and enforcing the `Double` though.

generators/chipyard/src/main/scala/TestHarness.scala
abejgonzalez: Huh. Ok. Ill double check and change this.

generators/firechip/src/main/scala/FireSim.scala
abejgonzalez: Seems like most people care about ordered maps so Ill change this.

generators/chipyard/src/main/scala/clocking/DividerOnlyClockGenerator.scala
abejgonzalez: I'll probably just delete the helper object.

generators/chipyard/src/main/scala/ChipTop.scala
```diff
@@ -15,6 +15,9 @@ import barstools.iocell.chisel._
 
 case object BuildSystem extends Field[Parameters => LazyModule]((p: Parameters) => new DigitalTop()(p))
 
+trait HasReferenceClockFreq {
+  var refClockFreqMHz: Option[Double] = None
```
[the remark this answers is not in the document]
abejgonzalez: I had alot of trouble making this a `def` since the `def` has to be defined within the clock scheme function... Maybe I'm missing something but can you assign a `def` inside a function?

generators/firechip/src/main/scala/FireSim.scala
abejgonzalez: Yup. I can rename things better.

generators/firechip/src/main/scala/TargetConfigs.scala
abejgonzalez: Wait Im actually a bit confused what you mean by this. There are already detailed comments next to each line on what things do.

generators/chipyard/src/main/scala/ChipTop.scala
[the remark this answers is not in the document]
jerryz123: You can make the clock scheme function return the freq.
So in ChipTop you could have 
```
val refClockFreqMHz = p(ClockingSchemeKey)(this)
```
`val` can override `def` in Scala.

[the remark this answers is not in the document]
abejgonzalez: I'm aware of that... I was hitting some snags with the frequency determined in the module body and passing that out of the clock function. I'll try hacking away to find a solution.

abejgonzalez:
Ok this PR is ready for a final review. One thing to note though. If not using the `load_mem` feature in FireSim, loading a binary is extremely slow (have to send the binary over the serial link to the chip, then back over the serial link to DRAM). While this is slow to begin with, this is exacerbated by FESVR step size in FireSim being very very large (https://github.com/firesim/firesim/blob/7e32128bea5390ca528052dcdffb13656ee41162/sim/firesim-lib/src/main/cc/bridges/serial.cc#L20) which slows down the binary loading. A future PR can address this since this PR is fairly big to begin with.

jerryz123:
This LGTM, although I think there can be some more deduplication that we can do later.

docs/Advanced-Concepts/Chip-Communication.rst
```diff
@@ -129,14 +116,110 @@ Then you can run simulations with the new DMI-enabled top-level and test-harness
...
+.. literalinclude:: ../../generators/chipyard/src/main/scala/config/RocketConfigs.scala
+    :language: scala
+    :start-after: DOC include start: MulticlockAXIOverSerialConfig
+    :end-before: DOC include end: MulticlockAXIOverSerialConfig
+
+Bringup Setup of the Example Test Chip after Tapeout
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Assuming this example test chip is taped out and now ready to be tested, we can communicate with the chip using this serial-link.
+For example, some Berkeley tapeouts of Chipyard chips have an FPGA running a RISC-V soft-core that is able to speak to the DUT.
```
alonamid, line 213:
```suggestion
For example, a common test setup used at Berkeley to evaluate Chipyard-based test-chips includes an FPGA running a RISC-V soft-core that is able to speak to the DUT (over an FMC).
```

alonamid (approved):
LGTM (minor nit comment)

generators/firechip/src/main/scala/TargetConfigs.scala
davidbiancolin: I had to unresolve this.

docs/Advanced-Concepts/Chip-Communication.rst
```diff
@@ -45,33 +45,21 @@ Using the Tethered Serial Interface (TSI)
...
-chip, it is deserialized and masters a bus on the chip. The following image shows this flow:
-
-.. image:: ../_static/images/chip-bringup.png
-
-.. note::
-    The ``TLSerdesser`` can also be used as a slave (client), so it can sink memory requests from the chip
-    and connect to off-chip backing memory. Or in other words, ``TLSerdesser`` creates a bi-directional TileLink
-    interface.
+TSI protocol is an implementation of HTIF that is used to send commands to the RISC-V DUT.
+These TSI commands are simple R/W commands that are able to probe the DUT's memory space.
```
colinschmidt, line 49: probe->access

docs/Advanced-Concepts/Chip-Communication.rst
```diff
@@ -45,33 +45,21 @@ Using the Tethered Serial Interface (TSI)
...
+This ``SimSerial`` Verilog module then sends the TSI command recieved by the simulation stub
+to an adapter that converts the TSI command into a TileLink request.
+This conversion is done by the ``SerialAdapter`` module (located in the ``generators/testchipip`` project).
+After the transaction is converted to TileLink, the ``TLSerdesser`` (located in ``generators/testchipip``) serializes the
+transaction and sends it to the chip (this ``TLSerdesser`` is sometimes also referred to as a digital serial-link or SerDes).
+Once the serialized transaction is received on the chip, it is deserialized and masters a TileLink bus on the chip
+which handles the request.
+In simulation, FESVR resets the DUT, writes into memory the test program, and indicates to the DUT to start the program
+through an interrupt (see :ref:`customization/Boot-Process:Chipyard Boot Process`).
+Using TSI is currently the fastest mechanism to communicate with the DUT in simulation and is also used by FireSim.
```
colinschmidt, line 62: If this is in comparison to DTM/JTAG it might be worth mentioning that? Also not sure if it is worth talking about the width here.

docs/Advanced-Concepts/Chip-Communication.rst
```diff
@@ -129,14 +116,110 @@ Then you can run simulations with the new DMI-enabled top-level and test-harness
...
+
+    make CONFIG=RocketConfig
+
+2. Run the simulation with remote bit-bang enabled. Since we hope to load/run the binary using JTAG,
+   we can pass ``none`` as a binary (prevents FESVR from loading the program). (Adapted from: https://github.com/chipsalliance/rocket-chip#3-launch-the-emulator)
+
+.. code-block:: bash
+
+    # note: this uses Chipyard make invocation to run the simulation to properly wrap the simulation args
+    make CONFIG=RocketConfig BINARY=none SIM_FLAGS="+jtag_rbb_enable=1 --rbb-port=9823" run-binary
```
colinschmidt, line 153: Great!

docs/Advanced-Concepts/Chip-Communication.rst
```diff
@@ -129,14 +116,110 @@ Then you can run simulations with the new DMI-enabled top-level and test-harness
...
+    to that documentation: `Rocket Chip GDB Docs <https://github.com/chipsalliance/rocket-chip#-debugging-with-gdb>`__,
+    `riscv-isa-sim GDB Docs <https://github.com/riscv/riscv-isa-sim#debugging-with-gdb>`__
+
+Example Test Chip Bringup Communication
+---------------------------------------
+
+Intro to Typical Chipyard Test Chip
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Most, if not all, Chipyard configurations are tethered using TSI (over a serial-link) and have access
```
colinschmidt, line 168: I'd like this sentence to make it clear that having TSI doesn't mean all operation will be tethered, but I don't know how to do that.

docs/Advanced-Concepts/Chip-Communication.rst
```diff
@@ -129,14 +116,110 @@ Then you can run simulations with the new DMI-enabled top-level and test-harness
...
+to external memory through an AXI port (backing AXI memory).
+The following image shows the DUT with these set of default signals:
+
+.. image:: ../_static/images/default-chipyard-config-communication.png
+
+In this setup, the serial-link is connected to the TSI/FESVR peripherals while the AXI port is connected
+to a simulated AXI memory.
+However, AXI ports tend to have many signals associated with them so instead of creating an AXI port off the DUT,
```
colinschmidt, line 176: "many signals, and thus wires,"

docs/Advanced-Concepts/Chip-Communication.rst
```diff
@@ -129,14 +116,110 @@ Then you can run simulations with the new DMI-enabled top-level and test-harness
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Assuming this example test chip is taped out and now ready to be tested, we can communicate with the chip using this serial-link.
+For example, a common test setup used at Berkeley to evaluate Chipyard-based test-chips includes an FPGA running a RISC-V soft-core that is able to speak to the DUT (over an FMC).
+This RISC-V soft-core would serve as the host of the test that will run on the DUT.
+This is done by the RISC-V soft-core running FESVR, sending TSI commands to a ``SerialAdapter`` / ``TLSerdesser`` programmed on the FPGA.
+Once the commands are converted to serialized TileLink, then they can be sent over some medium to the DUT
+(like an FMC cable or a set of wires connecting FPGA outputs to the DUT board).
+Similar to simulation, if the chip requests offchip memory, it can then send the transaction back over the serial-link.
+Then the request can be serviced by the channel of FPGA DRAM.
```
colinschmidt, line 219:
```suggestion
Then the request can be serviced by the FPGA DRAM.
```

docs/Advanced-Concepts/Harness-Clocks.rst
```diff
@@ -0,0 +1,38 @@
...
+    :start-after: DOC include start: HarnessClockInstantiatorEx
+    :end-before: DOC include end: HarnessClockInstantiatorEx
+
+Here you can see the ``p(HarnessClockInstantiatorKey)`` is used to request a clock and reset at ``memFreq`` frequency.
+
+.. note::
+    In the case that the reference clock entering ``ChipTop`` is not the overall reference clock of the simulation
+    (i.e. not the clock/reset coming into the ``TestHarness`` module), the ``harnessClock`` and ``harnessReset`` can
+    differ from the implicit ``TestHarness`` clock and reset. For example, if the ``ChipTop`` reference is 500MHz but an
+    extra harness clock is requested at 1GHz, the ``TestHarness`` implicit clock/reset will be at 1GHz while the ``harnessClock``
```
colinschmidt, line 37: This sentence is confusing to me. I would have expected it to end with "while the reference clock entering ChipTop will be at 500MHz."
If it is currently equivalent to that perhaps we should change the name of those signals.

jerryz123 (approved):

generators/chipyard/src/main/scala/ChipTop.scala
```diff
@@ -24,15 +27,16 @@ case object BuildSystem extends Field[Parameters => LazyModule]((p: Parameters)
...
   // The system module specified by BuildSystem
   lazy val lazySystem = LazyModule(p(BuildSystem)(p)).suggestName("system")
 
-  // The implicitClockSinkNode provides the implicit clock and reset for the System
+  // The implicitClockSinkNode provides the implicit clock and reset for the system (connected by clocking scheme)
   val implicitClockSinkNode = ClockSinkNode(Seq(ClockSinkParameters(name = Some("implicit_clock"))))
 
   // Generate Clocks and Reset
-  p(ClockingSchemeKey)(this)
+  val mvRefClkFreq = p(ClockingSchemeKey)(this)
```
colinschmidt, line 38: Why do we even need this val? I think it should just be inlined below.

generators/chipyard/src/main/scala/HarnessBinders.scala
```diff
@@ -139,6 +139,46 @@ class WithSimAXIMem extends OverrideHarnessBinder({
...
+            th.harnessReset)
+// DOC include end: HarnessClockInstantiatorEx
+          val success = SerialAdapter.connectSimSerial(harnessMultiClockAXIRAM.module.io.tsi_ser, th.harnessClock, th.harnessReset.asBool)
+          when (success) { th.success := true.B }
+
+          // connect SimDRAM from the AXI port coming from the harness multi clock axi ram
+          (harnessMultiClockAXIRAM.mem_axi4 zip harnessMultiClockAXIRAM.memNode.edges.in).map { case (axi_port, edge) =>
+            val memSize = sVal.memParams.size
+            val lineSize = p(CacheBlockBytes)
+            val mem = Module(new SimDRAM(memSize, lineSize, BigInt(memFreq.toInt), edge.bundle)).suggestName("simdram")
```
colinschmidt, line 171: Does this work above MAX_INT hertz? For example at the default firesim 3.2GHz? It seems like to me that would fail and we should instead use `toLong`

sims/vcs/Makefile
```diff
@@ -66,7 +66,7 @@ $(sim_debug): $(sim_vsrcs) $(sim_common_files) $(dramsim_lib) $(EXTRA_SIM_REQS)
 #########################################################################################
 .PRECIOUS: $(output_dir)/%.vpd %.vpd
 $(output_dir)/%.vpd: $(output_dir)/% $(sim_debug)
-	(set -o pipefail && $(sim_debug) $(PERMISSIVE_ON) $(SIM_FLAGS) $(EXTRA_SIM_FLAGS) $(VERBOSE_FLAGS) +vcdplusfile=$@ $(PERMISSIVE_OFF) $< </dev/null 2> >(spike-dasm > $<.out) | tee $<.log)
+	(set -o pipefail && $(sim_debug) $(PERMISSIVE_ON) $(SIM_FLAGS) $(EXTRA_SIM_FLAGS) $(SEED_FLAG) $(VERBOSE_FLAGS) +vcdplusfile=$@ $(PERMISSIVE_OFF) $< </dev/null 2> >(spike-dasm > $<.out) | tee $<.log)
```
colinschmidt, line 69: I don't think I saw this getting a default in a makefile fragement.

colinschmidt (commented):
Looks pretty good, a couple of comments/potential issues.
Also did we already have some discussion about what to do about images and the source that generated them? If so what was the decision?

docs/Advanced-Concepts/Chip-Communication.rst
abejgonzalez: I think we can glide over the width here since that is really an implementation detail that can be changed (if the user understands what TSI is). In any case we briefly discuss AXI vs TSI widths in the testchip section.

docs/Advanced-Concepts/Chip-Communication.rst
abejgonzalez: I think this is fine since this entire section is referring to "things that inspect the target memory for testing purposes". Its not talking about different ports off the DUT, etc. (If I'm understanding your question correctly).

sims/vcs/Makefile
abejgonzalez: This is set by `vcs.mk`

generators/chipyard/src/main/scala/ChipTop.scala
abejgonzalez: Leaving this out breaks things. I think since the `def` is like a `lazy val` the clocking scheme function will get called in the  harness (non-lazy domain) causing a bunch of diplomacy errors in the `ChipTop`. So you 1st need to have the function called, then you can reference it with a `def`. Scala'y people can probably tell you the exact issue here.

docs/Advanced-Concepts/Harness-Clocks.rst
abejgonzalez: Yea the naming is off here. `harnessClock` really means `buildtopRefClock`.

abejgonzalez:
> Also did we already have some discussion about what to do about images and the source that generated them? If so what was the decision?

I think we had one in passing. And we didn't make a definitive decision. For now, I think the sources should not be put here since Chipyard is already big (we could just give a Drive link with the relevant sources and update them there).

docs/Advanced-Concepts/Harness-Clocks.rst
abejgonzalez: Changed to `buildtopClock`

generators/chipyard/src/main/scala/HarnessBinders.scala
abejgonzalez: Changed

sims/vcs/Makefile
colinschmidt: Ok so this is just a bug fix that it was not included before.

colinschmidt (approved):
LGTM.
I think the image source size issue doesn't make sense to me as I bet the source for those pngs is smaller than the output, but I'm not willing to die on this bike shed.

PR #826, opened 2021-03-12
Fix TileResetCtrl to be ahead of reset synchronizers.

jerryz123:
This PR moves the TileResetCtrl registers to be ahead of the reset synchronizers. The TileResetCtrl registers emit an async reset that a downstream synchronizer should catch and sync.
Additionally, this gives the TileResetCtrl registers the async reset coming directly from ChipTop, through a new AsyncResetBroadcast mini network.
While in theory the ChipTop async reset can be broadcast through the same diplomacy graph as the clocks, I think that approach is a bit confusing. When we expect the async reset to be coming directly from ChipTop, the code should make it clear that that is what is happening.

**Type of change**: bug fix 

**Impact**: rtl change

### DIFF
--- a/generators/chipyard/src/main/scala/Clocks.scala
+++ b/generators/chipyard/src/main/scala/Clocks.scala
@@ -70,16 +70,16 @@ object ClockingSchemeGenerators {
 
     // Add a control register for each tile's reset
     val resetSetter = chiptop.lazySystem match {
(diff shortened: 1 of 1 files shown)


davidbiancolin (approved):
LGTM.

colinschmidt (approved):
LGTM

PR #828, opened 2021-03-15
Add option to add async queues between chip-serialIO and harness serdes

jerryz123:

**Impact**: rtl change


abejgonzalez (approved):
LGTM. We discussed this offline.

PR #832, opened 2021-03-20
Don't pass JTAG oe signal off-chip

jerryz123:
The RocketChip JTAGIO exposes the TDO.enable signal, which shouldn't go off chip.
This adds a JTAGChipIO which we use as the type of the JTAG IOs for ChipTop.

**Type of change**: bug fix

**Impact**: rtl change


colinschmidt (approved):
LGTM. Was this a change in the rocket-chip interface or just a bug in our initial implementation?

jerryz123:
Bug in the initial implementation.

PR #835, opened 2021-03-25
Add support for Xilinx Virtex-7 FPGA VC709 board

mbs0221:
Tested with [sifive/fpga-shells#160](https://github.com/sifive/fpga-shells/pull/160)

**Related issue**: 

**Impact**: new fpga target

**Release Notes**

Add support for Xilinx Virtex-7 FPGA VC709 board.

Currently, the Linux kernel (version 5.11.0-rc3) can boot with the help of Open-SBI (version v0.9), and login with UART/Serial port on the Xilinx Virtex-7 FPGA VC709 board!


fpga/Makefile
```diff
@@ -120,9 +135,19 @@ debug-bitstream: $(build_dir)/obj/post_synth.dcp
...
+
+.PHONY: mcs
+mcs: $(MCS_FILE)
+
 #########################################################################################
 # general cleanup rules
 #########################################################################################
 .PHONY: clean
 clean:
-	rm -rf $(gen_dir)
```
abejgonzalez, line 128: In Chipyard we have this `clean` target "clean" everything (hence the `gen_dir` being deleted). I wouldn't be opposed to adding a new target like `clean-specific` (or a better name) that just deletes the sources related to this config.

fpga/Makefile
```diff
@@ -71,6 +85,7 @@ default: $(mcs)
 # misc. directories
 #########################################################################################
 fpga_dir := $(base_dir)/fpga/fpga-shells/$(FPGA_BRAND)
+fpga_board_script_dir := $(fpga_dir)/$(BOARD)/tcl
```
abejgonzalez, line 88: Where is this used?

fpga/Makefile
```diff
@@ -120,9 +135,19 @@ debug-bitstream: $(build_dir)/obj/post_synth.dcp
 			$(build_dir)/debug_obj \
 			$(fpga_common_script_dir)
 
+MCS_FILE := $(build_dir)/obj/$(MODEL).mcs
```
abejgonzalez, line 138: What is the MCS file used for? I removed this when originally setting up this work to keep things clean.

fpga/src/main/resources/vc709/uartboot/gpoi.c
```diff
@@ -0,0 +1,9 @@
+/*
```
abejgonzalez, line 1: Mis-spelled filename `gpoi` -> `gpio`.

fpga/README.md
```diff
@@ -0,0 +1,50 @@
+# Guide
```
abejgonzalez, line 1: 2 points here. 1st. We are currently trying to get FireMarshal to target the prototype platforms so a lot involving building Linux won't be necessary (stay tuned for that). 2nd. At a minimum, this should be put in markdown and put in the Chipyard docs in the prototyping section. Once that is done we can talk more about how to make this a bit more understandable for folks.

abejgonzalez:
First of all, thanks for the contribution! I gave a quick 1st pass on the PR. 

May you change the target to the `dev` branch. We use that branch for development work.

As mentioned in this post: https://github.com/ucb-bar/chipyard/pull/784#issuecomment-775686421... we are still trying to figure out some testing infrastructure since it's hard to test this stuff when we don't have the boards to run regressions. I'll let you know what the status is on that since that is a bit slow right now.

fpga/Makefile
mbs0221: Since every time we power off the board, the uploaded bit-stream will be lost. When working around the kernel after building up the prototype, we would not want to upload the bit-stream too often. Having a mcs file is better choice. It will save our time in the post stage of testing the prototype!

fpga/Makefile
mbs0221: I agree with your suggestion that It would be better to have a new target like `clean-specific` for a certain config!

fpga/Makefile
mbs0221: It is originally used in [sifive/freedom](https://github.com/sifive/freedom/blob/master/common.mk). I do not remember the original intention. It can be removed since we no longer need it currently.

fpga/src/main/resources/vc709/uartboot/gpoi.c
mbs0221: It is a mistake to be corrected.

PR #836, opened 2021-03-26
Sane FireSim Default Target Freqs

alonamid:
**Related issue**: N/A

**Impact**: rtl change

**Release Notes**
Update FireSim default frequencies for better tile/system/memory ratios.

This PR changes the default FireSim target tile frequency from 3.2 GHz to 2 GHz (with a memory bus of 1 GHz), because 3.2 GHz generates an unrealistic Tile/Memory cycles ratio.
The default configs are also updated with a slower system bus (1 GHz).

The PR also adds two additional packaged config tweaks which are representative of High-Perf (Legacy) target configs, and test-chip target configs (1 GHz tile instead of 2 GHz). 

- [ ] Upon approval, need to generate new pre-built AGFIs.

### DIFF
--- a/generators/firechip/src/main/scala/TargetConfigs.scala
+++ b/generators/firechip/src/main/scala/TargetConfigs.scala
@@ -12,7 +12,7 @@ import freechips.rocketchip.rocket.DCacheParams
 import freechips.rocketchip.subsystem._
 import freechips.rocketchip.devices.tilelink.{BootROMLocated, BootROMParams}
 import freechips.rocketchip.devices.debug.{DebugModuleParams, DebugModuleKey}
-import freechips.rocketchip.diplomacy.LazyModule
+import freechips.rocketchip.diplomacy.{LazyModule, AsynchronousCrossing}
 import testchipip.{BlockDeviceKey, BlockDeviceConfig, TracePortKey, TracePortParams}
 import sifive.blocks.devices.uart.{PeripheryUARTKey, UARTParams}
 import scala.math.{min, max}
@@ -85,8 +85,46 @@ class WithFireSimDesignTweaks extends Config(
 
 // Tweaks to modify target clock frequencies / crossings to firesim defaults
 class WithFireSimDefaultFrequencyTweaks extends Config(
-  // Optional*: Removing this will require adjusting the UART baud rate and
-  // potential target-software changes to properly capture UART output
+  // Optional: This sets the default frequency for all buses in the system to 2 GHz
+  // (since unspecified bus frequencies will use the pbus frequency)
+  new chipyard.config.WithPeripheryBusFrequency(2000.0) ++
+  // Optional: These three configs will put the system bus at a frequency of 1 GHz
+  // Which is more representative of on uncore working at a lower frequency than the tiles
+  new chipyard.config.WithSystemBusFrequency(1000.0) ++
+  new chipyard.config.WithSbusToCbusCrossingType(AsynchronousCrossing()) ++ // Add Async crossing between SBUS and CBUS
+  new chipyard.config.WithFbusToSbusCrossingType(AsynchronousCrossing()) ++ // Add Async crossing between FBUS and SBUS
+  // Optional: These three configs put the DRAM memory system in it's own clock domian.
+  // Removing the first config will result in the FASED timing model running
+  // at the pbus freq (above, 3.2 GHz), which is outside the range of valid DDR3 speedgrades.
+  // 1 GHz matches the FASED default, using some other frequency will require
+  // runnings the FASED runtime configuration generator to generate faithful DDR3 timing values.
+  new chipyard.config.WithMemoryBusFrequency(1000.0) ++
+  new chipyard.config.WithAsynchrousMemoryBusCrossing ++
+  new testchipip.WithAsynchronousSerialSlaveCrossing
+)
+
+// Tweaks to modify target clock frequencies / crossings to testchip defaults
+class WithFireSimTestchipFrequencyTweaks extends Config(
+  // Optional: This sets the default frequency for all buses in the system to 1 GHz
+  // (since unspecified bus frequencies will use the pbus frequency).
+  // This frequency is representative of Rocket/BOOM-based test chips
+  new chipyard.config.WithPeripheryBusFrequency(1000.0) ++
+  // Optional: These three configs put the DRAM memory system in it's own clock domian.
+  // Removing the first config will result in the FASED timing model running
+  // at the pbus freq (above, 3.2 GHz), which is outside the range of valid DDR3 speedgrades.
+  // 1 GHz matches the FASED default, using some other frequency will require
+  // runnings the FASED runtime configuration generator to generate faithful DDR3 timing values.
+  new chipyard.config.WithMemoryBusFrequency(1000.0) ++
+  new chipyard.config.WithAsynchrousMemoryBusCrossing ++
+  new testchipip.WithAsynchronousSerialSlaveCrossing
+)
+
+// Tweaks to modify target clock frequencies / crossings to legacy firesim defaults
+class WithFireSimHighPerfFrequencyTweaks extends Config(
+  // Optional: This sets the default frequency for all buses in the system to 3.2 GHz
+  // (since unspecified bus frequencies will use the pbus frequency)
+  // This frequency selection matches FireSim's legacy selection and is required
+  // to support 200Gb NIC performance. You may select a smaller value.
   new chipyard.config.WithPeripheryBusFrequency(3200.0) ++
   // Optional: These three configs put the DRAM memory system in it's own clock domian.
   // Removing the first config will result in the FASED timing model running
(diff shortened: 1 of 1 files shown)


generators/firechip/src/main/scala/TargetConfigs.scala
```diff
@@ -85,8 +85,46 @@ class WithFireSimDesignTweaks extends Config(
 
 // Tweaks to modify target clock frequencies / crossings to firesim defaults
 class WithFireSimDefaultFrequencyTweaks extends Config(
-  // Optional*: Removing this will require adjusting the UART baud rate and
-  // potential target-software changes to properly capture UART output
+  // Optional: This sets the default frequency for all buses in the system to 2 GHz
+  // (since unspecified bus frequencies will use the pbus frequency)
+  new chipyard.config.WithPeripheryBusFrequency(2000.0) ++
```
jerryz123, line 90: Why do we use pbus frequency as default again? @davidbiancolin

Seems more intuitive to use sbus frequency as default, set that and pbus frequency to 1000, and set tile frequency to 2000.

colinschmidt:
I think the new configs should be:
FireSimReasonablePerf: 2GHz(tile):1GHz(system/memory):500MHz(pbus)
TestChip:1GHz(tile):500MHz(sys/mem):250MHz(pbus)

generators/firechip/src/main/scala/TargetConfigs.scala
davidbiancolin: PBUS was chosen because at the time it was the only frequency being set.

davidbiancolin:
While I'd love to see the pbus running a slower frequency than the sbus, unless there have been some recent changes in RC, that design point probably won't function correctly without further modification. IIRC RTC handling expects that SBUS and PBUS are running at the same freq.

generators/firechip/src/main/scala/TargetConfigs.scala
```diff
@@ -85,8 +85,46 @@ class WithFireSimDesignTweaks extends Config(
...
+  // at the pbus freq (above, 3.2 GHz), which is outside the range of valid DDR3 speedgrades.
+  // 1 GHz matches the FASED default, using some other frequency will require
+  // runnings the FASED runtime configuration generator to generate faithful DDR3 timing values.
+  new chipyard.config.WithMemoryBusFrequency(1000.0) ++
+  new chipyard.config.WithAsynchrousMemoryBusCrossing ++
+  new testchipip.WithAsynchronousSerialSlaveCrossing
+)
+
+// Tweaks to modify target clock frequencies / crossings to legacy firesim defaults
+class WithFireSimHighPerfFrequencyTweaks extends Config(
```
davidbiancolin, line 123: Can we just call these "Frequencies" or "Clocking" and drop the tweaks

generators/firechip/src/main/scala/TargetConfigs.scala
```diff
@@ -85,8 +85,46 @@ class WithFireSimDesignTweaks extends Config(
 
 // Tweaks to modify target clock frequencies / crossings to firesim defaults
 class WithFireSimDefaultFrequencyTweaks extends Config(
```
davidbiancolin, line 87: I feel like these frequency selection configs aren't FireSim specific and could probably go into the chipyard package.

abejgonzalez: I think the testchip one is a reasonable one to go in Chipyard. The FireSim based one (3.2GHz) makes sense to stay here since it determines the 200GB/s NIC. The default one can probably also go in Chipyard... but since it is a default it should stay here no?

generators/firechip/src/main/scala/TargetConfigs.scala
jerryz123: Can we make the sbus the default then now, and always set the sbus frequency?
Its pretty unintuitive that PBus frequency ends up setting tile frequency, since the tiles aren't even directly adjacent to the pbus.

generators/firechip/src/main/scala/TargetConfigs.scala
```diff
@@ -205,7 +255,7 @@ class FireSimCVA6Config extends Config(
 //* Multiclock Configurations
 //*********************************************************************************/
 class FireSimMulticlockRocketConfig extends Config(
-  new chipyard.config.WithTileFrequency(6400.0) ++ //lol
+  new chipyard.config.WithTileFrequency(6000.0) ++ //lol
```
jerryz123, line 258: If the default is FireSimDefaultFrequency (2 GHz tile, 1 GHz everything else) , then do we still need this config around?

davidbiancolin: @alonamid Yeah, you can delete these, and delete the one "multiclock" scalatest, since they are all multiclock now. You'll need to remove it from CI too.

generators/chipyard/src/main/scala/ConfigFragments.scala
```diff
@@ -319,3 +319,17 @@ class WithControlBusFrequency(freqMHz: Double) extends Config((site, here, up) =
 
 class WithRationalMemoryBusCrossing extends WithSbusToMbusCrossingType(RationalCrossing(Symmetric))
 class WithAsynchrousMemoryBusCrossing extends WithSbusToMbusCrossingType(AsynchronousCrossing())
+
+class WithTestChipBusFreqs extends Config(
+  // Frequency specifications
+  new chipyard.config.WithTileFrequency(1600.0) ++       // Matches the maximum frequency of U540
+  new chipyard.config.WithSystemBusFrequency(800.0) ++   // Put the system bus at a lower freq, representative of ncore working at a lower frequency than the tiles. Same freq as U540
+  new chipyard.config.WithMemoryBusFrequency(1000.0) ++  // 2x the U540 freq (appropriate for a 128b Mbus)
+  new chipyard.config.WithPeripheryBusFrequency(100) ++  // Match the sbus and pbus frequency
```
davidbiancolin, line 328: Again, i'm worried about this differing from the SBUS freq. I think the RTC might report incorrect times, and linux scheduling will behave weird. I think we should tie them together for now.

generators/chipyard/src/main/scala/ConfigFragments.scala
```diff
@@ -319,3 +319,17 @@ class WithControlBusFrequency(freqMHz: Double) extends Config((site, here, up) =
...
+  new chipyard.config.WithSystemBusFrequencyAsDefault ++ // All unspecified clock frequencies, notably the implicit clock, will use the sbus freq (800 MHz)
+  //  Crossing specifications
+  new chipyard.config.WithCbusToPbusCrossingType(AsynchronousCrossing()) ++ // Add Async crossing between PBUS and CBUS
+  new chipyard.config.WithSbusToMbusCrossingType(AsynchronousCrossing()) ++ // Add Async crossings between backside of L2 and MBUS
+  new freechips.rocketchip.subsystem.WithRationalRocketTiles ++   // Add rational crossings between RocketTile and uncore
```
davidbiancolin, line 333: Can you add the BOOM equivalent mixin here, so that this config works with both rocket and boom cores?

Unless i'm mistaken and this works with both.

generators/firechip/src/main/scala/TargetConfigs.scala
```diff
@@ -98,9 +101,26 @@ class WithFireSimDefaultFrequencyTweaks extends Config(
   new testchipip.WithAsynchronousSerialSlaveCrossing
 )
 
-// Tweaks that are generally applied to all firesim configs
+// Tweaks that are generally applied to all firesim configs setting a single clock domain at 1000 MHz
 class WithFireSimConfigTweaks extends Config(
-  new WithFireSimDefaultFrequencyTweaks ++
+  // 1 GHz matches the FASED default, using some other frequency will require
+  // runnings the FASED runtime configuration generator to generate faithful DDR3 timing values.
+  new chipyard.config.WithPeripheryBusFrequency(1000.0) ++
```
davidbiancolin, line 108: Perhaps adding a comment saying that these buses (PBUS + MBUS) will be driven to 100 MHz (because of assignments in the Chisel, i'm guessing) unless overridden here, would be valuable.

generators/chipyard/src/main/scala/ConfigFragments.scala
```diff
@@ -319,3 +319,18 @@ class WithControlBusFrequency(freqMHz: Double) extends Config((site, here, up) =
...
+  // Frequency specifications
+  new chipyard.config.WithTileFrequency(1600.0) ++       // Matches the maximum frequency of U540
+  new chipyard.config.WithSystemBusFrequency(800.0) ++   // Put the system bus at a lower freq, representative of ncore working at a lower frequency than the tiles. Same freq as U540
+  new chipyard.config.WithMemoryBusFrequency(1000.0) ++  // 2x the U540 freq (appropriate for a 128b Mbus)
+  new chipyard.config.WithPeripheryBusFrequency(800.0) ++  // Match the sbus and pbus frequency
+  new chipyard.config.WithSystemBusFrequencyAsDefault ++ // All unspecified clock frequencies, notably the implicit clock, will use the sbus freq (800 MHz)
+  //  Crossing specifications
+  new chipyard.config.WithCbusToPbusCrossingType(AsynchronousCrossing()) ++ // Add Async crossing between PBUS and CBUS
+  new chipyard.config.WithSbusToMbusCrossingType(AsynchronousCrossing()) ++ // Add Async crossings between backside of L2 and MBUS
+  new freechips.rocketchip.subsystem.WithAsynchronousRocketTiles(4,4) ++   // Add Async crossings between RocketTile and uncore
```
davidbiancolin, line 333: Am i mistaken, but weren't these rational earlier in the PR? They definitely should be.

davidbiancolin (approved):
LGTM

PR #855, opened 2021-04-14
Runs `make clean` after `make-install` to reduce used disk

chick:
Make install leaves many .o and other files laying around
Fixes problems with disk space on Github Actions runner

**Impact**: other
Reduces used disk space after toolchain build

**Release Notes**
Intermediate .o and other files are removed after building toolchain components.
Reduces disk requirements

### DIFF
--- a/scripts/build-toolchains.sh
+++ b/scripts/build-toolchains.sh
@@ -39,6 +39,7 @@ EC2FASTINSTALL="false"
 IGNOREQEMU=""
 RISCV=""
 ARCH=""
+CLEANAFTERINSTALL=""
 
 # getopts does not support long options, and is inflexible
 while [ "$1" != "" ];
@@ -51,16 +52,19 @@ do
             RISCV=$(realpath $1) ;;
         --ignore-qemu )
             IGNOREQEMU="true" ;;
-	-a | --arch )
-	    shift
-	    ARCH=$1 ;;
-        riscv-tools | esp-tools)
-            TOOLCHAIN=$1 ;;
-        ec2fast )
-            EC2FASTINSTALL="true" ;;
-        * )
-            error "invalid option $1"
-            usage 1 ;;
+        --clean-after-install )
+            shift
+            CLEANAFTERINSTALL="--clean-after-install" ;;
+        -a | --arch )
+            shift
+            ARCH=$1 ;;
+              riscv-tools | esp-tools)
+                  TOOLCHAIN=$1 ;;
+              ec2fast )
+                  EC2FASTINSTALL="true" ;;
+              * )
+                  error "invalid option $1"
+                  usage 1 ;;
     esac
     shift
 done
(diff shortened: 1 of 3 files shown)


abejgonzalez (approved):
This LGTM, though we should quickly poll the Chipyard group if this is fine with them. Hopefully, there isn't a corner use case that wants the intermediate files.

jerryz123:
Lets make this a flag that CI sets. 
It is useful to not run make clean if you are developing/debugging toolchain issues.

abejgonzalez:
> Lets make this a flag that CI sets.
> It is useful to not run make clean if you are developing/debugging toolchain issues.

If the toolchain fails, then the intermediate files will not be deleted. The `set -eo pipefail` will catch it.

a0u:
I'd prefer having a flag to control this behavior.  Cleaning after each step makes it harder to debug cases where an earlier step might succeed but was built with incorrect options, so the failure only manifests after attempting a later step that depends on the first.  Retaining the intermediate files is also useful for incremental builds.

abejgonzalez:
Sounds good, flag it is. This PR should probably also bump the CircleCI `config.yml` toolchain version so that the toolchain gets rebuilt.

chick:
@abejgonzalez I've implemented a flag to keep the existing behavior the default. I'm not particularly enthused by the implementation. Feedback appreciated

scripts/build-toolchains.sh
```diff
@@ -51,16 +52,19 @@ do
...
-        * )
-            error "invalid option $1"
-            usage 1 ;;
+        --clean-after-install )
+            shift
+            CLEANAFTERINSTALL="--clean-after-install" ;;
+        -a | --arch )
+            shift
+            ARCH=$1 ;;
+              riscv-tools | esp-tools)
```
abejgonzalez, line 61: This is mis-aligned, the `riscv-tools` and `ec2fast` sections should be indented the same as the `-a`.

abejgonzalez:
I'm fine with this implementation (though I'm not picky at all about it). Can you bump the `config.yml` file as mentioned here:

> Sounds good, flag it is. This PR should probably also bump the CircleCI config.yml toolchain version so that the toolchain gets rebuilt (or have the hash depend on the build-toolchains and build-util.sh scripts).

scripts/build-util.sh
```diff
@@ -81,6 +86,9 @@ module_build() ( # <submodule> [configure-arg..]
         "${MAKE}"
         echo "==>  Installing ${name}"
         "${MAKE}" install
+        if [ "$clean_after_install" = "true" ] ; then
```
a0u, line 89: Based on the original patch, I assumed the intent was to clean up after all the submodules rather than just riscv-gnu-toolchain:

```suggestion
        if [ -n "$CLEANAFTERINSTALL" ] ; then
```

which makes parsing `--clean-after-install` unnecessary.

If a particular submodule needs to be excluded, the variable could be overridden at the call site:
```sh
CLEANAFTERINSTALL= module_build ...
```

chick:
@a0u I changed this as you suggested, and I think it serves my purposes and preserves default behavior.
I don't want to clutter things but do you think I should add to the `build-toolchains.sh` usage to note its
sensitivity to the `CLEANAFTERINSTALL` env var?

a0u:
@chick Sorry, I should have been more specific.  I meant that `build-toolchains.sh` would still parse `--clean-after-install` and set `CLEANAFTERINSTALL` accordingly.  `module_build` and `module_make` would just test whether this shell variable inherited from the outer subshell is non-empty, rather than parsing `--clean-after-install` a second time.  This makes the behavior automatically consistent for all submodules using `module_build`; otherwise `--clean-after-install` would need to be explicitly passed at each call.

chick:
@a0u No problem, I think it's a lot clearer with it explicitly in the flags for build-toolchains.

a0u (approved):

PR #867, opened 2021-04-26
Update Initial Setup Scripts

KarlJoad:
Multiple small updates to the initial Chipyard setup scripts.

**Related issue**: 

**Type of change**: Update documentation

**Impact**: software change

**Release Notes**

PR #870, opened 2021-04-28
Update initial setup scripts

abejgonzalez:
**Related issue**: #867 

**Impact**:other

**Release Notes**
Test if these changes are backwards compatible.

@KarlJoad

### DIFF
--- a/scripts/centos-req.sh
+++ b/scripts/centos-req.sh
@@ -4,7 +4,13 @@ set -ex
 
 sudo yum groupinstall -y "Development tools"
 sudo yum install -y gmp-devel mpfr-devel libmpc-devel zlib-devel vim git java java-devel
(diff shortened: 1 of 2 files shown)


alonamid (approved):
LGTM, but it might be worth it to kick CI to make sure nothing was accidently cached.

timsnyder-siv:
I think bintray is also mentioned in https://github.com/ucb-bar/chipyard/blob/KarlJoad-update-initial-setup-scripts/scripts/centos-req.sh and https://github.com/ucb-bar/chipyard/blob/KarlJoad-update-initial-setup-scripts/scripts/ubuntu-req.sh that get included into the docs.  Those scripts should also be updated so that the docs are consistent.

timsnyder-siv:
Please ignore my earlier comment.  I'm trying to wake up this morning...

abejgonzalez:
Opened single PR to test this #874

Issue #820, opened 2021-03-08
Prototyping Harness Sim Testbench


jamesdunn:
I haven't made as much progress on this as I'd like over the last couple weeks, so creating a feature request issue to keep it on my list and put down what I've thought about so far.

The feature request is to create simulation testbenches for FPGA harnesses that verify IO functionality. Someone submitting a PR for a new FPGA board could run these tests, or we could make it part of our CI.

It seems straightforward enough to use VCS to simulate Vivado-generated Verilog, including behavioral models provided for Xilinx IP blocks, since the .f files generated in the prototyping make flow include pointers to any Xilinx IP used.

I'll add a VCS sim with arguments and includes as a make target to `fpga/Makefile`, reusing a lot of what is already in `sims/vcs/Makefile`.

A simple testbench wrapper around the FPGA testharness that shorts certain Tx and Rx signals would allow a loopback test for many interfaces like UART, JTAG, and SPI. Then compiled loopback test programs like [this one for UART](https://github.com/sifive/example-uart-testbench) can be ran by the simulator binary, and I can write new test programs for JTAG and SPI.

Let me know if people have any feedback.

jamesdunn:
Update on progress: the branch `arty-sim` now has a config and Make SUB_PROJECT that builds with a SimSerial model, and the harness success IO not tied off.

Now I need to write an appropriate TestDriver.v to place the Harness in a configuration that tests the IOs of peripherals like UART and JTAG, and add a make target that calls VCS with proper arguments.

I'm also thinking about a concurrent XSIM testbench that would allow users without access to a VCS license to test their FPGA prototyping harnesses.

jamesdunn:
I'm now able to generate a VCS sim binary for an ArtyTestDriver that submodules the ArtyTestHarness. This includes all of the behavioral models for any Xilinx IP used. Here's what I did to get there:

The current way that Chipyard provides behavioral models to VCS is by passing Verilog and SystemVerilog in list files with the `-f` flag, and the C collateral with includes in the `CFLAGS`. This works when all of the input behavioral HDL is in a single language like Verilog. All of the Xilinx IP primitives and most of the macro blocks have behavioral Verilog, and can be passed in as described here. However, there are some macro blocks where only behavioral VHDL is provided, and there is one such piece of IP used in Arty (reset_sys).

[...]

PR #871, opened 2021-04-29
FPGA prototyping platform VCS simulation

jamesdunn:
**Related issue**:  #820

**Impact**: software change

**Release Notes**

This PR contains a draft of FPGA prototyping sim support. It currently targets Arty, but I have tried to generalize the flow such that it should be able to target the VCU118 with minor changes.

The sim binary builds, but exhibits the same behavior described in #820, so please do not merge. However, I wanted to get in a PR so that the behavior can be reproduced with a make target. In that issue, the sim binary is missing the program-as-argument functionality given in testchip_tsi.cc. David suggested looking into `+permissive_on/off` flags, which I will do next.

To summarize what this PR contains, I have added a `sim-fpga` target in `fpga/Makefile` that runs a .tcl script, `/fpga/scripts/generate_vcs_collateral.tcl`, to generate Vivado behavioral sim collateral for VCS. It then runs through the VCS mixed-language flow, analyzing the Verilog and VHDL, then elaborating the simulation binary.

I will definitely do a lot of cleanup before merging this. Partcularly de-duplication between this Makefile and some things in `vcs.mk` that were hard-coded. Also figuring out a cleaner way to bring in an FPGA-specific testdriver, which for now uses an additional case in `generators/utilities/src/main/scala/Simulator.scala` for a new `sim_name`.

To run:
```
./scripts/init-fpga.sh
cd fpga
make SUB_PROJECT=artysim sim-fpga
```


fpga/Makefile
```diff
@@ -120,6 +137,84 @@ debug-bitstream: $(build_dir)/obj/post_synth.dcp
...
 			$(fpga_common_script_dir)
 
+#########################################################################################
+# sim rules
+#########################################################################################
+
+# hack until able to restructure fpga-shells .tcl scripts
+project-creation-script := $(fpga_common_script_dir)/create_project.tcl
+$(project-creation-script): $(fpga_common_script_dir)/vivado.tcl
+	cd $(fpga_common_script_dir); sed -n '1,10p' vivado.tcl > create_project.tcl
```
abejgonzalez, line 147: I'm inclined to say that this should just reside in the `generated-src/$LONG_NAME` folder when created (and thus gets deleted when that is deleted). Then you don't need to re-generate this everytime a new sim is done.

fpga/Makefile
```diff
@@ -120,6 +137,84 @@ debug-bitstream: $(build_dir)/obj/post_synth.dcp
...
+	rm $(project-creation-script)
+
+# generate vcs collateral
+vivado-vcs-collateral := $(build_dir)/$(MODEL).cache/compile_simlib/synopsys_sim.setup
+$(vivado-vcs-collateral): $(sim_dir)/scripts/generate_vcs_collateral.tcl $(vivado-project)
+	cd $(build_dir); vivado $(vivado-project) \
+		-nojournal -mode batch \
+		-source $(sim_dir)/scripts/generate_vcs_collateral.tcl \
+		-tclargs \
+			$(FPGA_FAMILY) \
```
abejgonzalez, line 170: I'm surprised that you don't need to pass the board itself (like Arty A7)...

fpga/src/main/scala/arty/Configs.scala
```diff
@@ -24,7 +22,6 @@ class WithDefaultPeripherals extends Config((site, here, up) => {
     idcodePartNum = 0x000,
     idcodeManufId = 0x489,
     debugIdleCycles = 5)
-  case SerialTLKey => None // remove serialized tl port
```
abejgonzalez, line 27: I think we should leave this as the default, then we re-add this in the "Sim" config. By default, most people will probably just go directly to building the bitstream and ignore simulation.

fpga/src/main/scala/arty/TestHarness.scala
```diff
@@ -29,7 +29,7 @@ class ArtyFPGATestHarness(override implicit val p: Parameters) extends ArtyShell
 
   val harnessClock = clock_32MHz
   val harnessReset = hReset
-  val success = false.B
+  val success = IO(Output(Bool()))
```
abejgonzalez, line 32: Have you tried to rebuild the bitstream with this? Also, I would match what the Chipyard testharness does if this works in the bitstream (i.e. create the IO and assign it to `false.B` instead of floating).

generators/utilities/src/main/resources/vsrc/ArtyTestDriver.v
```diff
@@ -0,0 +1,173 @@
+// See LICENSE.SiFive for license details.
+//VCS coverage exclude_file
+`ifndef RESET_DELAY
```
abejgonzalez, line 3: Can you remind me why we can't just use the RC TestDriver again?

abejgonzalez (commented):
Overall, a good start. It might be worth it to add a Makefile check for Vivado versions (what Vivado version are you using?).

fpga/Makefile
```diff
@@ -9,7 +9,7 @@ base_dir=$(abspath ..)
 sim_dir=$(abspath .)
 
 # do not generate simulation files
-sim_name := none
+sim_name := arty
```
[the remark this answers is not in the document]
jamesdunn: Thanks and agreed—fixed in most recent commit.

fpga/src/main/scala/arty/Configs.scala
jamesdunn: Thanks and agreed—fixed in most recent commit.

jamesdunn:
> Overall, a good start. It might be worth it to add a Makefile check for Vivado versions (what Vivado version are you using?).

Thanks for your review and comments, Abe. I have added a Makefile variable `VIVADO_VERSION` that gets the version. I am currently using v2019.2. I anticipate other versions being compatible, but I agree that it's good to support only one version and do a check. Do you recommend erroring out of the build on a version mismatch? If so, what make target should that be under? The bitstream generation and simulation targets are independent of one another, so I could put it under both.

fpga/Makefile
jamesdunn: I agree. I ended up structuring this a bit differently in the most recent commit, creating a permanent `fpga/scripts/create_project.tcl` script, so no need for a sed-hacked temporary script living in fpga-shells.

fpga/Makefile
jamesdunn: The Vivado tcl command in `generate_vcs_collateral.tcl` that takes the `FPGA_FAMILY` is `compile_simlib`, which compiles behavioral models for primitives (such as IOBUFs) which fpga-shells blackboxes. These are specific to an FPGA family (e.g. Ultrascale, Kintex), and not peripherals of the dev board on which the FPGA lives.

fpga/src/main/scala/arty/Configs.scala
```diff
@@ -38,4 +38,10 @@ class WithArtyTweaks extends Config(
 class TinyRocketArtyConfig extends Config(
   new WithArtyTweaks ++
   new chipyard.TinyRocketConfig)
+
+class TinyRocketArtySimConfig extends Config( // any additional IO needed for VCS sim go here.
+  new TinyRocketArtyConfig ++
+  new chipyard.harness.WithTiedOffDebug ++
```
jerryz123, line 44: I believe the ordering of these fragments should be reversed (TinyRocketArtyConfig last).

fpga/src/main/scala/arty/TestHarness.scala
jamesdunn: Good point. When building the bitstream, I get this error:
```ERROR: [DRC UCIO-1] Unconstrained Logical Port: 1 out of 10 logical ports have no user assigned specific location constraint (LOC). This may cause I/O contention or incompatibility with the board power or connectivity affecting performance, signal integrity or in extreme cases cause damage to the device or the components to which it is connected. To correct this violation, specify all pin locations. This design will fail to generate a bitstream unless all logical ports have a user specified site LOC constraint defined.
```
Which I am sure pertains to the success signal. I will work on this.

fpga/src/main/scala/arty/Configs.scala
jamesdunn: Thanks, fixed this in commit.

generators/utilities/src/main/resources/vsrc/ArtyTestDriver.v
jamesdunn: There were a couple of reasons. The first is because the ArtyFPGATestHarness has different names for clock and reset, extended from fpga-shells' ArtyShell (e.g. `CLK100MHz` instead of `clock` and `ck_rst` instead of `reset`). I couldn't think of a way to fix this without wrapping the ArtyShell.

The second is because I wanted to start playing around with placing the ArtyFPGATestHarness in different configurations, like shorting UART Tx and Rx in the driver. Though as we mentioned in our meeting, the different harness configurations for exercising peripherals can come in a different PR.

fpga/Makefile
```diff
@@ -14,6 +14,9 @@ sim_name := none
 #########################################################################################
 # include shared variables
 #########################################################################################
+
+VIVADO_VERSION := $(shell vivado -version | head -1 | grep -o -P '(?<=\s).*(?=\s)')
```
abejgonzalez, line 18: If Vivado isn't on the path this pollutes the terminal output. Maybe pipe stderr to `/dev/null`. Also, where is this checked?

jamesdunn:
Figured out what is going on with the simulation binary not accepting programs as argument, described [here](https://github.com/ucb-bar/chipyard/issues/820#issuecomment-841693950).

Think this is related to the behavioral model for the Xilinx reset state machine IP, and will investigate with waveforms.

jamesdunn:
Will also rebase this PR onto #879, which will break a circular dependency on `utilities` and fix the hacky way I currently include the ArtyTestDriver.

jamesdunn:
Just wanted to update this PR. Thanks to Jerry for helping me resolve the reset issue.

I have an open support ticket with Synopsys to determine why the SimSerial and testchip_tsi SystemC functionality is not making its way into the VCS compiled simulation binary.

jamesdunn:
> Just wanted to update this PR. Thanks to Jerry for helping me resolve the reset issue.
> 
> I have an open support ticket with Synopsys to determine why the SimSerial and testchip_tsi SystemC functionality is not making its way into the VCS compiled simulation binary.

I experimented with this some more, and found that this only happens whenever VHDL is present in the design and the vhdlan analysis tool is used. If I remove the VHDL module (only to test the flow since the full design will not function without the VHDL behavioral model of the XIlinx reset state machine) and omit the vhdlan analysis in the mixed-signal flow, the simv binary accepts the program as an argument (the argv to testchip_tsi in testchip_tsi.cc). In both cases, the System C files are compiled (.o files exist in the output directory), but it appears those are not being linked into the simv binary whenever the vhdlan analysis tool is used in this flow. I met with a Synopsys rep and he does not understand why, so I am working with him to replicate the full environment so that Synopsys can test this internally.

jamesdunn:
I met with the Synopsys rep, who analyzed the design and determined that the message I was getting:
```
Warning-[RUNT-OPT-UNKNOWN] Unknown runtime option
  Unknown VCS runtime option 'rv64ui-p-simple' passed as part of command
  './simv-chipyard.fpga.arty-TinyRocketArtySimConfig rv64ui-p-simple '.
  Please review valid vcs runtime options by executing ./<simulation
  executable> -help
```
Is actually benign and the simv binary *does* accept the program as an argument. He filed a VCS enhancement to get rid of this message since it is confusing.

So, it does look like this is an issue with my design itself. Looking at the waveforms, I think the SimSerial is not loading in the program because it is getting a strange clock:
2021-07-27 15_23_52-NoMachine - bwrcrdsl
Top waveform is the Harness clock, below it is the SimSerial's clock.

They're synchronous in the simulation binary generated in the VCS sims flow:
2021-07-27 16_19_52-NoMachine - bwrcrdsl
 
I'll take a closer look at the RTL for the clocks. There is a Xilinx clock manager called the [mmcm](https://www.xilinx.com/products/intellectual-property/mmcm_module.html) in the Arty hardware design, and a behavioral model of that in this sim design, so I will see if the SimSerial is being hooked up to that and getting strange clocks from it.

PR #716, opened 2020-11-10
Enable FireSim distributed elaboration

timsnyder-siv:
**Related issue**: firesim/firesim#646

**Type of change**: other enhancement

**Impact**: rtl change | software change | unknown | other

**Release Notes**
* build.sbt assembly merge strategy made more conservative and custom strategy for preferring non-rocketchip artifacts
* `chipyard.config.WithBootROM` now looks for contentsFile in `TargetDirKey` instead of relative to current-working directory
   * removes `firesim.firesim.WithBootROM` because things are simplified

PR #872, opened 2021-04-30
Support firesim distributed elaboration

timsnyder-siv:
**Related issue**: firesim/firesim#646
This is #716 re-incarnated using a local branch so that the CI will run (and rebased several months forward).

**Impact**: rtl change & infrastructure change
Do you guys consider changes to the makefiles as 'software' change or is that target software?
RTL change removes the FireSim/firechip duplication of `WithBootROM` and instead 

**Release Notes**

* Updates sbt-assembly config so that assembly is able to merge everything into a fat jar once again.
* Adds a message to assertion in `chipyard.Generator` `--legacy-config` parser.  Error without message was confusing.
* Squelch message about not finding `fd`,  if it isn't found, makefiles fall back to `find`.
* Adjust makefiles to support elaborating with assembly JAR and FireSim's new `gen-replace-rtl-script` target

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -60,7 +60,7 @@ include $(base_dir)/tools/dromajo/dromajo.mk
 #########################################################################################
 # Returns a list of files in directory $1 with file extension $2.
 # If available, use 'fd' to find the list of files, which is faster than 'find'.
-ifeq ($(shell which fd),)
+ifeq ($(shell which fd 2>/dev/null),)
 	lookup_srcs = $(shell find -L $(1)/ -name target -prune -o -iname "*.$(2)" -print 2> /dev/null)
 else
 	lookup_srcs = $(shell fd -L ".*\.$(2)" $(1))
--- a/common.mk
+++ b/common.mk
@@ -6,7 +6,7 @@ SHELL=/bin/bash
 ifndef RISCV
 $(error RISCV is unset. You must set RISCV yourself, or through the Chipyard auto-generated env file)
 else
-$(info Running with RISCV=$(RISCV))
+$(info #Running with RISCV=$(RISCV))
 endif
 
 #########################################################################################
--- a/generators/firechip/src/main/scala/TargetConfigs.scala
+++ b/generators/firechip/src/main/scala/TargetConfigs.scala
@@ -4,6 +4,7 @@ import java.io.File
 
 import chisel3._
 import chisel3.util.{log2Up}
+import chipyard.config.WithBootROM
 import freechips.rocketchip.config.{Parameters, Config}
 import freechips.rocketchip.groundtest.TraceGenParams
 import freechips.rocketchip.tile._
@@ -23,19 +24,6 @@ import testchipip.WithRingSystemBus
 import firesim.bridges._
 import firesim.configs._
 
-class WithBootROM extends Config((site, here, up) => {
-  case BootROMLocated(x) => {
-    val chipyardBootROM = new File(s"./generators/testchipip/bootrom/bootrom.rv${site(XLen)}.img")
-    val firesimBootROM = new File(s"./target-rtl/chipyard/generators/testchipip/bootrom/bootrom.rv${site(XLen)}.img")
-
-    val bootROMPath = if (chipyardBootROM.exists()) {
-      chipyardBootROM.getAbsolutePath()
-    } else {
-      firesimBootROM.getAbsolutePath()
-    }
-    up(BootROMLocated(x), site).map(_.copy(contentFileName = bootROMPath))
-  }
-})
 
 // Disables clock-gating; doesn't play nice with our FAME-1 pass
 class WithoutClockGating extends Config((site, here, up) => {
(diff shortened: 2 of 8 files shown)


timsnyder-siv:
I understand why the BootROM change breaks things.  I'll go fix the makefiles.

timsnyder-siv:
Ok.  Now that I'm past the FireSim failures, I see the breakages I thought I might see in `sims/verilator/Makefile`.  I'm looking for where I would need to make a change similar to https://github.com/firesim/firesim/pull/651/files#diff-51c697eb76cf079de5f54272f1dbb3f8f9a087446f3280b154b526c85a5d66a8R42-R49 in the Chipyard makefiles (copying the bootrom.img files into the target-dir for the compilation).

abejgonzalez:
> Do you guys consider changes to the makefiles as 'software' change or is that target software?

I think this could be clearer... but I normally refer to target software.

> Should I bump FireSim to point to the firesim#651 PR?

Yup!

> Ok. Now that I'm past the FireSim failures, I see the breakages I thought I might see in `sims/verilator/Makefile`. I'm looking for where I would need to make a change similar to https://github.com/firesim/firesim/pull/651/files#diff-51c697eb76cf079de5f54272f1dbb3f8f9a087446f3280b154b526c85a5d66a8R42-R49 in the Chipyard makefiles (copying the bootrom.img files into the target-dir for the compilation).

If you want to copy the bootrom files from testchipip, you can do that either in `GenerateSimFiles` (in the `generators/utilities` area) (which is what I think you are doing right now), or you can do it in the `common.mk` section of the makefile (make it a pre-requisite of the main generator). I think changing `GenerateSimFiles` makes sense though.

common.mk
```diff
@@ -6,7 +6,7 @@ SHELL=/bin/bash
 ifndef RISCV
 $(error RISCV is unset. You must set RISCV yourself, or through the Chipyard auto-generated env file)
 else
-$(info Running with RISCV=$(RISCV))
+$(info #Running with RISCV=$(RISCV))
```
jerryz123, line 9: Is there a reason to remove this? We used to run into lots of issues when RISCV is set to an unintended path.

timsnyder-siv: I'm not removing it.  I'm only prepending the message with the shell comment character so that when I do my thing to convert makefile into a runscript to send to a remote host, the info string doesn't get interpreted as a command. See https://github.com/firesim/firesim/pull/651/files#diff-d3631732b22e280769ab899d454c5e77021c7835748f416f2f4b76a7cfa3eb0aR235-R250

timsnyder-siv:
w00t CI is passing everything except the FireSim is on master check!

common.mk
jerryz123: Oh I see. Misread this line.

variables.mk
```diff
@@ -170,6 +170,7 @@ SBT_NON_THIN ?= $(subst $(SBT_CLIENT_FLAG),,$(SBT))
 define run_scala_main
 	cd $(base_dir) && $(SBT) ";project $(1); runMain $(2) $(3)"
 endef
+run_main = cd $(base_dir) && java $(JAVA_ARGS) -cp $(FAT_JAR) $(2) $(3)
```
abejgonzalez, line 173: I assume you mean `$(JAVA_OPTS)` here? This API changed since you last worked on this (I also see this change needing to be done in the FireSim PR).

Additionally, I think we should match these functions (either both have "define" or both are one-liners).

Finally, this is really just used in FireSim correct (in `target-agnostic.mk`), I wonder if this makes more sense to live in FireSim instead of Chipyard's makefiles especially since Chipyard doesn't use this.

abejgonzalez (commented):
Small comments here and there. Overall looks good.

generators/firechip/src/main/scala/TargetConfigs.scala
```diff
@@ -4,6 +4,7 @@ import java.io.File
 
 import chisel3._
 import chisel3.util.{log2Up}
+import chipyard.config.WithBootROM
```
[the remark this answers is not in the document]
timsnyder-siv: Abe, are you saying I can remove it from the concrete FireSim configs toward the bottom of the file?  See https://github.com/ucb-bar/chipyard/pull/872/files/b21ae89b599a93adf222937cdea34a424b4bd770#diff-e4468b96461e4c7fcb103a17a6a01bc40296e9aeeb712d1ea4218d57b0dd3e80R126

If those usages further down in the file aren't needed anymore, I'm happy to remove them and the import.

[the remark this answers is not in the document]
abejgonzalez: Nope. I missed that. This is fine.

generators/utilities/src/main/scala/Simulator.scala
```diff
@@ -122,15 +122,15 @@ object GenerateSimFiles extends App with HasGenerateSimConfig {
       case None => Seq()
     })
 
-  def writeBootrom(): Unit = {
-    firrtl.FileUtils.makeDirectory("./bootrom/")
-    writeResource("/testchipip/bootrom/bootrom.rv64.img", "./bootrom/")
-    writeResource("/testchipip/bootrom/bootrom.rv32.img", "./bootrom/")
-    writeResource("/bootrom/bootrom.img", "./bootrom/")
+  def writeBootrom(cfg: GenerateSimConfig): Unit = {
```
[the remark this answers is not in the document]
timsnyder-siv: addressed in 
be498d4

variables.mk
timsnyder-siv: took this back out of variables.mk since FireSim only uses it in 
090eb1a 
I also addressed the `JAVA_ARGS` uses in firesim.  Thanks for catching this!

joey0320:
Closing stale PR

PR #876, opened 2021-05-04
avoid git.qemu.org

timsnyder-siv:

**Impact**:  other

**Release Notes**

git.qemu.org seems to be down or under heavy load. Latest qemu
seems to have changed it's .gitmodules to use gitlab.  Since we're
using a github.com mirror, when cloning qemu use rewrite rules to
use github instead of git.qemu.org.  Also install the rewrite rules
recursively into the local config of qemu and it's submodules so
that any further git commands done interactively by the user
will also use github.

Failures to connect to git.qemu.org have caused FireSim CI to fail all day today.


davidbiancolin (approved):
It's breaking CY CI as well. I

davidbiancolin:
Might be worth adding a note reminding us we can remove this once we bump far enough.

timsnyder-siv:
I thought about a note, or about checking `toolchains/qemu/.gitmodules` for the offending url and printing a message to remove it when it's no longer needed.  I can add that in a followup.  I'd like to just let the CI finish if it is going to and merge this.

abejgonzalez (approved):
LGTM

PR #877, opened 2021-05-04
test whether git.qemu.org rewrite still needed

timsnyder-siv:
**Related issue**: #876 

**Impact**: other

**Release Notes**

In https://github.com/ucb-bar/chipyard/pull/876#issuecomment-831639151
it was requested that we add a note to build-toolchains.sh saying that the
url rewrites can be removed.  I went a step further and made it self-checking.

Once qemu is bumped far enough that none of it's submodules come from
git.qemu.org, then the test added in this commit will fail CI during the bump PR
and everyone will be reminded to remove the url renaming rules and change
module_build back to module_all for qemu.


abejgonzalez (approved):
LGTM

PR #879, opened 2021-05-06
Remove GenerateSimFiles and use make instead

abejgonzalez:
**Related issue**: #872 

**Impact**: other

**Release Notes**
By using `make` we can track the files created by `GenerateSimFiles` and properly move them / update them on changes. This additionally comes with a bonus of removing 1 SBT call.

@timsnyder


sims/common-sim-flags.mk
```diff
@@ -21,3 +21,18 @@ SIM_LDFLAGS = \
...
+SIM_FILE_REQS += \
+	$(TESTCHIP_RSRCS_DIR)/testchipip/csrc/SimSerial.cc \
+	$(TESTCHIP_RSRCS_DIR)/testchipip/csrc/SimDRAM.cc \
+	$(TESTCHIP_RSRCS_DIR)/testchipip/csrc/testchip_tsi.cc \
+	$(TESTCHIP_RSRCS_DIR)/testchipip/csrc/testchip_tsi.h \
+	$(TESTCHIP_RSRCS_DIR)/testchipip/csrc/mm.h \
+	$(TESTCHIP_RSRCS_DIR)/testchipip/csrc/mm.cc \
+	$(TESTCHIP_RSRCS_DIR)/testchipip/csrc/mm_dramsim2.h \
+	$(TESTCHIP_RSRCS_DIR)/testchipip/csrc/mm_dramsim2.cc \
+	$(ROCKETCHIP_RSRCS_DIR)/vsrc/EICG_wrapper.v \
```
jerryz123, line 34: Why do we need to enumerate these here, instead of relying on addResource to copy these in.

sims/vcs/Makefile
```diff
@@ -31,6 +31,22 @@ include $(base_dir)/vcs.mk
...
 
+#########################################################################################
+# simulaton requirements
+#########################################################################################
+SIM_FILE_REQS += \
+	$(ROCKETCHIP_RSRCS_DIR)/vsrc/TestDriver.v
+
+# ignore *.h files since vcs has +incdir
+$(sim_files): $(SIM_FILE_REQS)
+	mkdir -p $(dir $(sim_files))
```
jerryz123, line 42: What is setting +incdir? Is it cleaner to just copy the header files as well?

abejgonzalez: https://github.com/ucb-bar/chipyard/blob/45684f1b875f018a1945476cec9269068e1618ca/vcs.mk#L44

I should clarify this comment. I copy all files to the gen-src area but I don't include the headers in the `sim_files.h`.

sims/common-sim-flags.mk
abejgonzalez: Hmm. So it looks like the `sim_files.common.f` which is generated from the BB files and `sim_files.f` holds the `.v/c/h` files altogether. So it looks like the only reason to have this is to add the non-blackbox files like `EICG_wrapper.v` and the VCS/Verilator specific files. I'll make adjustments tomorrow for this.

generators/utilities/src/main/scala/Simulator.scala
```diff
@@ -1,143 +0,0 @@
...
-        "/vsrc/TestDriver.v",
-      )
-      case None => Seq()
-    })
-
-  def writeBootrom(): Unit = {
-    firrtl.FileUtils.makeDirectory("./bootrom/")
-    writeResource("/testchipip/bootrom/bootrom.rv64.img", "./bootrom/")
-    writeResource("/testchipip/bootrom/bootrom.rv32.img", "./bootrom/")
-    writeResource("/bootrom/bootrom.img", "./bootrom/")
```
timsnyder-siv, line 129: What was this `/bootrom/bootrom.img` in the old code?  I didn't completely understand where it was used in Chipyard before but I knew it wouldn't interfere with what I was doing in FireSim.  I'm only asking because I don't see where this is being maintained in the new makefile version.

abejgonzalez: Good catch. This is Rocket Chips bootrom file: https://github.com/ucb-bar/chipyard/blob/dev/generators/utilities/src/main/resources/bootrom. If you want to use the config fragment in RC (https://github.com/chipsalliance/rocket-chip/blob/b888d0d0eb9975344292d695072e7a9b438cefbb/src/main/scala/subsystem/Configs.scala#L40) you need to have the RC bootrom reside at the top-level in `CHIPYARD/bootrom/bootrom.img`. Ideally RC does what CY does and uses `TargetDirKey`... but for now, this should suffice.

abejgonzalez: I should also note... that 99% of the time people will not use the RC bootrom since the CY one does the same thing for the most part.

sims/common-sim-flags.mk
```diff
@@ -21,3 +21,6 @@ SIM_LDFLAGS = \
 	-lfesvr \
 	-ldramsim \
 	$(EXTRA_SIM_LDFLAGS)
+
+SIM_FILE_REQS += \
+	$(ROCKETCHIP_RSRCS_DIR)/vsrc/EICG_wrapper.v
```
abejgonzalez, line 26: Note: If https://github.com/chipsalliance/rocket-chip/pull/2810 goes through then this is unnecessary.

sims/common-sim-flags.mk
abejgonzalez: Ok. `emulator.cc` requires the other `.cc` and `.h` files to compile for Verilator. VCS I don't think needs it.

generators/utilities/src/main/scala/Simulator.scala
timsnyder-siv: Ignoring it seems fine to me if they do the same thing.  I didn't realize that's where it was coming from or I would have tried to address that in the PR where I added the `TargetDirKey`.

jerryz123: I don't think they quite do the exact same thing anymore. There are some features we support, which the RC bootrom does not, and vice versa.

We should always use the CY bootrom though.

abejgonzalez: In the past we had a `SUB_PROJECT=rocketchip` which setup things to explicitly use RC collateral and not any CY collateral. However, we have drifted from that over time. So I think we can just ignore the RC bootrom.

fpga/Makefile
```diff
@@ -73,6 +73,22 @@ default: $(mcs)
 fpga_dir := $(base_dir)/fpga/fpga-shells/$(FPGA_BRAND)
 fpga_common_script_dir := $(fpga_dir)/common/tcl
 
+#########################################################################################
+# setup misc. sim files
+#########################################################################################
+SIM_FILE_REQS += \
+	$(ROCKETCHIP_RSRCS_DIR)/vsrc/EICG_wrapper.v
+
+# copy files but ignore *.h files in *.f (match vcs)
```
timsnyder-siv, line 82: Are you trying to match vcs here or are you assuming that all FPGA synthesis tools will support `+incdir` like VCS does and you can do the same thing as you're doing for VCS?  If you're really trying to match VCS, why?

sims/vcs/Makefile
```diff
@@ -38,7 +38,7 @@ SIM_FILE_REQS += \
 	$(ROCKETCHIP_RSRCS_DIR)/vsrc/TestDriver.v
 
 # copy files but ignore *.h files in *.f since vcs has +incdir+$(build_dir)
-$(sim_files): $(SIM_FILE_REQS)
+$(sim_files): $(SIM_FILE_REQS) | $(build_dir)
 	mkdir -p $(dir $(sim_files))
```
timsnyder-siv, line 42: Since you have `| $(build_dir)` in the deps for the rule, I don't think you need this `mkdir` anymore.

sims/verilator/Makefile
```diff
@@ -46,12 +46,26 @@ debug: $(sim_debug)
...
+	$(TESTCHIP_RSRCS_DIR)/testchipip/csrc/mm_dramsim2.cc \
+	$(ROCKETCHIP_RSRCS_DIR)/csrc/SimDTM.cc \
+	$(ROCKETCHIP_RSRCS_DIR)/csrc/SimJTAG.cc \
+	$(ROCKETCHIP_RSRCS_DIR)/csrc/remote_bitbang.h \
+	$(ROCKETCHIP_RSRCS_DIR)/csrc/remote_bitbang.cc
 
 # copy files and add -FI for *.h files in *.f
-$(sim_files): $(SIM_FILE_REQS)
+$(sim_files): $(SIM_FILE_REQS) | $(build_dir)
 	mkdir -p $(dir $(sim_files))
```
timsnyder-siv, line 69: Again, I don't think you need this `mkdir` in the recipe if you're doing it in it's own recipe.

sims/verilator/Makefile
```diff
@@ -46,12 +46,26 @@ debug: $(sim_debug)
...
+	$(TESTCHIP_RSRCS_DIR)/testchipip/csrc/testchip_tsi.h \
+	$(TESTCHIP_RSRCS_DIR)/testchipip/csrc/SimDRAM.cc \
+	$(TESTCHIP_RSRCS_DIR)/testchipip/csrc/mm.h \
+	$(TESTCHIP_RSRCS_DIR)/testchipip/csrc/mm.cc \
+	$(TESTCHIP_RSRCS_DIR)/testchipip/csrc/mm_dramsim2.h \
+	$(TESTCHIP_RSRCS_DIR)/testchipip/csrc/mm_dramsim2.cc \
+	$(ROCKETCHIP_RSRCS_DIR)/csrc/SimDTM.cc \
+	$(ROCKETCHIP_RSRCS_DIR)/csrc/SimJTAG.cc \
+	$(ROCKETCHIP_RSRCS_DIR)/csrc/remote_bitbang.h \
+	$(ROCKETCHIP_RSRCS_DIR)/csrc/remote_bitbang.cc
```
timsnyder-siv: ugh.  makefiles are so fun, aren't they.  Ideally when you're commenting on a list of files like this, you'd want to make it harder for someone to make your comment really confusing by shoving new files in at the top of the list that really belong below `verilator.h`.  You might consider restructuring this to use two assigns even though you don't strictly need to use two.  Like this:
```make

SIM_FILE_REQS += \
	$(CHIPYARD_RSRCS_DIR)/csrc/emulator.cc \
	$(ROCKETCHIP_RSRCS_DIR)/csrc/verilator.h

# Don't be tempted to remove the following files.  Even if you are not using them at runtime
# they are needed by emulator.cc to compile the simulator.
SIM_FILE_REQS += \
	$(TESTCHIP_RSRCS_DIR)/testchipip/csrc/SimSerial.cc \
	$(TESTCHIP_RSRCS_DIR)/testchipip/csrc/testchip_tsi.cc \
	$(TESTCHIP_RSRCS_DIR)/testchipip/csrc/testchip_tsi.h \
	$(TESTCHIP_RSRCS_DIR)/testchipip/csrc/SimDRAM.cc \
	$(TESTCHIP_RSRCS_DIR)/testchipip/csrc/mm.h \
	$(TESTCHIP_RSRCS_DIR)/testchipip/csrc/mm.cc \
	$(TESTCHIP_RSRCS_DIR)/testchipip/csrc/mm_dramsim2.h \
	$(TESTCHIP_RSRCS_DIR)/testchipip/csrc/mm_dramsim2.cc \
	$(ROCKETCHIP_RSRCS_DIR)/csrc/SimDTM.cc \
	$(ROCKETCHIP_RSRCS_DIR)/csrc/SimJTAG.cc \
	$(ROCKETCHIP_RSRCS_DIR)/csrc/remote_bitbang.h \
	$(ROCKETCHIP_RSRCS_DIR)/csrc/remote_bitbang.cc
```

I say "ugh" because it would be nice to just put the comment in the middle of the file list like you would in a normal language but since you're using line continuation, it gets gnarly if you try to do that and you end up making two assignments.

fpga/Makefile
```diff
@@ -73,6 +73,22 @@ default: $(mcs)
...
+$(sim_files): $(SIM_FILE_REQS) | $(build_dir)
+	mkdir -p $(dir $(sim_files))
```
timsnyder-siv, line 84: Since you have `| $(build_dir)` in the deps for the rule, I don't think you need this `mkdir` anymore.

timsnyder-siv (commented):
Thanks for reminding me to look back through this.

fpga/Makefile
abejgonzalez: This matches what `GenerateSimFiles` did in the past (filter out headers). I guess another justification for doing this is that #871 is using VCS to do FPGA-level simulation.

build.sbt
```diff
@@ -192,14 +192,10 @@ lazy val chipyard = (project in file("generators/chipyard"))
 
 lazy val tracegen = (project in file("generators/tracegen"))
   .sourceDependency(testchipip, testchipipLib)
-  .dependsOn(rocketchip, sifive_cache, boom, utilities)
+  .dependsOn(rocketchip, sifive_cache, boom)
   .settings(libraryDependencies ++= rocketLibDeps.value)
   .settings(commonSettings)
 
-lazy val utilities = (project in file("generators/utilities"))
-  .sourceDependency(testchipip, testchipipLib)
```
jerryz123, line 200: Nice. Everything I see this I have to remind myself why we need this. Glad to see it getting removed

sims/verilator/Makefile
```diff
@@ -41,6 +43,37 @@ endif
...
+# the following files are needed for emulator.cc to compile
+SIM_FILE_REQS += \
+	$(TESTCHIP_RSRCS_DIR)/testchipip/csrc/SimSerial.cc \
+	$(TESTCHIP_RSRCS_DIR)/testchipip/csrc/testchip_tsi.cc \
+	$(TESTCHIP_RSRCS_DIR)/testchipip/csrc/testchip_tsi.h \
+	$(TESTCHIP_RSRCS_DIR)/testchipip/csrc/SimDRAM.cc \
+	$(TESTCHIP_RSRCS_DIR)/testchipip/csrc/mm.h \
+	$(TESTCHIP_RSRCS_DIR)/testchipip/csrc/mm.cc \
+	$(TESTCHIP_RSRCS_DIR)/testchipip/csrc/mm_dramsim2.h \
+	$(TESTCHIP_RSRCS_DIR)/testchipip/csrc/mm_dramsim2.cc \
```
jerryz123, line 62: Why do these have to be copied over using the Make system for verilator, but not VCS?

abejgonzalez: Because these are used directly in `emulator.cc` whereas VCS doesn't use an `emulator.cc`.

jerryz123 (approved):

sims/verilator/Makefile
abejgonzalez: Actually, for future reference let me clarify more. By default, if you use something like `SimSerial`, the `addResources` will bring in both the `cc` and `v` files. If you don't use it... then those files won't be included. This is the case for VCS... only files necessary for the simulation will be copied over. However, for Verilator, you need to include all `cc/h` files for `emulator.cc` to compile even if you don't necessarily use the corresponding `v` or `scala` source.

PR #882, opened 2021-05-07
[firechip] Memoize Clock RecordMap creation to fix SuperNode

davidbiancolin:
Supernode fails to elaborate using the new harness extensions to clock bridge instantiation.  I got a little carried away with stylistic changes (i can undo those), but this fixes that by permitting multiple requests to `requestClockRecordMap`. 

**Type of change**: bug fix

**Impact**: rtl change


abejgonzalez (approved):
LGTM

PR #902, opened 2021-06-16
Add marshal support to the Dockerfile

NathanTP:
**Related issue**: 

**Impact**: software change

**Release Notes**
The included Dockerfile now supports FireMarshal.

### DIFF
--- a/dockerfiles/Dockerfile
+++ b/dockerfiles/Dockerfile
@@ -15,7 +15,9 @@ RUN apt-get update && \
                sudo \
                ca-certificates \
                keyboard-configuration \
-               console-setup
+               console-setup \
+               bc \
+               unzip
 
 WORKDIR /root
 
@@ -50,6 +52,17 @@ RUN cd chipyard && \
 RUN cd chipyard && \
         export MAKEFLAGS=-"j $(nproc)" && \
         ./scripts/build-toolchains.sh esp-tools 1>/dev/null
+
+
+# Set up FireMarshal
+SHELL ["/bin/bash", "-c"] 
+RUN cd chipyard && \
+        source env.sh && \
+        cd software/firemarshal && \
+        ./init-submodules.sh && \
+        pip3 install -r python-requirements.txt && \
+        marshal build br-base.json 
+        
 
 # Run script to set environment variables on entry
 ENTRYPOINT ["chipyard/scripts/entrypoint.sh"]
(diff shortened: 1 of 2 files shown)


dockerfiles/README.md
```diff
@@ -13,7 +13,8 @@ Build and Deploy the Container
     sudo docker tag <IMAGE_ID> <PATH_NAME>:tag . # to tag the image after the build (ex. 0.0.3)
     sudo docker login # login into the account to push to
     sudo docker push <PATH_NAME>:tag # to push to repo with tag
-    sudo docker run -it <IMAGE_ID> bash # to run an interactive version of the container
+    sudo docker run -it --privileged <IMAGE_ID> bash # to run an interactive version of the container
```
abejgonzalez, line 16: Why does this need `--privileged` access? Is this to mount the FS for modifications?

dockerfiles/Dockerfile
```diff
@@ -51,6 +53,17 @@ RUN cd chipyard && \
         export MAKEFLAGS=-"j $(nproc)" && \
         ./scripts/build-toolchains.sh esp-tools 1>/dev/null
 
+
+# Set up FireMarshal
+SHELL ["/bin/bash", "-c"]
```
abejgonzalez, line 58: Can we just enforce this at the beginning of the Dockerfile instead of right when using FireMarshal? I assume most people will prefer `bash` as default instead of `sh`

dockerfiles/Dockerfile
```diff
@@ -51,6 +53,17 @@ RUN cd chipyard && \
...
 
+
+# Set up FireMarshal
+SHELL ["/bin/bash", "-c"] 
+RUN cd chipyard && \
+        source env.sh && \
+        cd software/firemarshal && \
+        ./init-submodules.sh && \
+        pip3 install -r python-requirements.txt && \
+        marshal build br-base.json
```
abejgonzalez, line 64: To make the Docker image smaller... should we do a `marshal clean br-base.json` to keep the main files but delete the temp files?

dockerfiles/README.md
NathanTP: Yes. I haven't been able to find a way to mount filesystems in docker without --privileged. You'd think FUSE would be fine, but even that doesn't work with the way the kernel handles filesystem access. You can see firesim/firemarshal#199 for the issue tracking this.

dockerfiles/Dockerfile
NathanTP: That would be reasonable, it would save like 500MB. The one that takes a long time is the distro image (br.XXXX.img) which is what we want to avoid.

dockerfiles/Dockerfile
NathanTP: sure

PR #918, opened 2021-07-12
Use the correct harness annotation file path

jwright6323:
**Related issue**: ucb-bar/barstools#104

**Type of change**: bug fix

**Impact**: other

**Release Notes**
- Fixes a bug that ignores the HARNESS_ANNO Makefile variable


jerryz123 (approved):
LGTN

jwright6323:
Hm, I should've checked but why did CI not run on this?

PR #933, opened 2021-08-02
generate the .sbtopt file during repo initialization. ucb-bar#932 and ucb-bar#880 

zslwyuan:
**Related issue**:  ucb-bar#932 and ucb-bar#880

**Impact**: other (change a repo initialization script)

**Release Notes**
change a repo initialization script [init-submodules-no-riscv-tools-nolog.sh](https://github.com/zslwyuan/chipyard/blob/master/scripts/init-submodules-no-riscv-tools-nolog.sh) so it can generate the .sbtopt file in the Chipyard directory during repo initialization and use the absolute path for the SBT workspace variables in .sbtopts. I think it is ok since the initialization script also generates variables in env.sh.

Hope it can adapt to more compilation scenarios. For example, now users should be able to open the Chipyard project directly using Intellij.


scripts/init-submodules-no-riscv-tools-nolog.sh
```diff
@@ -81,3 +81,5 @@ echo "# line auto-generated by init-submodules-no-riscv-tools.sh" >> env.sh
 echo '__DIR="$(dirname "$(readlink -f "${BASH_SOURCE[0]:-${(%):-%x}}")")"' >> env.sh
 echo "PATH=\$__DIR/bin:\$PATH" >> env.sh
 echo "PATH=\$__DIR/software/firemarshal:\$PATH" >> env.sh
+echo "-Dsbt.sourcemode=true" > .sbtopts
```
abejgonzalez, line 84: Can you delete the `.sbtopts` file in the main repo area?

abejgonzalez (changes requested):
LGTM. Just need to delete the original file.

scripts/init-submodules-no-riscv-tools-nolog.sh
abejgonzalez: Once that is done, I'll open this PR in our CI to double-check that things work.

scripts/init-submodules-no-riscv-tools-nolog.sh
```diff
@@ -81,3 +81,5 @@ echo "# line auto-generated by init-submodules-no-riscv-tools.sh" >> env.sh
...
+echo "-Dsbt.workspace=$CHIPYARD_DIR/tools"  >> .sbtopts
```
abejgonzalez, line 85: You should also update the `.gitignore` to not track the file.

abejgonzalez (changes requested):

tymcauley:
LGTM too! This should fix the local changes I have to make to get [Metals](https://scalameta.org/metals/) working.

scripts/init-submodules-no-riscv-tools-nolog.sh
zslwyuan: Thanks and the files are changed accordingly.

abejgonzalez (changes requested):
Ok it looks like the changes break CI since CI copies the `.sbtopts` from the Docker image to the remote build server. I've pushed a commit in #955. Assuming that passes, can you pull that into this PR?

zslwyuan:
> Ok it looks like the changes break CI since CI copies the `.sbtopts` from the Docker image to the remote build server. I've pushed a commit in #955. Assuming that passes, can you pull that into this PR?

Hi @abejgonzalez , just  pull your #955  into this PR ^_^

abejgonzalez (approved):
LGTM since #955 (test PR passed).

abejgonzalez:
Thanks for the PR!

PR #936, opened 2021-08-10
qemu: build without -Werror

ekiwi:
**Type of change**: bug fix

**Impact**: software change


ekiwi:
With this change I am able to build the tools on a Fedora 34 machine with:

```
./scripts/build-toolchains.sh riscv-tools
```

PR #937, opened 2021-08-10
qemu: build without -Werror

ekiwi:
While -Werror is nice for developer builds, it can easily lead to breakage when people use newer compilers which add new warnings.

**Type of change**: bug fix 

**Impact**: tools chain now compiles qemu without -Werror

With this change, I am able to successfully build the risc-v tools on my Fedora 34 machine.


abejgonzalez (approved):
LGTM pending @a0u

PR #963, opened 2021-08-27
Escape $1 in awk command

jwright6323:
I'm not really sure how this ever worked before, but this `$1` needs to be escaped in `common.mk`

**Type of change**: bug fix

**Impact**: other

**Release Notes**
- Fixed awk print bug in common.mk


jerryz123 (approved):

abejgonzalez (approved):
Uh. IDK why this worked before too. Good catch.

PR #965, opened 2021-08-30
Print entire line in `sim_files.f` for Verilator 

abejgonzalez:
**Related issue**: #963 

**Type of change**: bug fix

**Impact**: other

**Release Notes**
The Verilator makefile adds a `-FI` to header files before they get filtered out. Thus, when you print by `$$1` you print each word of a line one-by-one causing parsing issues (ex. `-FI` then `<long_file>` then `-FI` then `<next_file>` etc). This fix removes the `$$1` which tells `awk` to print the entire line at once (this was the previous functionality because the make variable `$1` resolved to nothing before).

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -173,7 +173,7 @@ harness_macro_temp: $(HARNESS_SMEMS_CONF) | top_macro_temp
 # remove duplicate files and headers in list of simulation file inputs
 ########################################################################################
 $(sim_common_files): $(sim_files) $(sim_top_blackboxes) $(sim_harness_blackboxes)
-	awk '{print $$1;}' $^ | sort -u | grep -v '.*\.\(svh\|h\)$$' > $@
+	awk '{print}' $^ | sort -u | grep -v '.*\.\(svh\|h\)$$' > $@
 
 #########################################################################################
 # helper rule to just make verilog files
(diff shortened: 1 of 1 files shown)


jerryz123:
I am confused how there are Verilator bugs that aren't being caught by testing.

jerryz123 (approved):

abejgonzalez:
The initial change (when it was non-escaped) made sense since `$1` resolved probably to nothing. The `$$1` change passing CI is a bit concerning.

jwright6323:
Hm, it looks like CI never ran on #963

common.mk
```diff
@@ -173,7 +173,7 @@ harness_macro_temp: $(HARNESS_SMEMS_CONF) | top_macro_temp
 # remove duplicate files and headers in list of simulation file inputs
 ########################################################################################
 $(sim_common_files): $(sim_files) $(sim_top_blackboxes) $(sim_harness_blackboxes)
-	awk '{print $$1;}' $^ | sort -u | grep -v '.*\.\(svh\|h\)$$' > $@
+	awk '{print}' $^ | sort -u | grep -v '.*\.\(svh\|h\)$$' > $@
```
jwright6323, line 176:
```suggestion
	sort -u $^ | grep -v '.*\.\(svh\|h\)$$' > $@
```
Isn't `awk {print} ...` a no-op?

jwright6323 (changes requested):

abejgonzalez:
> Hm, it looks like CI never ran on #963

I swear I saw the "checks passed" dialog... but I can't find it now...

common.mk
abejgonzalez: Nope this should print each line completely (this is what I saw when I tested it and what this article says: https://www.geeksforgeeks.org/awk-command-unixlinux-examples/)

abejgonzalez: This fix also works and completely avoids the issue :)

jwright6323 (approved):

PR #979, opened 2021-09-13
Ibex integration

schwarz-em:

**Impact**: software change

**Release Notes**

Integrate the SystemVerilog embedded core [Ibex](https://github.com/lowRISC/ibex) into Chipyard. This change adds the necessary Chisel wrapper module and adds Ibex to the Chipyard configurations. Documentation and CI updates are also included.
There was a also a change to the testchipip bootrom that I had to make in order to have Ibex boot correctly within Chipyard. See the pull request here: ucb-bar/testchipip#136


docs/Chipyard-Basics/Chipyard-Components.rst
```diff
@@ -24,6 +24,10 @@ Processor Cores
   An in-order RISC-V core written in System Verilog. Previously called Ariane.
   See :ref:`Generators/CVA6:CVA6 Core` for more information.
 
+**Ibex Core**
+  An in-order RISC-V core writeen in System Verilog.
```
chick, line 28:
```suggestion
  An in-order RISC-V core written in System Verilog.
```

.circleci/run-tests.sh
```diff
@@ -100,6 +100,9 @@ case $1 in
     chipyard-sodor)
         run_asm ${mapping[$1]}
         ;;
+    chipyard-ibex)
+        make run-binary-fast -C $LOCAL_SIM_DIR ${mapping[$1]} BINARY=$RISCV/riscv64-unknown-elf/share/riscv-tests/benchmarks/multiply.riscv
```
abejgonzalez, line 104: Should this instead use a 32b based test? IIRC these tests are compiled for 64b platforms by default.

abejgonzalez, line 104: Should this instead use a 32b based test? IIRC these tests are compiled for 64b platforms by default.

.gitmodules
```diff
@@ -140,3 +140,6 @@
 [submodule "tools/rocket-dsp-utils"]
 	path = tools/rocket-dsp-utils
 	url = https://github.com/ucb-bar/rocket-dsp-utils
+[submodule "generators/ibex"]
+	path = generators/ibex
+	url = https://github.com/schwarz-em/ibex-wrapper
```
abejgonzalez, line 145: This wrapper repo should be added to `ucb-bar`. Then you should change this link.

docs/Chipyard-Basics/Chipyard-Components.rst
abejgonzalez: I would expand on this a bit and say its a 32b core only to distinguish it between all the other cores which are potentially 64 or 32b.

docs/Generators/Ibex.rst
```diff
@@ -0,0 +1,14 @@
+Ibex Core
+====================================
+
+`Ibex <https://github.com/lowRISC/ibex>`__ is a parameterizable RV32 embedded core written in SystemVerilog, currently maintained by lowRISC.
```
abejgonzalez, line 4: Can you expand on the RV32... RV32IMC.

abejgonzalez: Can you link `lowRISC`.

docs/Generators/Ibex.rst
```diff
@@ -0,0 +1,14 @@
+Ibex Core
+====================================
+
+`Ibex <https://github.com/lowRISC/ibex>`__ is a parameterizable RV32 embedded core written in SystemVerilog, currently maintained by lowRISC.
+The `Ibex core` is wrapped in an `Ibex tile` so it can be used with the `Rocket Chip SoC generator`.
+The core exposes a custom memory interface, interrupt ports, and other misc. ports that are connected from within the tile to TileLink buses and other parameterization signals.
+
+.. Warning:: The Ibex mtvec register is 256 byte aligned. When writing/running tests, ensure that the trap vector is also 256 byte aligned.
+
+.. Warning:: The Ibex reset vector is located at 0x80.
```
abejgonzalez, line 10: Isn't the more appropriate thing to say is that the reset vector is located at ADDRESS+OFFSET where OFFSET is 0x80.

generators/chipyard/src/main/scala/config/IbexConfigs.scala
```diff
@@ -0,0 +1,14 @@
...
+
+import freechips.rocketchip.config.{Config}
+
+// ---------------------
+// Ibex Configs
+// ---------------------
+
+class IbexConfig extends Config(
+  new chipyard.config.WithBootROM ++               // Ibex reset vector is at 0x80
+  new ibex.WithNIbexCores(1) ++                    // single Ibex core
```
abejgonzalez, line 13: To confirm... this has all the default params that the Ibex core normally comes with (ex. 4 PMP registers, etc)?

generators/chipyard/src/main/scala/config/IbexConfigs.scala
```diff
@@ -0,0 +1,14 @@
...
+import chisel3._
+
+import freechips.rocketchip.config.{Config}
+
+// ---------------------
+// Ibex Configs
+// ---------------------
+
+class IbexConfig extends Config(
+  new chipyard.config.WithBootROM ++               // Ibex reset vector is at 0x80
```
abejgonzalez, line 12: This comment is a bit unnecessary esp. since you made the modifications to all cores. I would remove the comment.

docs/Generators/Ibex.rst
```diff
@@ -0,0 +1,14 @@
...
+The `Ibex core` is wrapped in an `Ibex tile` so it can be used with the `Rocket Chip SoC generator`.
+The core exposes a custom memory interface, interrupt ports, and other misc. ports that are connected from within the tile to TileLink buses and other parameterization signals.
+
+.. Warning:: The Ibex mtvec register is 256 byte aligned. When writing/running tests, ensure that the trap vector is also 256 byte aligned.
+
+.. Warning:: The Ibex reset vector is located at 0x80.
+
+While the core itself is not a generator, we expose the same parameterization that the Ibex core provides so that all supported Ibex configurations are available.
+
+For more information, see the `GitHub repository <https://github.com/lowRISC/ibex>`__.
```
abejgonzalez, line 14:
```suggestion
For more information, see the `GitHub repository for Ibex <https://github.com/lowRISC/ibex>`__.
```

abejgonzalez (changes requested):
Couple of comments:

- Have you tried a multi-core Ibex configuration? If it doesn't work then we should prevent there being multiple cores from being instantiated.
- Have you tried a heterogeneous setup? Ibex + Rocket... etc. Similar to above, then we should prevent this using something like `requires`

abejgonzalez (changes requested):
Looks like the CI is failing the tutorial setup and the core tests. Can you rebase these changes on the most recent `dev` branch? Also, can you fix the tutorial setup, you probably just need to re-generate the git diff patch.

generators/chipyard/src/main/scala/config/IbexConfigs.scala
jerryz123: Wouldn't this WithBootROM be in the abstract config? Its not necessary here?

jerryz123: Also we probably don't want to modify the bootrom for all cores just to accomodate ibex...

abejgonzalez, line 12: I also this breaks all other cores since you need to change the hang address and the base address of the bootrom.

See: https://github.com/ucb-bar/testchipip/pull/136#issuecomment-918468095
Params you need to change in the `WithBootROM` fragment: https://github.com/chipsalliance/rocket-chip/blob/c4c43134e6cb604ecc329dd8e32614b8a976bae7/src/main/scala/devices/tilelink/BootROM.scala#L17-L19

By changing the WithBootROM, you might also need to fix up some of the Ibex core code as well so that it parses the hang or address params properly.

.gitmodules
abejgonzalez: Also can you open that new repo with a PR with all the new code so that I can comment/review that code.

generators/chipyard/src/main/scala/config/IbexConfigs.scala
schwarz-em: Yes, I have the correct default values for the parameters in the actual tile code.

docs/Generators/Ibex.rst
schwarz-em: Yeah, that sounds more correct. I'll fix it.

.circleci/run-tests.sh
schwarz-em: You're correct. I should have looked more carefully. I'll find a 32b test to change it to

abejgonzalez:
Looks like the CI is still failing, can you double check that this config builds in Verilator properly?

.circleci/config.yml
```diff
@@ -252,6 +252,13 @@ jobs:
                 group-key: "group-cores"
                 project-key: "chipyard-sodor"
                 timeout: "30m"
+    chipyard-ibex-run-tests:
```
abejgonzalez, line 255: What's going on with all the `.circleci` files being deleted?

.circleci/run-tests.sh
```diff
@@ -100,6 +100,9 @@ case $1 in
     chipyard-sodor)
         run_asm ${mapping[$1]}
         ;;
+    chipyard-ibex)
+        run_bmark ${mapping[$1]}
```
abejgonzalez, line 104: Wait... why does this pass? Shouldn't the bmarks be RV64... not RV32 (and thus should break)?

.github/README.md
```diff
@@ -0,0 +1,137 @@
+Chipyard Continuous Integration (CI)
```
abejgonzalez, line 1: This file shouldn't exist (was renamed in the most recent `dev` work).

abejgonzalez (changes requested):
Since there is no Ibex PR I'll put the comments here:
- Can you improve the Ibex README to be more like the https://github.com/ucb-bar/nvdla-wrapper and https://github.com/ucb-bar/cva6-wrapper
- Can you add a comment in the Makefile for why you need the `sed` fix/hack.

abejgonzalez (approved):
LGTM. Pending CI passing.

PR #983, opened 2021-09-16
Mergify Integration

abejgonzalez:
**Related issue**: 

**Impact**: other

**Release Notes**
This adds a bot to handle some PR request merging automatically and add `dev` comments. We can also add future functionality to make the bot backport changes to `master` if need be. For this to work, the PR needs to be backported to the `master` branch (since the configuration file is only picked up on the default branch). Maybe we need to quickly rediscuss having `dev` be the main branch because now 2 things depend on it, the GH issue/PR templates and this.

Changes
- Add CI success job that runs at the end
- Add mergify bot
  - Merge when all reviews approve
  - Merge when its just docs and the CI passes for just docs
  - Label docs changes
  - Add comment to change base branch if it's not dev targeted

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,0 +1,44 @@
+pull_request_rules:
+    - name: automatic merge on all ci success and review
+      conditions:
+          - "check-success=ci/circleci: success"
+          - "#approved-reviews-by>=1"
+          - base=dev
+          - label!="DO NOT MERGE"
+      actions:
+          merge:
+              method: merge
+              strict: smart
+              strict_method: merge
+
+    - name: automatically merge on just docs ci success and review (just docs)
+      conditions:
+          - check-success=ci/circleci: documentation-check
+          - -files~=^(!?docs/)
+          - "#approved-reviews-by>=1"
+          - base=dev
+          - label!="DO NOT MERGE"
+      actions:
+          merge:
+              method: merge
+              strict: smart
+              strict_method: merge
+
+    - name: label docs changes
+      conditions:
+          - -files~=^(!?docs/)
+      actions:
+          label:
+              add:
+                - docs
+
+    - name: add comment to warn that base branch isn't dev (ignore backports and releases)
+      conditions:
+          - -base=dev
+          - -body~=Dev -> Master
+          - -title~=Dev -> Master
+          - label!="backport"
+      actions:
+          comment:
+              message: |
+                  Please change the base branch of the PR to dev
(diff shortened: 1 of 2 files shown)


davidbiancolin:
Pleaseeeee lets make the development branch master.

.mergify.yml
```diff
@@ -0,0 +1,56 @@
+pull_request_rules:
+    - name: automatic merge on all ci success and review
+      conditions:
+          - "check-success=success"
```
abejgonzalez, line 4:
```suggestion
          - "check-success=ci/circleci: success"
```

.mergify.yml
```diff
@@ -0,0 +1,56 @@
+pull_request_rules:
+    - name: automatic merge on all ci success and review
```
davidbiancolin, line 2: I'm not sure we want to do this by default. Probably should be an agenda item.

davidbiancolin: I'd be more willing if the tag "DO NOT MERGE" was applied by default.

.mergify.yml
```diff
@@ -0,0 +1,56 @@
...
+          - "#changes-requested-reviews-by=0"
+          - base=dev
+          - label!="DO NOT MERGE"
+      actions:
+          merge:
+              method: merge
+              strict: smart
+              strict_method: merge
+
+    - name: automatically merge on just docs ci success and review (just docs/md files)
```
davidbiancolin, line 16: This is pretty cool.

.mergify.yml
abejgonzalez: Assuming all necessary reviews are "approve" and the CI passes, then why shouldn't we merge? I assume an author would add the "DO NOT MERGE" label if they don't want it to be merged...

davidbiancolin: I just think in practice people (me, at least) don't apply that label, because people don't expect others will click the green button for them. Additionally, some people (me) give conditional approval but still expect changes. 

Of course, these criticisms can be addressed with a process change -- we should talk about those at a chipyard meeting.

abejgonzalez: The `change-requested` flag requires that only reviews without changes requested will go through (probably need to double check though). Another option is to add a `MERGE` label which tells the "bot" to merge when ready.

abejgonzalez:
We will try out the `dev` branch as a main branch at some point in the future. Things to check on:

1. Warn Chipyard/FireSim users that they are not on the most latest release branch (prompt where they have to type : yes - have a bypass for CI)
2. Make sure that the docs always point to the latest release branch (not to the dev documents)
3. Rename `master` to `main`

abejgonzalez:
Newer version is #1157

PR #988, opened 2021-09-22
Fix readlink on macOS

tymcauley:
**Related issue**: N/A

Use GNU readlink on macOS, since BSD readlink doesn't work in the same way. This was resulting in the auto-generated `.sbtopts` file having the wrong contents (in the `sbt.workspace` setting).

**Type of change**: bug fix

**Impact**: script change

**Release Notes**
Fix `.sbtopts` generation on macOS

### DIFF
--- a/scripts/build-toolchains.sh
+++ b/scripts/build-toolchains.sh
@@ -6,9 +6,16 @@
 set -e
 set -o pipefail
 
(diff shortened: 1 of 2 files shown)


abejgonzalez (approved):
LGTM

PR #998, opened 2021-10-04
Chisel 3.5 Bump

abejgonzalez:
**Related issue**: https://github.com/firesim/firesim/pull/836

**Type of change**: new feature + other enhancement

**Impact**: rtl change

**Release Notes**


abejgonzalez:
TODO:

- The prior work refactoring `barstools` so that it is all in one area (`src/main/scala`) creates a circular dependency in CY. `chipyard` needs `iocells` but it is packaged into `tapeout` which then depends on `chipyard`.
- Need to check that adding a virtual io field to Hwacha and Sodor doesn't cause issues (this was a quick fix to move on with other issues).

abejgonzalez:
> * The prior work refactoring `barstools` so that it is all in one area (`src/main/scala`) creates a circular dependency in CY. `chipyard` needs `iocells` but it is packaged into `tapeout` which then depends on `chipyard`. 

This is hacked around... is there a better way to do this? Should something change in `barstools`?

PR #999, opened 2021-10-06
Dev use GitHub actions plus circle ci

chick:
**Related issue**: For comparison purposes to PR #997 

**Type of change**: new CI based on GA, comparing it to circleci

**Impact**: Ci changes

**Release Notes**
NA

PR #1004, opened 2021-10-08
Testing GH-A Improvements

abejgonzalez:
**Related issue**: #999 

**Impact**: other

**Release Notes**
Couple new improvements
- Tries to emulate the "Run from failed" that CircleCI gives to not rerun passing jobs.
- Small fix to `defaults.sh` when running a script with multiple arguments
- Tries to dedup as much as possible things within the workflow
- Switch to using build server `ferry` as a GH-A runner (significantly cleans up scripts)
Couple workarounds
- No support for `if:` clause in composite actions. This requires me to build/cache both toolchains at once... in reality this penalty is paid once every so often that the toolchain changes.

### DIFF
--- a/.github/actions/toolchain-build/action.yml
+++ b/.github/actions/toolchain-build/action.yml
@@ -8,12 +8,13 @@ runs:
       run: .github/scripts/create-hash.sh
       shell: bash
 
+    # since "hashFiles" function differs on self-hosted vs GH-A machines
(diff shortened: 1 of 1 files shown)


abejgonzalez:
Current issue is that `hashFiles` (needed for retrieving the toolchain) differs on a GH-A container vs on the self-hosted machine.

PR #1028, opened 2021-11-01
Bump marshal

NathanTP:
**Related issue**: 
firesim/firemarshal#207

**Type of change**: bug fix

**Impact**: software change

**Release Notes**
Update to the latest FireMarshal to get a number of bug fixes. Most importantly, this changes marshal to use a reasonable degree of parallelism for builds rather than unlimited which could cause crashes on some systems. See firesim/firemarshal#207


abejgonzalez (approved):
LGTM. CircleCI passes while GH-A almost passes (I'll have to look into this later)

PR #1030, opened 2021-11-03
Fix custom MODEL feature

jerryz123:
Previously, the TestDriver would set MODEL to TestHarness if unset. This fixes the feature to let us set MODEL from the Makefile.

**Related issue**: 

**Type of change**: bug fix

**Impact**: rtl change

**Release Notes**


abejgonzalez (approved):
LGTM

PR #1029, opened 2021-11-02
Modify GH Issues (All Feedback Welcome)

abejgonzalez:
**Related issue**: 

**Type of change**: new feature

**Impact**: other

**Release Notes**
This PR cleans up the GH issues to use the new issues template feature of GH and redirects questions/support to the mailing list / other forums (hopefully to move some question traffic there and just have GH serve as a features/bug tracker). An example of this working is here https://github.com/abejgonzalez/chipyard/issues when you open a new issue.

GH Issues Template Feature: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository

Changes
- Only support bug reports and feature requests (deletes the other and questions issue templates)
  - The issue templates now have checkboxes requiring people to read the docs / search things before submitting
  - Also, has other required boxes to fill out
- Show the mailing list / RC / Chisel support links right before creating an issue

PR #1035, opened 2021-11-05
Update README

abejgonzalez:
**Related issue**: 

**Type of change**: bug fix + other enhancement

**Impact**: other

**Release Notes**
Fixes issue where README shown would default to the `.github/README.md` instead of the top-level `README.md`. Additionally adds a new section at the top of the `README.md` for quick links. My reasoning for adding this is that it makes it extremely easy to find the link to the mailing list (called forum here) for people who don't care to read down the `README.md`. This aligns with #1029 where I pointed to the mailing list more.


jerryz123:
What is the README bug? Having multiple READMEs should be supported.

abejgonzalez:
> What is the README bug? Having multiple READMEs should be supported.

If there is a README in the `.github` area, it will show that one by default instead of the one in the top-level of the repo.

jerryz123:
Its useful to have README.md in the subdirectories, as it shows up when you browse there in Github, I think this may be a one-off rendering bug, rather than a repo bug.

Here's a screenshot of what I mean. If you rename it this will not longer show up.

image

abejgonzalez:
> Its useful to have README.md in the subdirectories, as it shows up when you browse there in Github, I think this may be a one-off rendering bug, rather than a repo bug.
> 
> Here's a screenshot of what I mean. If you rename it this will not longer show up.
> 
> image

I agree, but I think this is an issue with the priority of the `.github` vs the root directory `README.md` file... I would look at the two PR commits to see the difference.

jerryz123 (approved):

PR #1040, opened 2021-11-08
Increase default SerialTL width to 32 

jerryz123:
This should improve the performance of `run-binary`, `run-asm-tests`, `run-bmark-tests`, at the expense of making the default config less tapeout-friendly. However, its expected that tapeout users will need to adjust the SerialTL settings for a chip anyways, so this is a small cost.

The FireSim configs are not affected. The SerialTL width is set to be 4, as before, so no AGFI regeneration is needed.

**Type of change**:  other enhancement

**Impact**: rtl change 

**Release Notes**

Increased default SerialTL width to 32 from 4.


abejgonzalez (approved):
This should be fine. LGTM.

jerryz123:
Also do you handle the edge case where an integer inst older than a vector inst requests a replay while the vector inst is in the unroller? Does the pipeline flush the unroller to replay the older integer inst?

PR #1041, opened 2021-11-12
Cleanup how to override SBT + `make help` documentation update

abejgonzalez:
**Related issue**: 

**Impact**: other

**Release Notes**
`ENABLE_SBT_THIN_CLIENT` flag is best used with a pre-installed SBT (aka best with the SBT script). This allows you to just override the SBT binary/script with `SBT_BIN` and have the `--client` flag be handled by Chipyard's Makefile. This also adds more `make help` documentation, fixes an issue with the SBT server not being depended on, and removes unnecessary `?=`s and `override`s in `variables.mk` (this was done to properly enforce that the only way to modify the CY build system SBT is to change `SBT_BIN`).

Side note: When using the SBT script to launch things, the thin client acts nicely; properly exits on errors, displays colors, etc.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -18,7 +18,7 @@ HELP_COMPILATION_VARIABLES += \
 "   EXTRA_SIM_LDFLAGS      = additional LDFLAGS for building simulators" \
 "   EXTRA_SIM_SOURCES      = additional simulation sources needed for simulator" \
 "   EXTRA_SIM_REQS         = additional make requirements to build the simulator" \
-"   ENABLE_SBT_THIN_CLIENT = if set, use sbt's experimental thin client (works best when overridding SBT_BIN with sbt script)" \
+"   ENABLE_SBT_THIN_CLIENT = if set, use sbt's experimental thin client (works best when overridding SBT_BIN with the mainline sbt script)" \
 "   EXTRA_CHISEL_OPTIONS   = additional options to pass to the Chisel compiler" \
 "   EXTRA_FIRRTL_OPTIONS   = additional options to pass to the FIRRTL compiler"
 
@@ -41,13 +41,14 @@ NUMA_PREFIX = $(if $(filter $(NUMACTL),0),,$(shell $(base_dir)/scripts/numa_pref
 
 #----------------------------------------------------------------------------
 HELP_COMMANDS += \
-"   run-binary             = run [./$(shell basename $(sim))] and log instructions to file" \
-"   run-binary-fast        = run [./$(shell basename $(sim))] and don't log instructions" \
-"   run-binary-debug       = run [./$(shell basename $(sim_debug))] and log instructions and waveform to files" \
-"   verilog                = generate intermediate verilog files from chisel elaboration and firrtl passes" \
-"   firrtl                 = generate intermediate firrtl files from chisel elaboration" \
-"   run-tests              = run all assembly and benchmark tests" \
-"   launch-sbt             = start sbt terminal"
+"   run-binary                  = run [./$(shell basename $(sim))] and log instructions to file" \
+"   run-binary-fast             = run [./$(shell basename $(sim))] and don't log instructions" \
+"   run-binary-debug            = run [./$(shell basename $(sim_debug))] and log instructions and waveform to files" \
+"   verilog                     = generate intermediate verilog files from chisel elaboration and firrtl passes" \
+"   firrtl                      = generate intermediate firrtl files from chisel elaboration" \
+"   run-tests                   = run all assembly and benchmark tests" \
+"   launch-sbt                  = start sbt terminal" \
+"   {shutdown,start}-sbt-server = shutdown or start sbt server if using ENABLE_SBT_THIN_CLIENT" \
 
 #########################################################################################
 # include additional subproject make fragments
@@ -105,7 +106,7 @@ $(FIRRTL_FILE) $(ANNO_FILE): generator_temp
 	@echo "" > /dev/null
 
 # AG: must re-elaborate if cva6 sources have changed... otherwise just run firrtl compile
-generator_temp: $(SCALA_SOURCES) $(sim_files) $(EXTRA_GENERATOR_REQS)
+generator_temp: $(SCALA_SOURCES) $(sim_files) $(SCALA_BUILDTOOL_DEPS) $(EXTRA_GENERATOR_REQS)
 	mkdir -p $(build_dir)
 	$(call run_scala_main,$(SBT_PROJECT),$(GENERATOR_PACKAGE).Generator,\
 		--target-dir $(build_dir) \
@@ -279,6 +280,7 @@ SBT_COMMAND ?= shell
 launch-sbt:
 	cd $(base_dir) && $(SBT_NON_THIN) "$(SBT_COMMAND)"
 
+.PHONY: check-thin-client
 check-thin-client:
 ifeq (,$(ENABLE_SBT_THIN_CLIENT))
 	$(error ENABLE_SBT_THIN_CLIENT not set.)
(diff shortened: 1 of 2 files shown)


variables.mk
```diff
@@ -175,7 +175,8 @@ override SCALA_BUILDTOOL_DEPS += $(SBT_THIN_CLIENT_TIMESTAMP)
 SBT_CLIENT_FLAG = --client
 endif
 
-SBT ?= java $(JAVA_TOOL_OPTIONS) -jar $(ROCKETCHIP_DIR)/sbt-launch.jar $(SBT_OPTS) $(SBT_CLIENT_FLAG)
+SBT_BIN ?= java $(JAVA_TOOL_OPTIONS) -jar $(ROCKETCHIP_DIR)/sbt-launch.jar $(SBT_OPTS)
+SBT ?= $(SBT_BIN) $(SBT_CLIENT_FLAG)
```
jerryz123, line 179: Should `SBT_OPTS` be part of `SBT_BIN` or `SBT`? I would expect `SBT_OPTS` to be the same for various settings of `SBT_BIN`

common.mk
```diff
@@ -18,7 +18,7 @@ HELP_COMPILATION_VARIABLES += \
 "   EXTRA_SIM_LDFLAGS      = additional LDFLAGS for building simulators" \
 "   EXTRA_SIM_SOURCES      = additional simulation sources needed for simulator" \
 "   EXTRA_SIM_REQS         = additional make requirements to build the simulator" \
-"   ENABLE_SBT_THIN_CLIENT = if set, use sbt's experimental thin client (works best with sbtn or sbt script)" \
+"   ENABLE_SBT_THIN_CLIENT = if set, use sbt's experimental thin client (works best when overridding SBT_BIN with sbt script)" \
 "   EXTRA_CHISEL_OPTIONS   = additional options to pass to the Chisel compiler" \
```
jerryz123, line 22: Can we add some help text here for `SBT_BIN`?

jerryz123 (changes requested):

variables.mk
jerryz123: Hmm this also drops `JAVA_TOOL_OPTIONS` when `SBT_BIN` is custom set. Is that intended?

abejgonzalez: When using the SBT script the variables in the `SBT_OPTS` makefile variable are passed to the script in two ways:
1. the `.sbtopts` file in the top-level
2. the env. variable `SBT_OPTS` (which is assumed to be set by users)

As for JAVA_TOOL_OPTIONS, I need to double check...

abejgonzalez: I wonder if Chipyard should just automatically download SBT (its just a .jar and a bash script) and use it by default...
then we wouldn't have to have all this makefile stuff around it.

abejgonzalez: Ok. Cleaned this up a bit but I can confirm that doing `export var` exports the variable to the recipe. Therefore, `JAVA_TOOL_OPTIONS` is passed to `java` no matter what.

jerryz123 (approved):

tymcauley:
So, the recommended usage of thin-client mode would be:
```
ENABLE_SBT_THIN_CLIENT=1 SBT_BIN=sbt make ...
```
Should we automatically set `SBT_BIN` to `sbt` when `ENABLE_SBT_THIN_CLIENT` is set?

Also, is there any downside to always using `SBT_BIN=sbt` as the default case, and just never using `sbt-launch.jar`? On one hand it means people need an extra binary on their `PATH` (`sbt` rather than just `java`), but I think that all of the Chipyard setup docs ensure that people have `sbt` installed.

abejgonzalez:
I think outside of me, you (@tymcauley) and Jerry are the two main users (at least the two main ones who I know about) so I would love to hear your feedback as always.

> So, the recommended usage of thin-client mode would be:
> 
> ```
> ENABLE_SBT_THIN_CLIENT=1 SBT_BIN=sbt make ...
> ```
> 

With that being said, I have both `ENABLE_SBT_THIN_CLIENT` and `SBT_BIN=sbt` set in my `.bashrc`. You could also set it the way that you specified. 

> Should we automatically set `SBT_BIN` to `sbt` when `ENABLE_SBT_THIN_CLIENT` is set?
> Also, is there any downside to always using `SBT_BIN=sbt` as the default case, and just never using `sbt-launch.jar`? On one hand it means people need an extra binary on their `PATH` (`sbt` rather than just `java`), but I think that all of the Chipyard setup docs ensure that people have `sbt` installed.

I have this debate every so often. While the installed `sbt` (which really just is just a bash script) could be used instead of the `sbt-launch.jar`, it would add one more thing for Chipyard to version against. My concern is mainly around... what version of the `sbt` script is being used, does it support `--client`, and how well does this script interact with a different version of `sbt` specified [here](https://github.com/ucb-bar/chipyard/blob/master/project/build.properties). If I have clearer answers to those concerns then I would be more than happy to switch (since it would "freeze" how you might set `sbt/java` args).

jerryz123:
If we make `SBT_BIN=sbt` the default case, we'd need to update all the setup instructions to make sure the user gets the right SBT on their `PATH`. While this is fine, I'd prefer that to go into a separate PR later, as we should dogfood this feature a bit more before changing default behavior.

tymcauley:
Regarding the `sbt` version, I don't really know what the minimum-supported-version should be, but as far as I can tell, there are two `sbt` versions to care about:

- The `sbt` Java binary, which is what actually runs the build. The version of this is set in `project/build.properties`.
- The version of the launcher script. This is the version of the `sbt` executable that you have installed on your machine. This is basically a shell script which downloads the desired `sbt` Java binary, which must match the version in `project/build.properties`.

I don't know what the relationship looks like between these two programs and their supported versions. Not sure where to find that.

Regarding `SBT_OPTS`, it looks to me like we're now depending on users to copy the settings from `JAVA_TOOL_OPTIONS` into `SBT_OPTS` when people set `SBT_BIN=sbt`, is that right?

I also like the idea of addressing setup/usage for this use-case in a future PR. I'm very happy to see that using `SBT_BIN=sbt` will now properly communicate the exit code of the `sbt` server task 🙂 Seems like this should make it a more broadly-useful feature now.

abejgonzalez:
> Regarding the `sbt` version, I don't really know what the minimum-supported-version should be, but as far as I can tell, there are two `sbt` versions to care about:
> 
> * The `sbt` Java binary, which is what actually runs the build. The version of this is set in `project/build.properties`.
> * The version of the launcher script. This is the version of the `sbt` executable that you have installed on your machine. This is basically a shell script which downloads the desired `sbt` Java binary, which must match the version in `project/build.properties`.
> 
> This was the most useful link I could find on that subject: [sbt/sbt#4503 (comment)](https://github.com/sbt/sbt/issues/4503#issuecomment-450586222)
> 
> I don't know what the relationship looks like between these two programs and their supported versions. Not sure where to find that.

Yup. Looks like we are referencing the same thing and having two versions to track (the script vs the `build.properties` file).

> Regarding `SBT_OPTS`, it looks to me like we're now depending on users to copy the settings from `JAVA_TOOL_OPTIONS` into `SBT_OPTS` when people set `SBT_BIN=sbt`, is that right?

The `JAVA_TOOL_OPTIONS` is `export`ed in the Makefile so it should be automatically passed as an environment variable to the `sbt` script (and thus to the underlying `java` bin). So my experience, no... you shouldn't have to change anything (you still set `JAVA_TOOL_OPTIONS` and `SBT_OPTS` normally).

> I also like the idea of addressing setup/usage for this use-case in a future PR. I'm very happy to see that using `SBT_BIN=sbt` will now properly communicate the exit code of the `sbt` server task 🙂 Seems like this should make it a more broadly-useful feature now.

Yup, I plan on using this (virtually) all the time. Although there still seems to be issues with nested SBT calls that @davidbiancolin noticed (frequently used in FireSim-land)

PR #1045, opened 2021-11-15
Fix MMIO IOBinder

jwright6323:
**Related issue**: None

**Type of change**: bug fix

**Impact**: rtl change

**Release Notes**
- Fixes a bug in the MMIO port IO binder when a mem port is not present

### DIFF
--- a/generators/chipyard/src/main/scala/IOBinders.scala
+++ b/generators/chipyard/src/main/scala/IOBinders.scala
@@ -297,7 +297,7 @@ class WithAXI4MMIOPunchthrough extends OverrideLazyIOBinder({
   (system: CanHaveMasterAXI4MMIOPort) => {
     implicit val p: Parameters = GetSystemParameters(system)
     val clockSinkNode = p(ExtBus).map(_ => ClockSinkNode(Seq(ClockSinkParameters())))
-    clockSinkNode.map(_ := system.asInstanceOf[HasTileLinkLocations].locateTLBusWrapper(MBUS).fixedClockNode)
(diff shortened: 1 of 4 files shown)


generators/chipyard/src/main/scala/IOBinders.scala
```diff
@@ -297,7 +297,7 @@ class WithAXI4MMIOPunchthrough extends OverrideLazyIOBinder({
   (system: CanHaveMasterAXI4MMIOPort) => {
     implicit val p: Parameters = GetSystemParameters(system)
     val clockSinkNode = p(ExtBus).map(_ => ClockSinkNode(Seq(ClockSinkParameters())))
-    clockSinkNode.map(_ := system.asInstanceOf[HasTileLinkLocations].locateTLBusWrapper(MBUS).fixedClockNode)
```
jerryz123, line 300: I thought the axi4 port takes its clock from the MBus, not the SBus?

jwright6323: It was on the sbus in the generated RTL I looked at. This seems to confirm, but I could be wrong:
https://github.com/chipsalliance/rocket-chip/blob/86a2f2cca699f149bcc082ef2828654a0a4e3f4b/src/main/scala/subsystem/Ports.scala#L147

jerryz123 (approved):

jwright6323:
@jerryz123 do you know why the tests are getting canceled?

abejgonzalez (changes requested):
As you can tell, CircleCI isn't working right now (I submitted a help ticket to get this resolved but it might take a while). Since GitHub Actions CI is still working, could you add your test to GH-A then we can merge?

jwright6323:
Thanks for picking up the slack @abejgonzalez

PR #1063, opened 2021-12-16
Don't initialize nvdla-workload with init-submodules

jerryz123:
nvdla-workload is very large, results in long init times. Make users initialize this explicitly (like we do with gemmini-software)

**Impact**: software change

**Release Notes**

NVDLA-workload submodule is not longer initialized with init-submodules script

### DIFF
--- a/scripts/init-software.sh
+++ b/scripts/init-software.sh
@@ -1,0 +1,14 @@
+#!/usr/bin/env bash
+# exit script if any command fails
+set -e
+set -o pipefail
+
+# Enable submodule update for software submodules
+git config --unset submodule.software/nvdla-workload.update || :
+git config --unset submodule.software/coremark.update || :
+git config --unset submodule.software/spec2017.update || :
+
+# Initialize local software submodules
+git submodule update --init --recursive software/nvdla-workload
+git submodule update --init --recursive software/coremark
+git submodule update --init --recursive software/spec2017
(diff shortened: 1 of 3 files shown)


scripts/init-software.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+# exit script if any command fails
+set -e
+set -o pipefail
+
+# Enable submodule update for software submodules
+git config --unset submodule.software/nvdla-workload.update || :
```
abejgonzalez, line 7: Shouldn't this be `git config --local --unset` like what the `init-submodules.sh` script does?

abejgonzalez (changes requested):
Can you add some documentation to https://chipyard.readthedocs.io/en/dev/Software/index.html to indicate that there is a `software/` dir and that you need to initialize it to use it?

scripts/init-software.sh
abejgonzalez: https://github.com/ucb-bar/chipyard/blob/02adf86b8261ca4b5baf35b3257e15cdbabd7f87/scripts/init-submodules-no-riscv-tools-nolog.sh#L54

jerryz123: I believe the `--local` flag is default.

> These values are specific to that single repository, and represent passing the --local option to git config. If you don’t specify which level you want to work with, this is the default.

abejgonzalez (approved):
LGTM

PR #1094, opened 2022-01-20
Document simulator-timeout settings

tymcauley:
**Related issue**: N/A

**Impact**: other

**Release Notes**

Document simulator-timeout settings.

### DIFF
--- a/variables.mk
+++ b/variables.mk
@@ -25,7 +25,9 @@ HELP_PROJECT_VARIABLES = \
 
 HELP_SIMULATION_VARIABLES = \
 "   BINARY                 = riscv elf binary that the simulator will run when using the run-binary* targets" \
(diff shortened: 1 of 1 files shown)


jerryz123:
Hmm, I have been using `timeout_cycles` without issue, without needing this change....

tymcauley:
Ah, interesting, let me give that a shot.

tymcauley:
Yup, you're right, setting those variables on the command-line works for me. Should we still add them to the help text, so they're more discoverable?

jerryz123:
Yeah, adding them to the help text is a good idea.

tymcauley:
Sounds good, I'll update this branch to only contain the help text info

tymcauley:
@jerryz123 I've cleaned this up, let me know what you think.

jerryz123 (approved):

tymcauley:
Thanks!

PR #1109, opened 2022-02-07
Have PRCI control registers be clocked by the bus they hang off of

jerryz123:
This was technically correct before if the CBUS clock was the implicit clock. but this change makes it correct when that is not the case

**Related issue**: 

**Type of change**: bug fix 

**Impact**: rtl change 

**Release Notes**


abejgonzalez (approved):
LGTM

PR #1114, opened 2022-02-11
Check that Chipyard is running a tagged release

abejgonzalez:
**Related issue**: 

**Impact**: other

**Release Notes**
As a part of moving from developers working on `dev` to `main`, have a check that verifies that the user is running on a tagged release in `init-submodules....sh`. This can be bypassed by adding the `--skip-validate` flag (for CI and other automation).


scripts/init-submodules-no-riscv-tools-nolog.sh
```diff
@@ -17,6 +45,21 @@ if [ "$MINGIT" != "$(echo -e "$MINGIT\n$MYGIT" | sort -V | head -n1)" ]; then
...
 
+# before doing anything verify that you are on a release branch/tag
+set +e
+tag=$(git describe --exact-match --tags)
+tag_ret_code=$?
+set -e
+if [ $tag_ret_code -ne 0 ]; then
+    if [ "$SKIP_VALIDATE" = false ]; then
+        read -p "WARNING: You are not on a tagged release of Chipyard. Type \"ok\" to continue: " validate
+        [[ $validate == [oO][kK] ]] || exit 3
```
jerryz123, line 56: Recommend making this more unix-y, wait for `y` or `Y` instead.

jerryz123 (changes requested):

jerryz123 (approved):
This is ok.

Issue #1049, opened 2021-11-18
Update QEMU for binutils 2.36 (Ubuntu 21.04)


allisonrandal:
**Impact**: software

**Description**
I recently hit the exact problem reported in this patch on the qemu-stable list, as I was compiling QEMU as part of Chipyard (at commit 5129bc7da5093f1d289c2fed27673cf51024eb0d) on Ubuntu 21.04:

https://lists.nongnu.org/archive/html/qemu-stable/2020-12/msg00015.html

> Recent binutils changes dropping unsupported options [1] caused a build
> issue in regard to the optionroms.
> 
>   ld -m elf_i386 -T /<<PKGBUILDDIR>>/pc-bios/optionrom//flat.lds -no-pie \
>     -s -o multiboot.img multiboot.o
>   ld.bfd: Error: unable to disambiguate: -no-pie (did you mean --no-pie ?)
> 
> This isn't really a regression in ld.bfd, filing the bug upstream
> revealed that this never worked as a ld flag [2] - in fact it seems we
> were by accident setting --nmagic).
> 
> Since it never had the wanted effect this usage of LDFLAGS_NOPIE, should be
> droppable without any effect. This also is the only use-case of
> LDFLAGS_NOPIE
> in .mak, therefore we can also remove it from being added there.
> 
> [1]: https://sourceware.org/git/?p=binutils-gdb.git;a=commit;h=983d925d
> [2]: https://sourceware.org/bugzilla/show_bug.cgi?id=27050#c5

The patch was applied in this commit: https://gitlab.com/qemu-project/qemu/-/commit/bbd2d5a8120771ec59b86a80a1f51884e0a26e53

And partially reverted in this commit: https://gitlab.com/qemu-project/qemu/-/commit/ffd205ef2901bd

Because Chipyard is pinning the `toolchains/qemu` git submodule at v5.0.0, it has neither of these fixes, but if Chipyard updates the git submodule to qemu v6.0.1, it will get both of these fixes.

To test the new version of QEMU, I updated the git submodule to v6.0.1 locally, and re-ran `scripts/build-toolchains.sh`. It gave me the following error and exited without building QEMU:

```
==>  PLEASE REMOVE qemu URL-REWRITING from scripts/build-toolchains.sh. It is no longer needed!
```

So, I removed the url rewriting (put in place to avoid git.qemu.org, which no longer exists) using the attached patch, and the toolchain build completes successfully.

[...]

PR #1120, opened 2022-02-15
Bump qemu to v6.0.1 to resolve compile errors on recent versions of Ubuntu and Arch Linux

allisonrandal:
**Related issue**: #1049 

**Type of change**: bug fix

**Impact**: software change

**Release Notes**

### DIFF
--- a/scripts/build-toolchains.sh
+++ b/scripts/build-toolchains.sh
@@ -171,20 +171,6 @@ if [ -z "$IGNOREQEMU" ] ; then
     echo "=>  Starting qemu build"
     dir="$(pwd)/toolchains/qemu"
     echo "==>   Initializing qemu submodule"
-    #since we don't want to use the global config we init passing rewrite config in to the command
-    git -c url.https://github.com/qemu.insteadOf=https://git.qemu.org/git submodule update --init --recursive "$dir"
-    echo "==>  Applying url-rewriting to avoid git.qemu.org"
-    # and once the clones exist, we recurse through them and set the rewrite
-    # in the local config so that any further commands by the user have the rewrite. uggh. git, why you so ugly?
-    git -C "$dir" config --local url.https://github.com/qemu.insteadOf https://git.qemu.org/git
-    git -C "$dir" submodule foreach --recursive 'git config --local url.https://github.com/qemu.insteadOf https://git.qemu.org/git'
-
-    # check to see whether the rewrite rules are needed any more
-    # If you find git.qemu.org in any .gitmodules file below qemu, you still need them
-    # the /dev/null redirection in the submodule grepping is to quiet non-existance of further .gitmodules
-    ! grep -q 'git\.qemu\.org' "$dir/.gitmodules" && \
-    git -C "$dir" submodule foreach --quiet --recursive '! grep -q "git\.qemu\.org" .gitmodules 2>/dev/null' && \
-    echo "==>  PLEASE REMOVE qemu URL-REWRITING from scripts/build-toolchains.sh. It is no longer needed!" && exit 1
 
     # now actually do the build
     SRCDIR="$(pwd)/toolchains" module_build qemu --prefix="${RISCV}" --target-list=riscv${XLEN}-softmmu --disable-werror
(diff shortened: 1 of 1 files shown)


abejgonzalez:
@mergifyio copy dev

abejgonzalez:
To keep you in the loop, we are looking to do a major release of CY (hopefully) by EOD today. If this doesn't make it, we will definitely take a look at this by the next minor release (we are aiming for right before ASPLOS).

allisonrandal:
The fullTest.py test script fails on Ubuntu 21.10 for reasons unrelated to qemu. The first failure is already captured in https://github.com/firesim/FireMarshal/issues/30 but after I apply that workaround, it fails to build images with errors "You must install 'python' on your build machine" (Which means it expects Python 2.x, which was EOL in 2020.) When I work around that (by installing python2 and creating a symlink to it so 'python' exists), then the tests fail with C compile errors in libfakeroot.c. Ultimately, I'd have to test the new version of qemu on an older version of Ubuntu, to avoid incompatibilities between the FireMarshal code and the more recent versions of dependencies on newer versions of Ubuntu. (At some point, FireMarshal will need to be updated too.)

On the centos 7 build issue, ec2fast installs a pre-compiled RISC-V toolchain, but it doesn't change anything for the qemu build. Still, it would be useful information to see whether you get the same error if you choose riscv-tools as the install type instead of ec2fast.

NathanTP:
1. marshal doesn't support python2 and I've tried to use python3 explicitly everywhere I can (though it's possible I missed one somewhere). On my environment, 'python' is python3. Maybe Linux or buildroot use "python" directly somewhere.
2. I'm not sure what libfakeroot is. Where exactly in the build is that failing? There should be some logs from the build (marshal prints the path to the detailed log every time you run it). It might be a buildroot thing.
3. I am having the configure.ac issue for "./build-toolchains.sh riscv-tools" as well. Looking a bit deeper, in chipyard/scripts there is a build-util.sh that has some bash functions in it for building stuff. module_build() assumes that whatever it's building uses autoconf. I assume the new QEMU either isn't using that or is using it differently. I'm not sure why this only shows up on centos though. You may need to skip the functions and run the correct commands directly in build-toolchains.sh.

NathanTP:
Looking again, it seems that the qemu submodule is not being initialized. This PR removes that code from build-toolchains.sh. I don't know what all the shenanigans in there were doing before, but at a minimum we still need to update the qemu submodule. Chipyard is pretty conservative about cloning submodules (hence the "no-riscv-tools" part of the init submodules script). build-toolchains.sh is responsible for setting that up.

allisonrandal:
Ubuntu dropped the binary named 'python' years ago, and only installs binaries named 'python3' (by default) and 'python2' (if you manually install the python2 packages). I got my info out of the log files, and happy to upload them, but since the marshal errors are unrelated to qemu, do you want me to open that as a separate issue on FireMarshal?

allisonrandal:
If you look at the issue #1049, you'll see that those lines from .build-toolchains.sh were deleted because the script itself demanded it:

```
==>  PLEASE REMOVE qemu URL-REWRITING from scripts/build-toolchains.sh. It is no longer needed!
```
I expect that one or more of those deleted lines needs to be modified to update the submodule without doing URL rewriting.

NathanTP:
Bumping bbl seems to fix it. I'm tracking that here: https://github.com/firesim/FireMarshal/pull/230

We'll need to wait for that to get fully tested and merged, but after that this QEMU bump should be OK by me.

As for the other build issues @allisonrandal, we can take that over to the mailing lists or marshal issues. I don't think they're related to this PR.

allisonrandal:
Thanks, I split out the unrelated FireMarshal build failures into https://github.com/firesim/FireMarshal/issues/232

abejgonzalez:
Now that we are using Conda, QEMU shouldn't need to be compiled from source (see #1163). We packaged up QEMU v5.0.0 but if v6 is needed we can support that through the Conda recipe https://github.com/ucb-bar/qemu-feedstock/blob/main/recipe/meta.yaml. Feel free to open a issue if that is needed and we can look into it.

PR #925, opened 2021-07-21
Dev floorplan

jwright6323:
**Related issue**: 

**Type of change**: new feature

**Impact**: rtl change (minor)

**Release Notes**

- Adds floorplanning support using the barstools aspect-based floorplan DSL

PR #1154, opened 2022-04-07
Aspect-based floorplanning

harrisonliew:
**Related PRs / Issues**:

**Type of change**:
- [ ] Bug fix
- [x] New feature
- [ ] Other enhancement

**Impact**:
- [ ] RTL change
- [ ] Build system change
- [x] Other

**Contributor Checklist**:

- [ ] (If applicable) Did you mark the PR as "Please Backport"?

**Release Notes**

Adds floorplanning support using the barstools aspect-based floorplan DSL (supercedes + closes #925)


harrisonliew:
@jwright6323 FPIR Pass error that needs to be solved. This occurs during `GenerateTopAndHarness`, so the Generator itself has successfully ran:

```
[error] Exception in thread "main" java.lang.ClassCastException: firrtl.annotations.InstanceTarget cannot be cast to firrtl.annotations.ReferenceTarget
[error]         at barstools.floorplan.firrtl.GenerateFloorplanIRPass$$anonfun$1.applyOrElse(Passes.scala:83)
[error]         at barstools.floorplan.firrtl.GenerateFloorplanIRPass$$anonfun$1.applyOrElse(Passes.scala:49)
[error]         at scala.PartialFunction.$anonfun$runWith$1$adapted(PartialFunction.scala
```
I'm just running the default config. Steps to replicate:
```
git clone git@github.com:ucb-bar/chipyard.git -b dev-floorplan-new
cd chipyard
./scripts/init-submodules-no-riscv-tools.sh
cd vlsi
make verilog
```

generators/chipyard/src/main/scala/floorplan/ChipTopFloorplans.scala
```diff
@@ -0,0 +1,29 @@
...
+
+import chipyard.TestHarness
+import chipyard.{ChipTopLazyRawModuleImp, BuildSystem, DigitalTop}
+import freechips.rocketchip.config.{Parameters}
+import barstools.floorplan.chisel.{FloorplanAspect, Floorplan, FloorplanFunction}
+
+object ChipTopFloorplans {
+
+  def default: FloorplanFunction = {
+    case top: ChipTopLazyRawModuleImp =>
```
harrisonliew, line 12: @jwright6323 Is there a way to not have to add the `ChipTopLazyRawModuleImp`? It appears this is the only way get the the `DigitalTop`'s tiles.

jerryz123: Can't we do a 
```
top.outer match { 
  case c: ChipTop => c.lazySystem match { 
    case d: DigitalTop => ...
    
    }
  }
}
```

This is admittedly ugly

jerryz123: Does `.wrapper` exist? Looking at the implementation of `LazyRawModuleImp`:
```
/** Actual description of a [[Module]] which can be instantiated by a call to [[LazyModule.module]].
  *
  * @param wrapper the [[LazyModule]] from which the `.module` call is being made.
  */
class LazyModuleImp(val wrapper: LazyModule) extends MultiIOModule with LazyModuleImpLike {
  /** Instantiate hardware of this `Module`. */
  val (auto, dangles) = instantiate()
}
```

harrisonliew: Yes, `.wrapper` exists, and I could do nested matches on `LazyRawModuleImp` followed by `ChipTop` and then `DigitalTop`, but the problem with that is that other `LazyRawModuleImp`s (many other modules in the design) get matched and would be expected to also have a floorplan, which is not what we want. The `ChipTopFloorplans` object should only apply to `ChipTop`.

jerryz123: Ok. I see based on how the FloorplanFunction works, what you have here is probably the best solution

joey0320:
Closing stale PR

PR #1156, opened 2022-04-07
Clean old repo contents in CI

sagark:
(@agebonzalez) This PR is now modified to try and fix CI for self-hosted runners. The main issue is that since we are running on self-hosted runners, self-hosted runners reuse a git clone across multiple runs (it does clean the top-level repo with `git clean -ffdx`). Normally this isn't an issue (if you have no submodules / don't change sub-submodule URLs). However, in the case that you change a sub-submodule URL (and probably other cases), you need to run a `git clean -ffdx` in all submodules recursively - which the `checkout@v2` action doesn't do. Since this is basically just deleting the clone completely, I opt for just deleting the repo contents which forces the `checkout` action to reclone the repo.

### DIFF
--- a/.github/workflows/chipyard-run-tests.yml
+++ b/.github/workflows/chipyard-run-tests.yml
@@ -31,6 +31,12 @@ jobs:
       image: ucbbar/chipyard-ci-image:554b436
       options: --entrypoint /bin/bash
     steps:
+      - name: Delete old checkout
+        run: |
+            ls -alh .
+            rm -rf ./* || true
+            rm -rf ./.* || true
+            ls -alh .
       - name: Checkout
         uses: actions/checkout@v2
       - name: Check commits of each submodule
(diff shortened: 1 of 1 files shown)


abejgonzalez:
@Mergifyio rebase

abejgonzalez:
This is the submodule issue that is hurting us: https://github.com/actions/checkout/issues/358.

abejgonzalez (approved):
LGTM. Pending CI.

abejgonzalez:
@jerryz123 Can you review this? @sagark Can you review this (with a comment since the GH UI won't allow you to be a review since you opened the PR).

jerryz123 (changes requested):
Can the repeated `Delete old checkout` steps not be coalesced into a single script? 
Also, for safety it might be good to have the script reference the directory defined in `defaults.sh`. `rm -rf ./*` is scary.

jerryz123 (approved):

PR #1160, opened 2022-04-24
remove objectmodule

sequencer:
This is needed for RC bumping after chipsalliance/rocket-chip#2967 get merged.


jerryz123:
See #1216

PR #1163, opened 2022-04-28
Use conda + Update initial setup docs

abejgonzalez:
This PR removes the `*-req.sh` scripts in favor of using Conda (same system that FireSim uses to manage system packages).  I've also changed the documentation to be a bit clearer on how to run things in a AWS EC2 context and in the normal scenario. Notable changes are below:

- Use conda for package management. Set of working packages are listed in `conda-requirements*.yaml` files. Additionally, lock files `*.conda-lock.yml` files are provided to ensure a consistent set of packages.
- New `setup.sh` file that automates the setup steps including setting up a conda environment, submodule cloning, and building extra toolchain collateral (i.e. Spike, PK, etc)
- Bump to Python 3.9 for FireMarshal (tested `buildroot`/`fedora` build with it). Also, use QEMU built by conda in the `ucb-bar` channel.
- Now using toolchain packages uploaded to `ucb-bar` conda channel (`riscv-tools` and `esp-tools` are both provided)
- Updated initial setup documentation. Point to FireSim/Chipyard docs in specific places. See https://chipyard.readthedocs.io/en/conda/Chipyard-Basics/Initial-Repo-Setup.html#initial-repository-setup
- Updated CI to use `self-hosted` runners for test runs. Needed since GH-A cache action now doesn't support caching between Docker containers and self-hosted runners.

**Related PRs / Issues**:

**Type of change**:
- [ ] Bug fix
- [x] New feature
- [x] Other enhancement

**Impact**:
- [ ] RTL change
- [x] Build system change
- [x] Other

**Contributor Checklist**:


docs/Chipyard-Basics/Initial-Repo-Setup.rst
```diff
@@ -10,19 +10,52 @@ Chipyard is developed and tested on Linux-based systems.
...
 .. Warning:: Working under Windows is not recommended.
 
+Running on AWS EC2 with FireSim
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-In CentOS-based platforms, we recommend installing the following dependencies:
+If you plan on using Chipyard alongside FireSim, you should refer to the `FireSim documentation `<https://docs.fires.im/>`__
+on "Initial Setup/Simulation".
+When you reach Section 2.3.2 on "Setting up the FireSim Repo"
+you can continue to :ref:`Setting up the Chipyard Repo` to continue to setup the repository with Chipyard as the top-level repository.
```
sagark, line 19: Can you directly link to the correct starting section of the FireSim docs?

abejgonzalez: That's the plan. I just need to figure out to point to either latest or stable depending on what the version of the CY docs is being built.

docs/Chipyard-Basics/Initial-Repo-Setup.rst
```diff
@@ -10,19 +10,52 @@ Chipyard is developed and tested on Linux-based systems.
...
+In Chipyard, we use `Conda <https://docs.conda.io/en/latest/>`__ to help manage system dependencies (``make``, ``gcc``, etc).
+We **highly** recommend running the following script to setup Chipyard's Conda environment.
 
-.. include:: /../scripts/ubuntu-req.sh
-   :code: bash
+.. code-block:: shell
+
+    ./scripts/machine-launch-script.sh
+
+.. Warning:: If you have ``sudo`` priveleges then this script should work fine.
```
jerryz123, line 31:
```suggestion
.. Warning:: If you have ``sudo`` privileges then this script should work fine.
```

docs/Chipyard-Basics/Initial-Repo-Setup.rst
```diff
@@ -10,19 +10,52 @@ Chipyard is developed and tested on Linux-based systems.
...
+    should run ``./scripts/machine-launch-script.sh --help`` to view relevant configuration parameters.
+
+.. Note:: When running on an Amazon Web Services EC2 FPGA-development instance
+    (for FireSim), FireSim includes a similar machine setup script that will install all
+    of the aforementioned dependencies (and some additional ones).
+
+Depending on where you ran the machine launch script,
+you should find a file called ``machine-launch-status.log`` and ``machine-launch-status`` indicating
+if things completed properly.
+
```
jerryz123, line 42: Why does this log it to a status file, instead of just showing it on the terminal.

docs/Chipyard-Basics/Initial-Repo-Setup.rst
```diff
@@ -10,19 +10,52 @@ Chipyard is developed and tested on Linux-based systems.
...
+.. code-block:: shell
 
-.. Note:: When running on an Amazon Web Services EC2 FPGA-development instance (for FireSim), FireSim includes a machine setup script that will install all of the aforementioned dependencies (and some additional ones).
+    cat ./machine-launch-status # should have "started" and "completed"
+
+Once run, please confirm that you "activated" the proper Conda environment and the prior script completed.
```
jerryz123, line 47: Should there be instructions on how to activate a Conda env?

docs/Chipyard-Basics/Initial-Repo-Setup.rst
```diff
@@ -10,19 +10,52 @@ Chipyard is developed and tested on Linux-based systems.
...
+By running the following command you should see "chipyard" output (the default environment is called "chipyard").
+
+.. code-block:: shell
+
+    echo $CONDA_DEFAULT_ENV
+
+.. Note:: If you don't have a Conda environment set (and the machine launch script completed),
+    you can run ``conda activate chipyard`` to "enter" the proper environment.
```
jerryz123, line 55: Does the launch script auto activate the env? 
I'm not sure I like the idea of scripts which hijack the environment of the user.

docs/Chipyard-Basics/Initial-Repo-Setup.rst
```diff
@@ -10,19 +10,52 @@ Chipyard is developed and tested on Linux-based systems.
...
 
-.. include:: /../scripts/centos-req.sh
-   :code: bash
+Default Requirements Installation
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-In Ubuntu/Debian-based platforms (Ubuntu), we recommend installing the following dependencies.
-These dependencies were written based on Ubuntu 16.04 LTS and 18.04 LTS - If they don't work for you, you can try out the Docker image (:ref:`Chipyard-Basics/Initial-Repo-Setup:Pre-built Docker Image`) before manually installing or removing dependencies:
+In Chipyard, we use `Conda <https://docs.conda.io/en/latest/>`__ to help manage system dependencies (``make``, ``gcc``, etc).
+We **highly** recommend running the following script to setup Chipyard's Conda environment.
```
jerryz123, line 25: Can you add a description here of what the conda env is pulling?

scripts/machine-launch-script.sh
```diff
@@ -0,0 +1,305 @@
+#!/bin/bash
+
```
jerryz123, line 2: Why don't we just force the user to install Conda before running this script, instead of installing it for them?

scripts/machine-launch-script.sh
```diff
@@ -0,0 +1,305 @@
...
+    echo "Usage: $0 [options]"
+    echo
+    echo "Options:"
+    echo "[--help]                  List this help"
+    echo "[--prefix <prefix>]       Install prefix for conda. Defaults to /opt/conda."
+    echo "                          If <prefix>/bin/conda already exists, it will be used and install is skipped."
+    echo "[--env <name>]            Name of environment to create for conda. Defaults to '$CONDA_ENV_NAME'."
+    echo "[--dry-run]               Pass-through to all conda commands and only print other commands."
+    echo "                          NOTE: --dry-run will still install conda to --prefix"
+    echo "[--reinstall-conda]       Repairs a broken base environment by reinstalling."
```
jerryz123, line 24: When is this `--reinstall-conda` thing useful?

scripts/machine-launch-script.sh
```diff
@@ -0,0 +1,305 @@
...
+    echo "Examples:"
+    echo "  % $0"
+    echo "     Install into default system-wide prefix (using sudo if needed) and add install to system-wide /etc/profile.d"
+    echo "  % $0 --prefix ~/conda --env my_custom_env"
+    echo "     Install into $HOME/conda and add install to ~/.bashrc"
+    echo "  % $0 --prefix \${CONDA_EXE%/bin/conda} --env my_custom_env"
+    echo "     Create my_custom_env in existing conda install"
+    echo "     NOTES:"
+    echo "       * CONDA_EXE is set in your environment when you activate a conda env"
+    echo "       * my_custom_env will not be activated by default at login see /etc/profile.d/conda.sh & ~/.bashrc"
```
jerryz123, line 36: I don't like that this script is needs sudo to do all these system-level changes. Basically the types of users are:
- Users running on a disposable F1 instance
- Users running on a managed cluster
- Users running on a personal computer

Only for the F1 use case is changing system-level settings acceptable.

jerryz123 (changes requested):
This seems mostly fine.
I would reduce the amount of stuff the machine-launch-script.sh has to do. If we assume Conda is installed already, than the script does not need SUDO, and it can be renamed to `conda-env-setup.sh` or something (which is much more descriptive than "machine launch").

We should generally avoid giant multi-purpose scripts with a bunch of flags that leave effects on the user's environment. This is really only acceptable for the F1 case, where the instances are "disposable".

scripts/machine-launch-script.sh
```diff
@@ -0,0 +1,305 @@
...
+    # platform-specific setup
+    case "$OS_FLAVOR" in
+        ubuntu)
+            ;;
+        centos)
+            ;;
+        *)
+            echo "::ERROR:: Unknown OS flavor '$OS_FLAVOR'. Unable to do platform-specific setup."
+            exit 1
+            ;;
```
jerryz123, line 123: What's the point of this? To say we only support ubunctu and centos?
Should the comment say `# detect supported platforms` instead?

scripts/machine-launch-script.sh
```diff
@@ -0,0 +1,305 @@
...
+    # other misc deps
+    CONDA_PACKAGE_SPECS+=(
+        sbt \
+        ca-certificates \
+        mosh \
+        gmp \
+        mpfr \
+        mpc \
+        zlib \
+        vim \
```
jerryz123, line 265:
```suggestion
        emacs \
        vim \
```

scripts/machine-launch-script.sh
```diff
@@ -0,0 +1,305 @@
...
+        dtc \
+        verilator==4.034 \
+    )
+
+    if [[ "$CONDA_ENV_NAME" == "base" ]]; then
+        # NOTE: arg parsing disallows installing to base but this logic is correct if we ever change
+        CONDA_SUBCOMMAND=install
+        CONDA_ENV_BIN="${CONDA_INSTALL_PREFIX}/bin"
+    else
+        CONDA_ENV_BIN="${CONDA_INSTALL_PREFIX}/envs/${CONDA_ENV_NAME}/bin"
```
jerryz123, line 285: Is it standard practice to put the envs in `/opt/conda`? This would require sudo? Why don't we just install it to `chipyard/conda-env` (for example)

docs/Chipyard-Basics/Initial-Repo-Setup.rst
abejgonzalez: It does both. The docs can be clarified

scripts/machine-launch-script.sh
abejgonzalez: I think we can have it default to a Chipyard path, but we want to give the user the option to override it to put it in a global area.

abejgonzalez:
I've sketched out where the conda activate calls should go. Once that is settled I'll update the docs to reflect that.

scripts/init-submodules-no-riscv-tools-nolog.sh
```diff
@@ -74,6 +74,8 @@ CHIPYARD_DIR="$(dirname "$DIR")"
 
 cd "$CHIPYARD_DIR"
 
+source $CHIPYARD_DIR/scripts/conda-activate-env.sh
+
```
jerryz123, line 78: Why does this need to be here? To get the most recent git version?

I would prefer that this script still works without `conda`. Just tell the user in docs to activate the conda env before running anything else in Chipyard.

Also I would prefer that this doesn't activate conda in the env.sh script.

So the documented flow for users should be 
1. Install Conda
2. conda-env-setup.sh
3. conda-activate-env.sh
4. Do all other chipyard stuff (including init submodules).

And power users or millennium users can skip 1-3.

scripts/conda-activate-env.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+if conda env list | grep -q "^firesim"; then
+    conda activate firesim
+else
+    conda activate chipyard
+fi
```
jerryz123, line 7: When would there be a `firesim` env? Why not just tell the user to always `conda activate chiypard`?

jerryz123 (changes requested):
Looks close, but I think with some changes we can support the no-conda use case too.

scripts/conda-activate-env.sh
abejgonzalez: FireSim by default calls it's env `firesim`. So if we want to support Chipyard in the FireSim setup (on AWS F1) then we should switch between them.

scripts/init-submodules-no-riscv-tools-nolog.sh
abejgonzalez: I get the feeling that if we don't explicitly call conda activate in our scripts, avg. users will forget about this and will get weird errors that should be solved with conda activated. For now, I will remove these explicit calls, but I'll add a warning if the environment isn't set up like the check for the hash (aka you need to answer the prompt to continue).

jerryz123: What tends to break in `init-submodules` without the conda-managed env? If the issue is the git version is too old, the script can check that and error with a message like.
```
ERROR: Git version {version} <= {supported version}. Did you create and activate the chipyard conda environment?
```

The prompt in build-toolchains is fine.

scripts/conda-activate-env.sh
jerryz123: The docs say
>  Warning:: When running on an Amazon Web Services EC2 FPGA-development instance (for FireSim), FireSim includes a similar machine setup script that will install all of the aforementioned dependencies (and some additional ones). Skip this section.

This instructs users running Chipyard in FireSim to skip running this script, since I believe the fsim docs should contain instructions for activating the firesim env. In that case, this script doesn't need to exist, and the docs can just say `conda activate chipyard`

For my own understanding, I believe the flow is:
1. If FireSim, follow firesim docs, then skip to 5
2. install conda
3. create chipyard conda env
4. activate conda env
5. init submodules

abejgonzalez:
Cancelling CI since there needs to be more work done to make that work.

scripts/machine-launch-script.sh
timsnyder-siv: If for example, you have removed python from your base environment or otherwise broken conda.  It's a passthrough to the conda installer.

scripts/machine-launch-script.sh
timsnyder-siv: It only uses sudo and changes the system-level settings if you tell it to install at a path that can't be written by the current user or you run the script as root.  That should be clarified.

scripts/init-submodules-no-riscv-tools-nolog.sh
timsnyder-siv: I agree with Jerry.  You should decouple setting up the environment from running the flow.  Ask users to supply you with the output of `env | sorted` whenever they submit an issue that should be enough to tell you whether they are on a mill machine and whether they have conda activated or not (look for CONDA_PREFIX or CONDA_DEFAULT_ENV).

scripts/install-conda.sh
```diff
@@ -0,0 +1,178 @@
+#!/bin/bash
+
+CONDA_INSTALL_PREFIX=/opt/conda
+CONDA_INSTALLER_VERSION=4.12.0-0
+CONDA_INSTALLER="https://github.com/conda-forge/miniforge/releases/download/${CONDA_INSTALLER_VERSION}/Miniforge3-${CONDA_INSTALLER_VERSION}-Linux-x86_64.sh"
+CONDA_CMD="conda" # some installers install mamba or micromamba
+CONDA_ENV_NAME="firesim"
```
abejgonzalez, line 7: TODO: Remove since this doesn't matter

scripts/install-conda.sh
```diff
@@ -0,0 +1,178 @@
...
+    # Use the fast solver by default
+    "${DRY_RUN_ECHO[@]}" $SUDO "$CONDA_EXE" config --system --set experimental_solver libmamba
+
+    conda_init_extra_args=()
+    if [[ "$INSTALL_TYPE" == system ]]; then
+        # if we're installing into a root-owned directory using sudo, or we're already root
+        # initialize conda in the system-wide rcfiles
+        conda_init_extra_args=(--no-user --system)
+    fi
+    $SUDO "${CONDA_EXE}" init $DRY_RUN_OPTION "${conda_init_extra_args[@]}" bash
```
abejgonzalez, line 172: TODO: Is this the right command?

.github/scripts/defaults.sh
```diff
@@ -5,32 +5,20 @@ CI_MAKE_NPROC=8
...
-HOME=$GITHUB_WORKSPACE
-
 # remote variables
 # CI_DIR is defined externally based on the GH repository secret BUILDDIR
 
 REMOTE_PREFIX=$CI_DIR/${GITHUB_REPOSITORY#*/}-${GITHUB_REF_NAME//\//-}
 REMOTE_WORK_DIR=$GITHUB_WORKSPACE
-REMOTE_RISCV_DIR=$GITHUB_WORKSPACE/riscv-tools-install
-REMOTE_ESP_DIR=$GITHUB_WORKSPACE/esp-tools-install
 REMOTE_CHIPYARD_DIR=$GITHUB_WORKSPACE
```
abejgonzalez, line 13:
```suggestion
REMOTE_CHIPYARD_DIR=$GITHUB_WORKSPACE
REMOTE_SIM_DIR=$REMOTE_CHIPYARD_DIR/sims/verilator
```

.gitmodules
```diff
@@ -28,40 +28,6 @@
...
-[submodule "toolchains/esp-tools/riscv-pk"]
-	path = toolchains/esp-tools/riscv-pk
-	url = https://github.com/riscv/riscv-pk.git
-[submodule "toolchains/esp-tools/riscv-tests"]
-	path = toolchains/esp-tools/riscv-tests
-	url = https://github.com/ucb-bar/esp-tests.git
-[submodule "toolchains/libgloss"]
-	path = toolchains/libgloss
-	url = https://github.com/ucb-bar/libgloss-htif.git
 [submodule "vlsi/hammer"]
```
jerryz123, line 31: Can we leave these submodules+build_toolchains.sh in? iI thought we are keeping the submodules, just removing them from the "default" instructions.

docs/Chipyard-Basics/Initial-Repo-Setup.rst
```diff
@@ -6,28 +6,41 @@ Requirements
...
+
+After Conda is installed and is on your ``PATH``, we need to install a version of ``git`` to initially checkout the repository.
+For this you can use the system package manager like ``yum`` or ``apt`` to install ``git``.
+This ``git`` is only used to first checkout the repository, we will later install a newer version of ``git`` with Conda.
 
 Setting up the Chipyard Repo
 -------------------------------------------
 
-Start by fetching Chipyard's sources. Run:
+Start by checkout out the proper Chipyard's version. Run:
```
jerryz123, line 43:
```suggestion
Start by checking out out the proper Chipyard's version. Run:
```

docs/Chipyard-Basics/Initial-Repo-Setup.rst
```diff
@@ -36,40 +49,90 @@ Start by fetching Chipyard's sources. Run:
...
-    ./scripts/build-toolchains.sh riscv-tools # for a normal risc-v toolchain
+    conda install -c ucb-bar ucb-bar-riscv-tools # for a normal risc-v toolchain
+
+.. Note:: If you are planning to use the Hwacha vector unit, or other RoCC-based accelerators, you should obtain the esp-tools toolchain by adding the ``ucb-bar-esp-tools`` argument to the command above.
+
+Once the command is run, the ``PATH``, ``RISCV``, and ``LD_LIBRARY_PATH`` environment variables will be set properly.
 
-.. Note:: If you are planning to use the Hwacha vector unit, or other RoCC-based accelerators, you should build the esp-tools toolchain by adding the ``esp-tools`` argument to the script above.
-  If you are running on an Amazon Web Services EC2 instance, intending to use FireSim, you can also use the ``--ec2fast`` flag for an expedited installation of a pre-compiled toolchain.
+.. Note:: If you are a power user and would like to build your own toolchain, you can refer to the https://github.com/ucb-bar/riscv-tools-feedstock and https://github.com/ucb-bar/esp-tools-feedstock repositories on how to build a toolchain yourself.
```
jerryz123, line 115: Didn't we flatten riscv-tools and esp-tools into chipyard in the first place to make it easier to bump those projects? This adds back that level of indirection.

In fact, doesn't this add two levels of indirection? Now we additionally have to track commits for riscv-tools-feedstock and riscv-tools.

scripts/install-conda.sh
```diff
@@ -0,0 +1,178 @@
+#!/bin/bash
+
+CONDA_INSTALL_PREFIX=/opt/conda
+CONDA_INSTALLER_VERSION=4.12.0-0
```
jerryz123, line 4: Can we just point to the official docs for installing conda instead?

jerryz123 (commented):
Did we discuss removing all the toolchain submodules? I'm wary of moving back to the riscv-tools-library-of-submodules approach.

.gitmodules
abejgonzalez: I'll re-add the modules once they work (or some version of them). Right now, I was thinking of two approaches:

- Pointing to the conda feedstock repos (aka the conda build recipes). Those repos have the conda requirements listed out, a pointer to the src repos, and a script to build a conda package.
- Point to the src repos aka the new submodules that hold what was `toolchains/esp-tools` and `toolchains/riscv-tools`. Each of those src repos has its own `build-toolchains.sh` script.

dockerfiles/Dockerfile
```diff
@@ -2,59 +2,53 @@
 
 # BUILD BASE FOR CI
 
-FROM ubuntu:18.04 as base
+FROM ubuntu:20.04 as base
```
abejgonzalez, line 5: Right now you can't use a Debian OS for Chipyard since installing the toolchains will break due to dependency conflicts in conda.

scripts/install-conda.sh
abejgonzalez: I need this script for CI. In the docs, I point to the official documentation.

timsnyder-siv: Perhaps this script should be in `.github/scripts` if it is intended only for CI?

dockerfiles/Dockerfile
timsnyder-siv: I'm curious to know more about this.

scripts/install-conda.sh
abejgonzalez: I think some users will like using this, but sure I can move it to the `.github/scripts` area.

dockerfiles/Dockerfile
abejgonzalez: I'm using the following Dockerfile:

```
FROM continuumio/miniconda3
SHELL ["/bin/bash", "-c"]
WORKDIR /root
RUN git clone https://github.com/ucb-bar/chipyard.git && \
    cd chipyard && \
    git checkout conda && \
    conda env create -f scripts/conda-requirements.yaml && \
    conda install -n chipyard -c ucb-bar ucb-bar-riscv-tools
```

I see the following error:

```
UnsatisfiableError: The following specifications were found to be incompatible with each other:

Output in format: Requested package -> Available versions

Package libzlib conflicts for:
ucb-bar-riscv-tools -> libcurl[version='>=7.83.1,<8.0a0'] -> libzlib[version='>=1.2.11,<1.3.0a0']
ucb-bar-riscv-tools -> libzlib[version='>=1.2.12,<1.3.0a0']

Package openssl conflicts for:
python=3.8 -> openssl[version='>=1.1.1d,<1.1.2a|>=1.1.1e,<1.1.2a|>=1.1.1g,<1.1.2a|>=1.1.1j,<1.1.2a|>=1.1.1k,<1.1.2a|>=1.1.1l,<1.1.2a|>=1.1.1n,<1.1.2a']
ucb-bar-riscv-tools -> libcurl[version='>=7.83.1,<8.0a0'] -> openssl[version='>=1.1.1o,<1.1.2a|>=1.1.1q,<1.1.2a']The following specifications were found to be incompatible with your system:

  - feature:/linux-64::__glibc==2.31=0
  - feature:|@/linux-64::__glibc==2.31=0
  - python=3.8 -> libgcc-ng[version='>=7.5.0'] -> __glibc[version='>=2.17']

Your installed version is: 2.31
```

.gitmodules
abejgonzalez: Re-added

docs/Chipyard-Basics/Initial-Repo-Setup.rst
abejgonzalez: Yes. But for the most part, the main item that is being rebuilt is Spike (for Gemmini folks). That can be rebuilt separately. In the future we will focus on splitting up things even more.

scripts/install-conda.sh
abejgonzalez: Moved.

scripts/conda-activate-env.sh
abejgonzalez: This was re-worded in the past. Better now?

dockerfiles/Dockerfile
abejgonzalez: I'll address this later since our main OS's (`centos` + `ubuntu`) are handled.

docs/Chipyard-Basics/Initial-Repo-Setup.rst
```diff
@@ -36,40 +49,90 @@ Start by fetching Chipyard's sources. Run:
...
-This will take about 20-30 minutes. You can expedite the process by setting a ``make`` environment variable to use parallel cores: ``export MAKEFLAGS=-j8``.
+Currently there are two toolchains, one for normal RISC-V programs called ``riscv-tools``, and another for Hwacha/Gemmini called ``esp-tools``.
+To get a basic ``riscv-tools`` installation (which is the only thing needed for most Chipyard use-cases), just the following steps are necessary.
 
 .. code-block:: shell
 
-    ./scripts/build-toolchains.sh riscv-tools # for a normal risc-v toolchain
+    conda install -c ucb-bar riscv-tools # for a normal risc-v toolchain
+
+.. Note:: If you are planning to use the Hwacha vector unit, or other RoCC-based accelerators, you should obtain the ``esp-tools`` toolchain by adding the ``esp-tools`` argument to the command above.
```
jerryz123, line 111: Shouldn't this be "substituting" and not "adding" the argument?

docs/Chipyard-Basics/Initial-Repo-Setup.rst
```diff
@@ -36,40 +49,90 @@ Start by fetching Chipyard's sources. Run:
...
-  If you are running on an Amazon Web Services EC2 instance, intending to use FireSim, you can also use the ``--ec2fast`` flag for an expedited installation of a pre-compiled toolchain.
+.. Note:: If you are a power user and would like to build your own toolchain, you can refer to the https://github.com/ucb-bar/riscv-tools-feedstock and https://github.com/ucb-bar/esp-tools-feedstock repositories (submoduled in the ``toolchains`` directory) on how to build a toolchain yourself.
 
-Once the script is run, a ``env.sh`` file is emitted that sets the ``PATH``, ``RISCV``, and ``LD_LIBRARY_PATH`` environment variables.
-You can put this in your ``.bashrc`` or equivalent environment setup file to get the proper variables, or directly include it in your current environment:
+.. Note:: If can deactivate/activate a toolchain (but keep it installed) by running ``source $CONDA_PREFIX/etc/conda/deactivate.d/deactivate-${PKG_NAME}.sh`` or ``$CONDA_PREFIX/etc/conda/activate.d/activate-${PKG_NAME}.sh`` (``PKG_NAME`` for example is ``ucb-bar-riscv-tools``). This will modify the aforementioned 3 environment variables.
```
jerryz123, line 117:
```suggestion
.. Note:: You can deactivate/activate a toolchain (but keep it installed) by running ``source $CONDA_PREFIX/etc/conda/deactivate.d/deactivate-${PKG_NAME}.sh`` or ``$CONDA_PREFIX/etc/conda/activate.d/activate-${PKG_NAME}.sh`` (``PKG_NAME`` for example is ``ucb-bar-riscv-tools``). This will modify the aforementioned 3 environment variables.
```

docs/Chipyard-Basics/Initial-Repo-Setup.rst
```diff
@@ -36,40 +49,90 @@ Start by fetching Chipyard's sources. Run:
...
+.. Note:: Refer to FireSim's :fsim_doc:`Conda documentation <Advanced-Usage/Conda.html>` on more information
+    on how to use Conda and some of its benefits.
+
+Next go ahead and activate the conda environment that was setup.
+
+.. code-block:: shell
+
+    conda activate chipyard
+
+We recommend that you add this "activate" command your ``.bashrc`` (or other environment setup file).
```
jerryz123, line 84:
```suggestion
We recommend that you add this "activate" command to your ``.bashrc`` (or other environment setup file).
```

scripts/conda-requirements.yaml
```diff
@@ -0,0 +1,57 @@
...
+  - binutils
+
+  # rocket-chip deps
+  - sbt
+  - openjdk
+  - dtc
+  - verilator==4.034
+
+  # chipyard deps
+  - dromajo # from ucb-bar
```
jerryz123, line 47: What is publishing dromajo?

docs/Chipyard-Basics/Initial-Repo-Setup.rst
```diff
@@ -6,28 +6,41 @@ Requirements
...
 
-.. include:: /../scripts/centos-req.sh
-   :code: bash
+Default Requirements Installation
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-In Ubuntu/Debian-based platforms (Ubuntu), we recommend installing the following dependencies.
-These dependencies were written based on Ubuntu 16.04 LTS and 18.04 LTS - If they don't work for you, you can try out the Docker image (:ref:`Chipyard-Basics/Initial-Repo-Setup:Pre-built Docker Image`) before manually installing or removing dependencies:
+In Chipyard, we use the `Conda <https://docs.conda.io/en/latest/>`__ package manager to help manage system dependencies.
+Conda allows users to create an "environment" that holds system dependencies like ``make``, ``gcc``, etc.
```
jerryz123, line 27:
```suggestion

.. Note:: Chipyard can also run on systems without a Conda installation. However, users on these systems must manually install toolchains and dependencies.

```

dockerfiles/Dockerfile
timsnyder-siv: I think that the conda install you pick up from `continuumio/miniconda3` includes the 'defaults' channel and that is contributing tho the solvability issue.  I'd recommend that you use `condaforge/miniforge3` as your base docker image if you want to do something like that.

docs/Chipyard-Basics/Initial-Repo-Setup.rst
```diff
@@ -6,28 +6,41 @@ Requirements
...
 
-.. include:: /../scripts/ubuntu-req.sh
-   :code: bash
+First Chipyard requires there to be Conda installed on the system.
+Please refer to the `Conda installation instructions <https://docs.conda.io/projects/conda/en/latest/user-guide/install/index.html>`__ on how to install Conda.
```
timsnyder-siv, line 29: If people install conda from one of the Anaconda Inc installers (miniconda or the full Anaconda), you will be configured to install packages from the 'defaults' channels built into conda (that come from Anaconda Inc).

'defaults' and 'conda-forge' are known to conflict in some gnarly ways and that's why `conda-forge` folks reccomend settings documented at https://conda-forge.org/docs/user/introduction.html#how-can-i-install-packages-from-conda-forge 

For more information on the types of conflicts that happen, see https://conda-forge.org/docs/user/tipsandtricks.html#multiple-channels

Alternately, you can avoid the whole issue by installing from one of the miniforge installers https://github.com/conda-forge/miniforge/#download which don't use anything built by Anaconda Inc and don't have the 'defaults' channel available.

Since you're using `conda env create -f scripts/conda-requirements.yml`, I _think_ the `channels` list in the YAML overrides any channels configured.  I'll go double check that I'm remembering that correctly.

scripts/conda-requirements.yaml
```diff
@@ -0,0 +1,57 @@
+name: chipyard
+channels:
+  - conda-forge
+  - ucb-bar
```
timsnyder-siv, line 4:
```suggestion
  - ucb-bar
  - nodefaults
```

I think to avoid any confusion about where you've installed conda from (Anaconda Inc vs miniforge installers), we should explicitly add the nodefaults meta-channel that prevents usage of the 'defaults' channel as documented at https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#creating-an-environment-file-manually

dockerfiles/Dockerfile
abejgonzalez: I'll give this a go

scripts/conda-requirements.yaml
abejgonzalez: This repo: https://github.com/riscv-boom/dromajo. I can re-add this but I think so little people use Dromajo that I can just add a comment pointing to the feedstock (same thing for the QEMU version)

docs/Chipyard-Basics/Initial-Repo-Setup.rst
abejgonzalez: Interesting, I'll try to update the docs and point to the miniforge installers since that's probably the best thing to do.

scripts/conda-requirements.yaml
jerryz123: Cool. I didn't catch that you added the conda scripts there too

timsnyder-siv: @abejgonzalez I think you're missing the `dromajo_cosim.h` header in your `dromajo` package.  At least the FireSim driver compile fails because it isn't in there.

vighneshiyer:
What would the flow look like if I wanted to use a custom version of riscv-isa-sim? If I were to clone it somewhere and compile it, how can I make sure it would override the riscv-isa-sim files included in the riscv-tools conda package when running RTL simulations (i.e. using the custom fesvr.so / spike)?

I think it would be better to have separate conda packages for riscv-gcc, libgloss, riscv-isa-sim, riscv-pk, and riscv-isa-tests. That way it is easy to uninstall one of them from your conda environment and replace it with a custom version. It also makes the dependency tree clearer (e.g. riscv-isa-sim doesn't depend on any of the other packages). In the interim, it is probably ok to leave everything but riscv-gcc as a source dependency in chipyard, considering they are fast to compile.

abejgonzalez:
> What would the flow look like if I wanted to use a custom version of riscv-isa-sim? If I were to clone it somewhere and compile it, how can I make sure it would override the riscv-isa-sim files included in the riscv-tools conda package when running RTL simulations (i.e. using the custom fesvr.so / spike)?

You would still build your version of Spike and put it on the `PATH` (I'm not sure if you need to modify the `LD_LIBRARY_PATH` for custom Spikes). If you really wanted to be absolutely sure there is no pollution of the normal tools with the custom Spike you could deactivate the normal tools by running `source $CONDA_PREFIX/etc/conda/deactivate.d/deactivate...riscv-tools.sh`.

> I think it would be better to have separate conda packages for riscv-gcc, libgloss, riscv-isa-sim, riscv-pk, and riscv-isa-tests. That way it is easy to uninstall one of them from your conda environment and replace it with a custom version. It also makes the dependency tree clearer (e.g. riscv-isa-sim doesn't depend on any of the other packages). In the interim, it is probably ok to leave everything but riscv-gcc as a source dependency in chipyard, considering they are fast to compile.

That is definitely what we plan on getting towards, this PR is an intermediate PR where we can 1st have everything bundled together. I think the final split of items will be Spike + PK in one package, and everything else in a separate package (i.e. everything that depends on the toolchain) until we split up the toolchain even more according to how conda normally packages toolchains. If people really care about having Spike + PK built separately, I can add that in this PR.

scripts/conda-requirements.yaml
abejgonzalez: It should be added now.

abejgonzalez:
Ok right now I have the following setup:

1. Install the RISC-V compiler with the conda package `conda install -c ucb-bar {riscv-tools/esp-tools}` (sets $RISCV/$PATH/$LD_LIB_PATH
2. Run `./scripts/build-toolchains-extra.sh {riscv-tools/esp-tools}` to build/install Spike/PK/tests in the $RISCV area made/set by the conda package (libgloss also installed to compiler area).

I wonder if this is better/would be better for people.

1. Install the RISC-V compiler with the conda package outside of $RISCV env (in normal $CONDA_PREFIX/bin) (also point $LD_LIBRARY_PATH to $CONDA_PREFIX/lib)
2. Install Spike/PK/tests to a local RISCV folder (like the current toolchain folder - `$CY/riscv-tools-install`). Env. variables would point to this area like before. libgloss would still have to be installed directly to where the compiler is located.

This setup would minimize the items needed to be built and installed in the conda managed area and would make it clearer what is managed by CY. I'm just not sure how much things would break with the compiler not present in the $RISCV path.

abejgonzalez:
@sagark @a0u @jerryz123 This is now ready for round 2 of review (CI is failing due to elfutils not getting found)

docs/Chipyard-Basics/Initial-Repo-Setup.rst
```diff
@@ -36,40 +53,101 @@ Start by fetching Chipyard's sources. Run:
...
+.. Warning:: When running on an Amazon Web Services EC2 FPGA-development instance
+    (for FireSim), FireSim includes a similar machine setup script that will install all
+    of the aforementioned dependencies (and some additional ones) and will activate the
+    proper conda environment. Skip this section.
+
+Next run the following command to create Chipyard's Conda environment.
+
+.. code-block:: shell
+
+    conda env create -f scripts/conda-requirements.yml
```
a0u, line 71:
```suggestion
    conda env create -f scripts/conda-requirements.yaml -n chipyard
```

The file extension should be `.yaml`.
In my case, the `-n` option turned out to be mandatory.

.github/CI_README.md
```diff
@@ -22,7 +22,7 @@ For example:
 This specifies that the `prepare-chipyard-cores` job needs the both the `make-keys` and the `setup-complete` steps to
 be completed before it can run.
 
-Chipyard runs its CI using a docker image created from `dockerfiles/Dockerfile`.
+Chipyard runs its CI using a docker image created from `dockerfiles/Dockerfile` and on Berkeley's millennium machines.
```
jerryz123, line 25:
```suggestion
Chipyard runs its CI using a docker image created from `dockerfiles/Dockerfile` on Berkeley's compute resources.
```

jerryz123 (approved):

abejgonzalez:
Unfortunately, something in Linux boot is breaking :/ causing a kernel panic. I'll have to look into if this is a QEMU issue, or more likely an issue with the buildroot build with the current dependencies

image

a0u:
> Unfortunately, something in Linux boot is breaking :/ causing a kernel panic. I'll have to look into if this is a QEMU issue, or more likely an issue with the buildroot build with the current dependencies

Check that the dynamic linker `ld.so` does not have `rpath` set.  Perhaps the Conda environment is interfering with the cross-compilation of riscv-glibc?

abejgonzalez:
@mergifyio rebase

abejgonzalez:
> Unfortunately, something in Linux boot is breaking :/ causing a kernel panic. I'll have to look into if this is a QEMU issue, or more likely an issue with the buildroot build with the current dependencies
> 
> image

Ok. I believe I resolved the issue. I tested the most recent bumped versions of the compilers (+ sysroot) with buildroot and fedora booting.

jerryz123:
This should be good to merge. I removed the CircleCI webhook from this repo settings, since we've moved away from that service.

abejgonzalez:
Waiting on a review from @sagark and @a0u before merging in.

a0u (approved):

.readthedocs.yml
```diff
@@ -3,13 +3,12 @@ version: 2
...
 formats: all
 
 sphinx:
     configuration: docs/conf.py
 
-python:
-    install:
-        - requirements: docs/requirements.txt
+conda:
+    environment: scripts/conda-requirements.yaml
```
abejgonzalez, line 14:
```suggestion
    environment: conda-requirements-riscv-tools.yaml
```

.github/CI_README.md
```diff
@@ -101,21 +99,9 @@ To get the CI to work correctly you need to create the following GH Repository S
...
-The SERVER is constructed like this:
-```bash
-SERVER = ${{ secrets.BUILDUSER }}@${{ secrets.BUILDSERVER }}
-```
-
-Additionally, you need to add under the "PERMISSIONS" "SSH Permissions" section a private key that is on the build server that you are using.
-After adding a private key, it will show a fingerprint that should be added under the jobs that need to be run.
-
-Note: On the remote server you need to have the `*.pub` key file added to the `authorized_keys` file.
+Additionally, you need to install conda on the build servers that exist
```
sagark, line 104:
```suggestion
Additionally, you need to install conda on the build servers that exist.
```

docs/Chipyard-Basics/Initial-Repo-Setup.rst
```diff
@@ -6,28 +6,53 @@ Requirements
...
 .. Warning:: Working under Windows is not recommended.
 
+Running on AWS EC2 with FireSim
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-In CentOS-based platforms, we recommend installing the following dependencies:
+If you plan on using Chipyard alongside FireSim, you should refer to the :fsim_doc:`FireSim documentation <>`.
+Specifically, you should follow the :fsim_doc:`Initial Setup/Simulation <Initial-Setup/index.html>`
+section of the docs up until :fsim_doc:`Setting up the FireSim Repo <Initial-Setup/Setting-up-your-Manager-Instance.html#setting-up-the-firesim-repo>`.
+At that point, instead of checking out FireSim you can checkout Chipyard by following :ref:`Chipyard-Basics/Initial-Repo-Setup:Setting up the Chipyard Repo`.
```
sagark, line 20:
```suggestion
At that point, instead of cloning FireSim you can clone Chipyard by following :ref:`Chipyard-Basics/Initial-Repo-Setup:Setting up the Chipyard Repo`.
```

docs/Chipyard-Basics/Initial-Repo-Setup.rst
```diff
@@ -6,28 +6,53 @@ Requirements
...
-These dependencies were written based on Ubuntu 16.04 LTS and 18.04 LTS - If they don't work for you, you can try out the Docker image (:ref:`Chipyard-Basics/Initial-Repo-Setup:Pre-built Docker Image`) before manually installing or removing dependencies:
+In Chipyard, we use the `Conda <https://docs.conda.io/en/latest/>`__ package manager to help manage system dependencies.
+Conda allows users to create an "environment" that holds system dependencies like ``make``, ``gcc``, etc.
 
-.. include:: /../scripts/ubuntu-req.sh
-   :code: bash
+.. Note:: Chipyard can also run on systems without a Conda installation. However, users on these systems must manually install toolchains and dependencies.
 
-.. Note:: When running on an Amazon Web Services EC2 FPGA-development instance (for FireSim), FireSim includes a machine setup script that will install all of the aforementioned dependencies (and some additional ones).
+First Chipyard requires there to be Conda installed on the system.
```
sagark, line 30:
```suggestion
First, Chipyard requires Conda to be installed on the system.
```

abejgonzalez:
This PR is good to merge. The errors for SPI flash/read are because we need to bump Verilator to 4.226 and the CVA6 error is because there is a segfault when CVA6 is used with DRAMSim2 in the new verilator (this is solved if you use magic memory)

jerryz123:
CVA6 with VCS + DRAMSIM is still fine?

abejgonzalez:
> CVA6 with VCS + DRAMSIM is still fine?

I haven't had the chance to verify this on this PR but pre-Conda I do remember this working.

docs/Chipyard-Basics/Initial-Repo-Setup.rst
```diff
@@ -1,33 +1,58 @@
...
+.. code-block:: shell
+
+    conda install -n base conda-lock
+    conda activate base
 
 Setting up the Chipyard Repo
 -------------------------------------------
 
-Start by fetching Chipyard's sources. Run:
+Start by checking out the proper Chipyard's version. Run:
```
sagark, line 55:
```suggestion
Start by checking out the proper Chipyard version. Run:
```

docs/Simulation/FPGA-Accelerated-Simulation.rst
```diff
@@ -6,30 +6,28 @@ FPGA-Accelerated Simulation
...
 FireSim allows RTL-level simulation at orders-of-magnitude faster speeds than software RTL simulators.
 FireSim also provides additional device models to allow full-system simulation, including memory models and network models.
 
-FireSim currently supports running only on Amazon EC2 F1 FPGA-enabled virtual instances.
-In order to simulate your Chipyard design using FireSim, if you have not
-already, follow the initial EC2 setup instructions as detailed in the `FireSim
-documentation  <http://docs.fires.im/en/latest/Initial-Setup/index.html>`__.
-Then clone Chipyard onto your FireSim manager
-instance, and setup your Chipyard repository as you would normally.
+FireSim supports running on Amazon EC2 F1 FPGA-enabled virtual instances and on locally managed Linux machines with FPGAs attached.
```
sagark, line 13:
```suggestion
FireSim supports running on Amazon EC2 F1 FPGA-enabled cloud instances and on locally managed Linux machines with FPGAs attached.
```

sagark (approved):
LGTM modulo a few fixes in comments.

abejgonzalez:
> 



> > CVA6 with VCS + DRAMSIM is still fine?
> 
> I haven't had the chance to verify this on this PR but pre-Conda I do remember this working. I can confirm that VCS + CVA6 + DRAMSim2 works

I've gone ahead and fixed this by removing a `getenv` call in DRAMSim2 (which I don't think is used often)(the `getenv` is causing a segfault presumably due to multi-thread race condition).

PR #1178, opened 2022-06-17
Script to get a clean git status after running standard initalization procedures

sagark:
See title. This is a quick fix for now, all of these problems really need to be fixed in the appropriate submodules.


a0u (approved):
Approving as a temporary hack

PR #1223, opened 2022-09-22
Misc. updates to bash scripts

### LINKED ISSUE #1221: Explicitly type `Yy` or `Nn` for validate step
### Background Work

- [X] Yes, I searched the [mailing list](https://groups.google.com/forum/#!forum/chipyard)
- [X] Yes, I searched [prior issues](https://github.com/ucb-bar/chipyard/issues)
- [X] Yes, I searched the [documentation](https://chipyard.readthedocs.io/)

### Chipyard Version and Hash

Release: `main`
Hash: bd5ca64

### OS Setup

N/A

### Other Setup

N/A

### Current Behavior

If the user types misc. things into the keyboard before hitting the "validate" part in any script it will exit prematurely since the check just checks for `Yy`. 

https://github.com/ucb-bar/chipyard/blob/bd5ca643b8d99bc0c964bf8840a16f2c36f7af82/scripts/init-submodules-no-riscv-tools-nolog.sh#L56

### Expected Behavior

Explictly check for `Nn` or reword the prompt to have an explicit `Yy/Nn` prompt.

### Other Information

_No response_

abejgonzalez:
* Rename `--skip-validate` to something more standard. I propose using `--batch`. This change requires changes in FireSim (will be done once we determine what name to choose)
* Add generating ctags to the `build-setup.sh` file.
* Add `utils.sh` bash library file that is shared across scripts (common code can be put here and reused).
* Try to preserve bash options more (move things into subshells, added new `<save/restore>_bash_options` functions)
* Changes validate check to check for `Yy` or `Nn` explicitly.

**Related PRs / Issues**:
Fixes #1221 

**Type of change**:
- [ ] Bug fix
- [ ] New feature
- [x] Other enhancement

**Impact**:
- [ ] RTL change
- [x] Build system change
- [ ] Other

**Contributor Checklist**:


build-setup.sh
```diff
@@ -50,8 +50,8 @@ do
             TOOLCHAIN=$1 ;;
         -ud | --unpinned-deps )
             USE_PINNED_DEPS=false ;;
-        --skip-validate)
-            SKIP_VALIDATE_FLAG=$1 ;;
+        --batch)
```
jerryz123, line 53: You should maintain existing behavior for `--skip-validate` to avoid breaking existing flows (like CI, for example).

scripts/utils.sh
```diff
@@ -0,0 +1,55 @@
+#/usr/bin/env bash
+
+#######################################
+# Common setup. Init MacOS compatibility
+# variables.
+# Globals:
+#   READLINK
+#######################################
+function common_setup
```
jerryz123, line 9: Why does this function need to be called explicitly to set up `$READLINK`? Can't the body just be executed when the script is sourced?

scripts/init-submodules-no-riscv-tools-nolog.sh
```diff
@@ -4,75 +4,71 @@
...
+            echo "ERROR: bad argument $1"
             usage
             exit 2
             ;;
     esac
     shift
 done
 
-# Check that git version is at least 1.7.8
-MYGIT=$(git --version)
```
jerryz123, line 36: Are you removing this because Conda always installs the correct git? Even so, I think it's a good idea to leave this around.

jerryz123 (changes requested):
I'm not sure I understand the usage of `--batch` in this context. `--force`/`-f` or `--assumeyes`/`-y` might be a better fit for this use case.

build-setup.sh
abejgonzalez: I'll have to update the other flows but I can say that the skip validate is deprecated and remove it later (just to avoid bumping FireSim initially)

scripts/utils.sh
abejgonzalez: True. I just decided that everything in this file should be a function. I think that makes things cleaner

scripts/init-submodules-no-riscv-tools-nolog.sh
abejgonzalez: Sure. I removed it because of conda but I can keep this for now

abejgonzalez:
> I'm not sure I understand the usage of `--batch` in this context. `--force`/`-f` or `--assumeyes`/`-y` might be a better fit for this use case.

I have no real preference. I was thinking `-f` made sense but I switched to `--batch` since Tim recommended it at some point. I'll probably just switch to "force" unless there is a better arg name.

build-setup.sh
jerryz123: What I mean is, leave in support for `--skip-validate`, and remove all references to this from the documentation.

abejgonzalez:
Ping @jz

jerryz123 (approved):

PR #1239, opened 2022-10-04
CIRCT Integration

abejgonzalez:
- Adds CIRCT `firtool` integration (changes barstools `GenerateTopAndHarness` executable to preprocess Firrtl if need be). Dramatically speeds up FIRRTL compile times!!!
- Emits design verilog as one module per verilog file

TODO:
- [x] Add/build `firtool` conda package
- [x] Create `.f` manifest files for top/harness level modules (do people care about the black-box `.f` files)
- [x] Cleanly add SiFive module annotation to `.anno.json`
- [x] ~Add a clean way to bypass the CIRCT compiler completely if wanted (which changes SFC to dump the same output as the SFC + CIRCT flow)~ - This will be probably more effort than its worth so unless this is absolutely necessary I'm ignoring this
- [x] Change `HARNESS_*` to `MODEL_*` for consistency
- [x] Change some `FIRTOOL_*` stuff to `SFC_FIRTOOL_*` (or a better name?)
- [x] Check output SyncReadMem verilog output (FIRTOOL output acts like SRAMS with async writes, requiring the user to inject a cycle manually before writes to replicate the SFC generated SRAM behavior)

**Related PRs / Issues**:

**Type of change**:
- [ ] Bug fix
- [x] New feature
- [ ] Other enhancement

**Impact**:
- [ ] RTL change
- [x] Build system change
- [ ] Other

**Contributor Checklist**:


abejgonzalez:
Note this only works up until `make verilog`

common.mk
```diff
@@ -114,46 +114,49 @@ generator_temp: $(SCALA_SOURCES) $(sim_files) $(SCALA_BUILDTOOL_DEPS) $(EXTRA_GE
...
-TOP_TARGETS = $(TOP_FILE) $(TOP_SMEMS_CONF) $(TOP_ANNO) $(TOP_FIR) $(sim_top_blackboxes)
-HARNESS_TARGETS = $(HARNESS_FILE) $(HARNESS_SMEMS_CONF) $(HARNESS_ANNO) $(HARNESS_FIR) $(sim_harness_blackboxes)
+CIRCT_TARGETS = $(VSRC_SMEMS_FILE) $(VSRC_MODH_JSON)
 
 # DOC include start: FirrtlCompiler
 # NOTE: These *_temp intermediate targets will get removed in favor of make 4.3 grouped targets (&: operator)
 .INTERMEDIATE: firrtl_temp
-$(TOP_TARGETS) $(HARNESS_TARGETS): firrtl_temp
+$(CIRCT_TARGETS): firrtl_temp
 	@echo "" > /dev/null
```
jerryz123, line 139: Is the plan to remove the Top/Harness differentiation? I'm fine with this, the VLSI tools shouldn't need to be passed a verily file with only TOP anyways.

common.mk
```diff
@@ -114,46 +114,49 @@ generator_temp: $(SCALA_SOURCES) $(sim_files) $(SCALA_BUILDTOOL_DEPS) $(EXTRA_GE
...
-		--harness-anno-out $(HARNESS_ANNO) \
-		--harness-dotf-out $(sim_harness_blackboxes) \
-		--harness-fir $(HARNESS_FIR) \
-		$(REPL_SEQ_MEM) \
-		$(HARNESS_CONF_FLAGS) \
-		--target-dir $(build_dir) \
-		--log-level $(FIRRTL_LOGLEVEL) \
-		$(EXTRA_FIRRTL_OPTIONS))
-	touch $(sim_top_blackboxes) $(sim_harness_blackboxes)
+	$(SCRATCH_HOME)/circt/build/bin/firtool \
```
jerryz123, line 142: What is this?

abejgonzalez: `firtool` is the LLVM MLIR-based CIRCT compiler. See https://github.com/llvm/circt  and https://github.com/chipsalliance/chisel3/blob/f4050157dc59cffff52eccfcd8613c9de058ae5d/src/main/scala/circt/stage/phases/CIRCT.scala#L121

common.mk
abejgonzalez: I plan on having two `.f`s that say what modules reside underneath Top and Harness (where the harness `.f` will include modules that reside in the Top as well).

abejgonzalez: @harrisonliew Any thoughts on this? Also does Hammer need the `.f` black-box manifest files?

common.mk
```diff
@@ -114,69 +114,68 @@ generator_temp: $(SCALA_SOURCES) $(sim_files) $(SCALA_BUILDTOOL_DEPS) $(EXTRA_GE
...
```
abejgonzalez, line 142: How does this work with Chisel aspects? Can you still inject FIRRTL passes through annotations?

common.mk
harrisonliew: Hammer can't use the `.f` files directly; the content of that file would need to be appended into the list of input files. That's trivial if the `.f` files are kept.

common.mk
harrisonliew: Related to your comment below - is firtool essentially some Phase that is pre-compiled into a binary (with Ninja, it seems), which is the main reason why it's faster than what we have now?

common.mk
harrisonliew: If my understanding above is correct, it seems like if you have your own custom FIRRTL passes, you'd need to run them separately, and if they work, make a Phase that extends CIRCT and compile that into your own custom firtool binary?

harrisonliew: Ah, some documentation: https://github.com/llvm/circt/blob/eba1bb5b81578e83a6457bf2cf32b94dda00b813/docs/Dialects/FIRRTL/RationaleFIRRTL.md and https://github.com/llvm/circt/blob/16f18f40b003023b8abb452ee85e4c5255107201/docs/Dialects/FIRRTL/FIRRTLAnnotations.md
From the first link:
> The FIRRTL dialect in CIRCT is designed to provide a drop-in replacement for the SFC for the subset of FIRRTL IR that is produced by Chisel and in common use.
It explicitly lacks `Fixed` and `Interval` datatype support (ouch, for DSP ppl - this might be a dealbreaker, such as for @milovanovic?), and from the second link seems to statically support the list of annotations in the second link. So yeah, looks like custom passes/annotations need to be run separately.

common.mk
abejgonzalez: It is basically a complete rewrite of the FIRRTL compiler passes into a LLVM binary (C++). So it is a full replacement of the scala FIRRTL compiler.

common.mk
abejgonzalez: That's roughly my understanding. You need to re-write any Scala FIRRTL passes into C++ and compile them into your binary. I do wonder if there is any way to inject passes (which was an old feature that the Scala FIRRTL compiler supported).

harrisonliew: Yeah, in addition to figuring out if passes can be injected, it would also be informative to know whether/if there are any migration rules for passes from SFC to CIRCT.

harrisonliew: Also, if CIRCT's subset is too small, can we keep targeting SFC as a fallback instead of replacing it entirely?

tymcauley: I was also curious about the topic of adding custom passes to CIRCT, and asked about it here: https://discourse.llvm.org/t/writing-custom-firrtl-passes/62216

The answer basically boiled down to: it's not currently supported, but that's only because of lack of work on the feature. Ideally custom passes could be upstreamed into CIRCT, but for those that can't, some work needs to be put in to develop a system for user-defined passes. Right now, the only way to make those happen is by forking CIRCT and compiling the `firtool` binary yourself after writing your custom pass.

abejgonzalez: I think my working goal right now is to have the following flow: Chisel > SFC (output .for) > CIRCT (output .sv). Here the SFC would do a couple of different things depending on if there are custom passes/unsupported constructs in the .fir. It would basically be a preprocessing step which would run extra passes (if found)(this would potentially lower  CHIRRTL and take a long time), and/or resolve unsupported FIRRTL constructs (again probably lowering the .fir)

harrisonliew: Hold on--is the main time-consuming step in SFC is lowering to any form of FIRRTL, or the actual transformation actions while the circuit is in FIRRTL, or the lowering of FIRRTL to Verilog? Seems like definitely not the last one, so wouldn't your proposed flow barely save time?

Separately, if unsupported constructs are limited to the mentioned data types, you should be able to do this pre-processing in Chisel?

abejgonzalez: My assumption is that in most cases the SFC will just be a no-op, passing the FIRRTL directly to CIRCT. However, if there are special types (i.e. `Interval`, etc) or if there are custom passes that should be run then SFC will do all of the lowering (time-consuming) (and CIRCT will just handle the low FIRRTL -> Verilog). This flow hopefully will have everyone use CIRCT + SFC instead of a make flag that enables SFC v.s. CIRCT.

I'm not sure how you do Chisel pre-processing for converting a `Interval`/`Fixed` type to a "normal" `UInt` (or other supported type). Can you clarify what you were thinking here?

harrisonliew: > This flow hopefully will have everyone use CIRCT + SFC instead of a make flag that enables SFC v.s. CIRCT.

Got it, sounds good.

Sorry, I meant just detection of the unsupported constructs, not conversion to a supported datatype. But your proposal of running a lightweight SFC pass that determines whether it passes high FIRRTL to CIRCT or not would do the job.

seldridge: Hey. Feel free to ping me directly if you all have questions about this. I don't always see these discussions.

> How does this work with Chisel aspects?

Aspects that run at Chisel-time and generate annotations still work as those are entirely in Chisel. There's no plan to support `InjectingAspects` (annotations that append blobs of FIRRTL into a module) as that is viewed as a pretty janky API.

Re: annotations. That `.md` may be slightly out of date. The guaranteed up-to-date list of accepted annotations in in this header file: https://github.com/llvm/circt/blob/main/include/circt/Dialect/FIRRTL/AnnotationDetails.h Annotations are currently allow-listed. `firtool` will error on an annotation it doesn't understand (unless you run with `--disable-annotation-unknown` as this PR is doing).

> It explicitly lacks Fixed and Interval datatype support...

We're on the fence about `Fixed`. It's not that much work to add it, we just don't use it. I'm not totally against an open source contribution to add it to CIRCT, but start a discussion before you go this route. Jack has also thought that `Fixed` may be entirely expressible in Chisel without relying on FIRRTL. This would be ideal. There are other discussions about user-defined types in FIRRTL which would be another way that this could show up.

`Intervals` is another story as that was always very experimental, but did get usage at Berkeley. Value range analysis is bread and butter for compilers and there is already some work in MLIR for doing this with the `arith` dialect (which is used for ML/software). I expect this will show up in CIRCT eventually, just not in FIRRTL.

> Hold on--is the main time-consuming step in SFC is lowering to any form of FIRRTL, or the actual transformation actions while the circuit is in FIRRTL, or the lowering of FIRRTL to Verilog? Seems like definitely not the last one, so wouldn't your proposed flow barely save time?

That's the main drawback of mixing a Scala-based FIRRTL Compiler (SFC) flow with an MLIR-based FIRRTL Compiler (MFC/CIRCT) flow for the purposes of running custom transforms. In my experience, almost all non-trivial custom transforms are written against Low FIRRTL because it's extremely difficult to work with higher level FIRRTL in the SFC. You then wind up running almost the whole compilation through SFC for minimal gain (other than it being an incremental step towards moving to MFC).

One new development that is happening is MFC will support Chisel's [Trace API](https://www.chisel-lang.org/api/latest/chisel3/experimental/Trace$.html). This will enable you to get the final path of anything you annotate. This is expected to be sufficient enough to replace many annotations/custom passes that are used to emit information from a design, e.g., TCL generation. There were also a large number of SiFive custom passes that made there way into CIRCT (without accompanying Chisel APIs). There may be overlap with existing custom SFC passes. (I see you've turned on `sifive.enterprise.firrtl.ModuleHierarchyAnnotation`. 😉 )

common.mk
abejgonzalez: Ok, now there is a `*.all.f` and `*.top.f` file which has all modules or just top modules within it. I think the VLSI makefile needs to be updated, @harrisonliew do you see anything that you might need?

common.mk
abejgonzalez: Thanks for the clarification! I agree that using SFC for custom transforms will most likely remove all the benefits of the CIRCT compiler. However, I expect that most people don't need custom transforms so this will still be a benefit.

common.mk
harrisonliew: Looks good. Yes, the VLSI Makefile does need to be updated, pending this request:

There are use cases in which you want to run some submodule through the VLSI flow (with the `VLSI_TOP` Make variable), but then simulate that module within the entire system. Right now this requires netlist hacking to substitute the Verilog that came out of the VLSI flow. Could you make your `dump-mods.py` script dump the following:
- `*.all.f`: all modules (this has Harness too, right? Not that useful in VLSI flow, but might as well have it anyway)
- `*.top.f`: ChipTop as the top module by default, but if some additional option is given to the script (e.g. `--vlsi-top`), then the top module is the one in `--vlsi-top`
- `*.???.f`: not sure what to name this file, but it should contain the remainder of the hierarchy above ChipTop or the module given by `--vlsi-top` option. This would be given to sim tools, since the netlist coming out of syn/par won't have any of these modules.

abejgonzalez: Ok for clarity you want the following:

`*.top.f` - Everything underneath `--dut-top`
`*.tb.f` - Everything underneath the topmost level (normally the testharness) that doesn't include the modules under `*.top.f`

Do you need CY to change the recipe (i.e. pass in `VLSI_TOP` in a switch) or do you just plan on re-running the `dump-mods` script with `VLSI_TOP`?

harrisonliew: If "underneath" is inclusive, then yes, that is correct.

It would be easiest if `VLSI_TOP` was passed in with a switch or something. No point duplicating the recipe.

abejgonzalez:
Notes (so I don't forget):

The script to create `.f` files expects that the module name == file name. This breaks when handling the black-box files (which don't have this necessarily). Right now, I just have a separate BB `.f` file that needs to be added in addition to the other `*.f` files. For more clarity this is what happens:

- `*.top.f` - Includes all Chisel module files under top (and maybe some BB files - if the mod == filename)
- `*.tb.f` - Same as `top` version but contains modules under TB only (not under the DUT)
- `*.all.f` - Just a concat of the top and TB `.f`s (so may not include all BBs)
- `firrtl_bb_rsrc_files.final.f` - List of BB files

common.mk
abejgonzalez: Can you remind me why you can't just override the `TOP` variable when running these simulations?

harrisonliew: Hm, interesting point. If we don't require specification of `TOP` for the Top/Harness separation anymore, then a separate `VLSI_TOP` may not be necessary? After all a simulation tool doesn't care about what's harness/dut, it just needs all the files.

scripts/split-mems-conf.py
```diff
@@ -0,0 +1,65 @@
...
+    parser = argparse.ArgumentParser(description='Use CIRCT (firtool) smems JSONs to create DUT and test harness smems confs')
+    parser.add_argument('--in-smems-conf', type=str, required=True, help='')
+    parser.add_argument('--in-dut-smems-json', type=str, required=True, help='')
+    parser.add_argument('--in-tb-smems-json', type=str, required=True, help='')
+    parser.add_argument('--out-dut-smems-conf', type=str, required=True, help='')
+    parser.add_argument('--out-tb-smems-conf', type=str, required=True, help='')
+    args = parser.parse_args()
+
+    with open(args.in_smems_conf) as isc:
+        with open(args.in_dut_smems_json) as idsj:
```
jerryz123, line 36: Try doing the multline thing when opening multiple files at once

https://legacy.python.org/dev/peps/pep-0008/#maximum-line-length

abejgonzalez: Done

common.mk
abejgonzalez: I'll leave the API such that what you want to synth. is called `TOP`.

abejgonzalez:
@harrisonliew could you modify the VLSI makefile and try this out with a Hammer flow? Also, could you link me a custom xform that you might want to run in the future?

.github/scripts/defaults.sh
```diff
@@ -26,7 +26,7 @@ LOCAL_FIRESIM_DIR=$LOCAL_CHIPYARD_DIR/sims/firesim/sim
 # key value store to get the build groups
 declare -A grouping
 grouping["group-cores"]="chipyard-cva6 chipyard-ibex chipyard-rocket chipyard-hetero chipyard-boom chipyard-sodor chipyard-digitaltop chipyard-multiclock-rocket chipyard-nomem-scratchpad"
-grouping["group-peripherals"]="chipyard-dmirocket chipyard-blkdev chipyard-spiflashread chipyard-spiflashwrite chipyard-mmios chipyard-lbwif"
```
jerryz123, line 29: What's breaking the spi tests?

common.mk
```diff
@@ -111,42 +111,100 @@ $(FIRRTL_FILE) $(ANNO_FILE) &: $(SCALA_SOURCES) $(sim_files) $(SCALA_BUILDTOOL_D
...
 		$(EXTRA_FIRRTL_OPTIONS))
-	touch $(sim_top_blackboxes) $(sim_harness_blackboxes)
+	$(if $(shell grep "Fixed<" $(FIRRTL_FILE)),mv $(SFC_FIRRTL_BASENAME).mid.fir $(SFC_FIRRTL_FILE),)
+	firtool \
+		--format=fir \
+		-O=release \
+		--dedup \
+		--export-module-hierarchy \
+		--emit-metadata \
+		--verify-each=false \
```
seldridge, line 171: We don't run with this unless we're drag racing. Or: the verifier shouldn't be slow. Notably, a slow verifier was sped up by a lot in https://github.com/llvm/circt/pull/4077 which mostly removes the need for turning off verification.

(For background, with `--verify-each=false`, verification runs once after parsing and then never again. With `--verify-each=true`, verification runs after every pass. An SFC run is closer to `--verify-each=false` as it runs all the check passes once and then never again. There is usual no risk in running with `--verify-each=false` except that you may see a compiler bug leading to a crash _later_ than you would after `--verify-each=true`.)

common.mk
```diff
@@ -111,42 +111,100 @@ $(FIRRTL_FILE) $(ANNO_FILE) &: $(SCALA_SOURCES) $(sim_files) $(SCALA_BUILDTOOL_D
...
+		--disable-annotation-classless \
+		--disable-annotation-unknown \
+		--warn-on-unprocessed-annotations \
+		--lowering-options=disallowPackedArrays,emittedLineLength=8192,noAlwaysComb,disallowLocalVariables \
```
seldridge, line 175: FYI: we are currently running with: `--lowering-options=emittedLineLength=2048,verifLabels,disallowLocalVariables,explicitBitcast,locationInfoStyle=wrapInAtSquareBracket`. If you do any running with Verilator, the last part will emit source locators as `// @[foo.bar line:col]` instead of `// foo.bar line:col`. This works around a rare bug where Verilator will interpret any comment that starts with "verilator" as a directive it is supposed to process, regardless of case or spacing. This only comes up if you have any files in your code base that start with "verilator".

common.mk
```diff
@@ -111,42 +111,100 @@ $(FIRRTL_FILE) $(ANNO_FILE) &: $(SCALA_SOURCES) $(sim_files) $(SCALA_BUILDTOOL_D
...
+		--repl-seq-mem \
+		--repl-seq-mem-circuit=$(MODEL) \
+		--repl-seq-mem-file=$(FIRTOOL_SMEMS_CONF) \
+		--annotation-file=$(SFC_ANNO_FILE) \
+		--split-verilog \
+		-o $(OUT_DIR) \
+		$(SFC_FIRRTL_FILE)
```
seldridge, line 182: Both `--verbose-pass-executions` and `--mlir-timing` are useful for showing progress and final compile time.

common.mk
```diff
@@ -111,42 +111,100 @@ $(FIRRTL_FILE) $(ANNO_FILE) &: $(SCALA_SOURCES) $(sim_files) $(SCALA_BUILDTOOL_D
...
+	$(if $(shell grep "Fixed<" $(FIRRTL_FILE)),mv $(SFC_FIRRTL_BASENAME).mid.fir $(SFC_FIRRTL_FILE),)
+	firtool \
+		--format=fir \
+		-O=release \
+		--dedup \
+		--export-module-hierarchy \
+		--emit-metadata \
+		--verify-each=false \
+		--disable-annotation-classless \
+		--disable-annotation-unknown \
```
seldridge, line 173: Having either of these options on will likely result in miscompiles as some annotations are definitely getting ignored. Watch out.

seldridge (commented):
Super cool. Thanks for getting this together!

common.mk
```diff
@@ -111,42 +111,100 @@ $(FIRRTL_FILE) $(ANNO_FILE) &: $(SCALA_SOURCES) $(sim_files) $(SCALA_BUILDTOOL_D
...
+		--warn-on-unprocessed-annotations \
+		--lowering-options=disallowPackedArrays,emittedLineLength=8192,noAlwaysComb,disallowLocalVariables \
+		--repl-seq-mem \
+		--repl-seq-mem-circuit=$(MODEL) \
+		--repl-seq-mem-file=$(FIRTOOL_SMEMS_CONF) \
+		--annotation-file=$(SFC_ANNO_FILE) \
+		--split-verilog \
+		-o $(OUT_DIR) \
+		$(SFC_FIRRTL_FILE)
+	sed -i 's/.*/& /' $(FIRTOOL_SMEMS_CONF) # need trailing space for SFC macrocompiler
```
tymcauley, line 183: So, the `sed -i` strategy gave me a few hiccups on macOS. First, I ran this and got some errors since macOS `sed` doesn't have the same `-i` flag as GNU `sed`. Homebrew provides the `coreutils` package, which gives you `gsed`, so I just replaced all the `sed` calls in here to use that.

Then I re-ran the build, and got an error related to the SFC not being able to parse the memory config statements. That was because this `sed -i` command failed, but this rule was never re-run, since `FIRTOOL_SMEMS_CONF` was already generated. So the `sed -i` strategy doesn't feel like it meshes so cleanly with a make-based flow. Or maybe this is such a narrow error case that we should be fine with it?

.github/scripts/defaults.sh
seldridge: What is the SFC doing here and what is more of the input FIRRTL? Is this a usage of an analog that is only attached port-to-instance or is it being routed up/down?

The answer in the past has been to make CIRCT work harder to sink `inout` into instances. However, we are usually very careful to never route analog other than port-to-instance.

abejgonzalez: Here is the FIRRTL:

```
circuit TestHarness :
  module TestHarness :
    input clock : Clock
    input reset : UInt<1>
    output io : { success : UInt<1>}

    wire buildtopClock : Clock @[TestHarness.scala 86:27]
    wire buildtopReset : Reset @[TestHarness.scala 87:27]
    inst chiptop of ChipTop @[TestHarness.scala 90:19]
    io.success <= UInt<1>("h0") @[TestHarness.scala 92:14]
    node dutReset = asAsyncReset(buildtopReset) @[TestHarness.scala 94:32]
    wire dtm_success : UInt<1>
    dtm_success <= UInt<1>("h0")
    when dtm_success : @[HarnessBinders.scala 251:28]
      io.success <= UInt<1>("h1") @[HarnessBinders.scala 251:41]
    wire jtag_wire : { TCK : Clock, TMS : UInt<1>, TDI : UInt<1>, flip TDO : { data : UInt<1>, driven : UInt<1>}} @[HarnessBinders.scala 252:29]
    jtag_wire.TDO.data <= chiptop.jtag.TDO @[HarnessBinders.scala 253:28]
    jtag_wire.TDO.driven <= UInt<1>("h1") @[HarnessBinders.scala 254:30]
    chiptop.jtag.TCK <= jtag_wire.TCK @[HarnessBinders.scala 255:15]
    chiptop.jtag.TMS <= jtag_wire.TMS @[HarnessBinders.scala 256:15]
    chiptop.jtag.TDI <= jtag_wire.TDI @[HarnessBinders.scala 257:15]
    inst SimJTAG of SimJTAG @[HarnessBinders.scala 258:26]
    SimJTAG.exit is invalid
    SimJTAG.init_done is invalid
    SimJTAG.enable is invalid
    SimJTAG.jtag is invalid
    SimJTAG.reset is invalid
    SimJTAG.clock is invalid
    node _T = asUInt(buildtopReset) @[HarnessBinders.scala 258:107]
    node _T_1 = asUInt(buildtopReset) @[HarnessBinders.scala 258:134]
    node _T_2 = not(_T_1) @[HarnessBinders.scala 258:115]
    jtag_wire.TCK <= SimJTAG.jtag.TCK @[Periphery.scala 217:15]
    jtag_wire.TMS <= SimJTAG.jtag.TMS @[Periphery.scala 218:15]
    jtag_wire.TDI <= SimJTAG.jtag.TDI @[Periphery.scala 219:15]
    SimJTAG.jtag.TDO.driven <= jtag_wire.TDO.driven @[Periphery.scala 220:17]
    SimJTAG.jtag.TDO.data <= jtag_wire.TDO.data @[Periphery.scala 220:17]
    SimJTAG.clock <= buildtopClock @[Periphery.scala 222:14]
    SimJTAG.reset <= _T @[Periphery.scala 223:14]
    inst plusarg_reader of plusarg_reader_134 @[PlusArg.scala 80:11]
    plusarg_reader.out is invalid
    SimJTAG.enable <= plusarg_reader.out @[Periphery.scala 225:18]
    SimJTAG.init_done <= _T_2 @[Periphery.scala 226:18]
    node _dtm_success_T = eq(SimJTAG.exit, UInt<1>("h1")) @[Periphery.scala 230:26]
    dtm_success <= _dtm_success_T @[Periphery.scala 230:15]
    node _T_3 = geq(SimJTAG.exit, UInt<2>("h2")) @[Periphery.scala 231:19]
    when _T_3 : @[Periphery.scala 231:27]
      node _T_4 = dshr(SimJTAG.exit, UInt<1>("h1")) @[Periphery.scala 232:59]
      node _T_5 = bits(reset, 0, 0) @[Periphery.scala 232:13]
      node _T_6 = eq(_T_5, UInt<1>("h0")) @[Periphery.scala 232:13]
      when _T_6 : @[Periphery.scala 232:13]
        printf(clock, UInt<1>("h1"), "*** FAILED *** (exit code = %d)\n", _T_4) : printf @[Periphery.scala 232:13]
      node _T_7 = bits(reset, 0, 0) @[Periphery.scala 233:11]
      node _T_8 = eq(_T_7, UInt<1>("h0")) @[Periphery.scala 233:11]
      when _T_8 : @[Periphery.scala 233:11]
        stop(clock, UInt<1>("h1"), 1) : stop @[Periphery.scala 233:11]
    node _bits_T = asClock(UInt<1>("h0")) @[SerialAdapter.scala 26:32]
    inst bits_out_queue of AsyncQueue @[SerialAdapter.scala 27:29]
    bits_out_queue.clock <= _bits_T
    bits_out_queue.reset <= UInt<1>("h0")
    bits_out_queue.io.enq <= chiptop.serial_tl.bits.out @[SerialAdapter.scala 28:24]
    bits_out_queue.io.enq_clock <= chiptop.serial_tl.clock @[SerialAdapter.scala 29:30]
    node _bits_out_queue_io_enq_reset_T = asUInt(buildtopReset) @[SerialAdapter.scala 30:39]
    bits_out_queue.io.enq_reset <= _bits_out_queue_io_enq_reset_T @[SerialAdapter.scala 30:30]
    bits_out_queue.io.deq_clock <= buildtopClock @[SerialAdapter.scala 31:30]
    node _bits_out_queue_io_deq_reset_T = asUInt(buildtopReset) @[SerialAdapter.scala 32:39]
    bits_out_queue.io.deq_reset <= _bits_out_queue_io_deq_reset_T @[SerialAdapter.scala 32:30]
    inst bits_in_queue of AsyncQueue_1 @[SerialAdapter.scala 33:28]
    bits_in_queue.clock <= _bits_T
    bits_in_queue.reset <= UInt<1>("h0")
    chiptop.serial_tl.bits.in <= bits_in_queue.io.deq @[SerialAdapter.scala 34:20]
    bits_in_queue.io.deq_clock <= chiptop.serial_tl.clock @[SerialAdapter.scala 35:29]
    node _bits_in_queue_io_deq_reset_T = asUInt(buildtopReset) @[SerialAdapter.scala 36:38]
    bits_in_queue.io.deq_reset <= _bits_in_queue_io_deq_reset_T @[SerialAdapter.scala 36:29]
    bits_in_queue.io.enq_clock <= buildtopClock @[SerialAdapter.scala 37:29]
    node _bits_in_queue_io_enq_reset_T = asUInt(buildtopReset) @[SerialAdapter.scala 38:38]
    bits_in_queue.io.enq_reset <= _bits_in_queue_io_enq_reset_T @[SerialAdapter.scala 38:29]
    wire bits : { flip in : { flip ready : UInt<1>, valid : UInt<1>, bits : UInt<32>}, out : { flip ready : UInt<1>, valid : UInt<1>, bits : UInt<32>}} @[SerialAdapter.scala 39:25]
    bits_in_queue.io.enq <= bits.in @[SerialAdapter.scala 40:23]
    bits.out.bits <= bits_out_queue.io.deq.bits @[SerialAdapter.scala 41:19]
    bits.out.valid <= bits_out_queue.io.deq.valid @[SerialAdapter.scala 41:19]
    bits_out_queue.io.deq.ready <= bits.out.ready @[SerialAdapter.scala 41:19]
    inst ram of SerialRAM @[SerialAdapter.scala 64:24]
    ram.clock <= buildtopClock
    ram.reset <= buildtopReset
    ram.io.ser <= bits @[SerialAdapter.scala 65:19]
    node _success_T = asUInt(buildtopReset) @[HarnessBinders.scala 307:112]
    inst success_sim of SimSerial @[SerialAdapter.scala 101:23]
    success_sim.exit is invalid
    success_sim.serial is invalid
    success_sim.reset is invalid
    success_sim.clock is invalid
    success_sim.clock <= buildtopClock @[SerialAdapter.scala 102:20]
    success_sim.reset <= _success_T @[SerialAdapter.scala 103:20]
    success_sim.serial <= ram.io.tsi_ser @[SerialAdapter.scala 104:21]
    when success_sim.exit : @[HarnessBinders.scala 308:24]
      io.success <= UInt<1>("h1") @[HarnessBinders.scala 308:37]
    inst plusarg_reader_1 of plusarg_reader_151 @[PlusArg.scala 80:11]
    plusarg_reader_1.out is invalid
    chiptop.custom_boot <= plusarg_reader_1.out @[HarnessBinders.scala 329:21]
    chiptop.clock.clock <= buildtopClock @[HarnessBinders.scala 340:17]
    node _chiptop_reset_T = asAsyncReset(buildtopReset) @[HarnessBinders.scala 342:51]
    chiptop.reset <= _chiptop_reset_T @[HarnessBinders.scala 342:31]
    inst spi_mem_0 of SimSPIFlashModel @[SPIFlash.scala 79:27]
    spi_mem_0.reset is invalid
    spi_mem_0.dq is invalid
    spi_mem_0.cs is invalid
    spi_mem_0.sck is invalid
    spi_mem_0.sck <= chiptop.spi_0.sck @[SPIFlash.scala 81:22]
    spi_mem_0.cs[0] <= chiptop.spi_0.cs[0] @[SPIFlash.scala 83:24]
    attach (spi_mem_0.dq[0], chiptop.spi_0.dq[0]) @[SPIFlash.scala 84:61]
    attach (spi_mem_0.dq[1], chiptop.spi_0.dq[1]) @[SPIFlash.scala 84:61]
    attach (spi_mem_0.dq[2], chiptop.spi_0.dq[2]) @[SPIFlash.scala 84:61]
    attach (spi_mem_0.dq[3], chiptop.spi_0.dq[3]) @[SPIFlash.scala 84:61]
    node _spi_mem_0_io_reset_T = asUInt(buildtopReset) @[SPIFlash.scala 85:33]
    spi_mem_0.reset <= _spi_mem_0_io_reset_T @[SPIFlash.scala 85:24]
    inst simdram of SimDRAM @[HarnessBinders.scala 190:23]
    simdram.axi is invalid
    simdram.reset is invalid
    simdram.clock is invalid
    simdram.axi <= chiptop.axi4_mem_0.bits @[HarnessBinders.scala 191:18]
    simdram.clock <= chiptop.axi4_mem_0.clock @[HarnessBinders.scala 210:20]
    simdram.reset <= chiptop.axi4_mem_0.reset @[HarnessBinders.scala 211:20]
    inst uart_sim_0 of UARTAdapter @[UARTAdapter.scala 132:28]
    uart_sim_0.clock <= clock
    uart_sim_0.reset <= reset
    uart_sim_0.io.uart.txd <= chiptop.uart_0.txd @[UARTAdapter.scala 134:28]
    chiptop.uart_0.rxd <= uart_sim_0.io.uart.rxd @[UARTAdapter.scala 135:18]
    wire refClkBundle : { clock : Clock, reset : Reset} @[TestHarness.scala 39:27]
    buildtopClock <= refClkBundle.clock @[TestHarness.scala 103:17]
    wire _buildtopReset_WIRE : Reset
    _buildtopReset_WIRE <= refClkBundle.reset
    buildtopReset <= _buildtopReset_WIRE @[TestHarness.scala 104:17]
    wire implicitHarnessClockBundle : { clock : Clock, reset : Reset} @[TestHarness.scala 106:40]
    implicitHarnessClockBundle.clock <= clock @[TestHarness.scala 107:36]
    implicitHarnessClockBundle.reset <= reset @[TestHarness.scala 108:36]
    refClkBundle.clock <= implicitHarnessClockBundle.clock @[TestHarness.scala 73:47]
    refClkBundle.reset <= implicitHarnessClockBundle.reset @[TestHarness.scala 74:47]
```

And here is the corresponding Chisel: https://github.com/ucb-bar/testchipip/blob/f99b1eb59a34d7934059904f03237d5bc8d4a680/src/main/scala/SPIFlash.scala#L84

Basically `Analog` ports from two modules (under the `TestHarness`) are being connected (one module is our DUT and the other is a SPI flash module). Are you saying this instance-to-instance connection isn't supported (only port-to-instance)?

common.mk
abejgonzalez: I tried doing `sed -i` calls to avoid having a bunch of extra temporary files that are really unused. I generally am comfortable with adding `sed -i`'s but if there is a better recommendation I'd be happy to use it.

In the interim, I think I'll add a flag to override sed (so you don't have to hack it in).

Side note: Do you use the Conda setup on Mac? I haven't tested it there and assumed it didn't work on Mac.

tymcauley: Sounds great to me! I'm actually not using the Conda setup on Mac. I basically just use my Mac for Verilog builds, not for simulations. I just built the `firtool` binary myself from source, and have it on my `PATH` for testing out this branch.

common.mk
```diff
@@ -111,42 +109,100 @@ $(FIRRTL_FILE) $(ANNO_FILE) &: $(SCALA_SOURCES) $(sim_files) $(SCALA_BUILDTOOL_D
...
+		--verify-each=false \
+		--disable-annotation-classless \
+		--disable-annotation-unknown \
+		--warn-on-unprocessed-annotations \
+		--lowering-options=disallowPackedArrays,emittedLineLength=2048,noAlwaysComb,disallowLocalVariables,explicitBitcast,verifLabels,locationInfoStyle=wrapInAtSquareBracket \
+		--repl-seq-mem \
+		--repl-seq-mem-circuit=$(MODEL) \
+		--repl-seq-mem-file=$(FIRTOOL_SMEMS_CONF) \
+		--annotation-file=$(SFC_ANNO_FILE) \
+		--split-verilog \
```
tymcauley, line 178: Would it be a huge pain to make `--split-verilog` optional? I could imagine some flows wanting to squash everything into a single file. It might be kinda ugly to implement that, since removing that flag makes the `-o` option behave differently...

common.mk
```diff
@@ -111,42 +109,100 @@ $(FIRRTL_FILE) $(ANNO_FILE) &: $(SCALA_SOURCES) $(sim_files) $(SCALA_BUILDTOOL_D
...
+		--annotation-file $(FINAL_ANNO_FILE) \
 		--log-level $(FIRRTL_LOGLEVEL) \
+		--allow-unrecognized-annotations \
+		-X $(if $(shell grep "Fixed<" $(FIRRTL_FILE)),middle,none) \
 		$(EXTRA_FIRRTL_OPTIONS))
-	touch $(sim_top_blackboxes) $(sim_harness_blackboxes)
+	$(if $(shell grep "Fixed<" $(FIRRTL_FILE)),mv $(SFC_FIRRTL_BASENAME).mid.fir $(SFC_FIRRTL_FILE),)
+	firtool \
+		--format=fir \
+		-O=release \
```
tymcauley, line 165: We might also want to be able to toggle this flag between `-O=release` and `-O=debug`, since the latter could be useful for verification. @seldridge can probably provide more context on the intended use of these flags.

scripts/split-mems-conf.py
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env python
```
tymcauley, line 1: Could we use `#!/usr/bin/env python3`? On my mac, the `python` binary either doesn't exist, or actually points to python 2. Are any Linux systems _only_ shipping with `python` binaries which are actually python 3?

scripts/split-module-files.py
```diff
@@ -0,0 +1,111 @@
+#!/usr/bin/env python
```
tymcauley, line 1: Same as above comment regarding python 2 vs python 3.

common.mk
abejgonzalez: I think it would be relatively painful but I can try messing around with things. Would it be possible for the flows to just take the `.f` filelist files and do a large `cat`?

tymcauley: So, that _might_ work, but it looks like each of the split files has lots of header material, which looks like:

```verilog
// Generated by CIRCT unknown git version
// Standard header to adapt well known macros to our needs.
`ifdef RANDOMIZE_REG_INIT
  `define RANDOMIZE
`endif // RANDOMIZE_REG_INIT
`ifdef RANDOMIZE_MEM_INIT
  `define RANDOMIZE
`endif // RANDOMIZE_MEM_INIT
...
```

And so on for a bunch more text macros. So while the `cat`-them-all-together strategy might not technically result in errors, I can see people getting annoyed by a bunch of warnings for redefined text macros.

.github/scripts/defaults.sh
seldridge: Nah, this should work. Anytime we can avoid emitting an alias/bi-directional assign is good as tool support is extremely spotty. We just don't happen to do this and only ever pass analog up or down (as far as I know).

Tracking issue: https://github.com/llvm/circt/issues/4112

common.mk
seldridge: You can also opt to just remove this as `-O=release` is the default. There's a lot of history to this option which is detailed in [this section](https://circt.llvm.org/docs/Dialects/FIRRTL/RationaleFIRRTL/#name-preservation-modes) of the FIRRTL Rational document (which is missing the final chapter, discussed below).

tl;dr: we are pretty confident that `-O=release` is now good enough for everyone as of the newest version of CIRCT, v1.19.0.

Having `-O=release` (preservation of named signals is best effort) and `-O=debug` (preservation of named signals is mandatory) as well as the underlying option that backs these `-preserve-values=[none, named, all]`) was a compromise between simulation speed/PPA and debuggability. Early versions of CIRCT were far too aggressive in name removal. However, after a ton of effort we've gotten it to a point where we believe (and have received feedback that) `-O=release` is better than the old SFC behavior (which was favored by designers) and better than existing MFC behavior of either `-O=release` or `-O=debug` to this point.

`-O=release` is already the default behavior and is expected to be the one, true optimization level that is used to compile designs. `-O=debug` may be useful if you are trying to debug something and desperately want to keep all named things from Chisel or `-preserve-values=all` may be an even stronger fallback of preserving named, temporary things from Chisel. However, neither `-O=debug` or `-preserve-values=all` are expected to be normal modes at which design or verification/PD use. For one-off debugging, I would suggest using a don't touch.

common.mk
tymcauley: Given this CIRCT issue (https://github.com/llvm/circt/issues/4111), it seems like a bad idea to pursue making `--split-verilog` optional for the time being. If `firtool` behavior changes, then maybe we can revisit this.

tymcauley:
A strange issue I'm having with this design of using SFC-then-`firtool`: We have some analysis-only SFC passes that we've written, which rely on some custom FIRRTL annotations inserted in Chisel. So, those annotations get emitted during the Chisel compile stage, and the SFC runs, and runs our analysis passes, and then runs the `ChirrtlEmitter`. However, our analysis passes depend on getting part-way through the low-FIRRTL passes. Thus, the `ChirrtlEmitter` runs _after_ all of that lowering has happened, and emits a bunch of partially-lowered FIRRTL. So far, I've found that this causes issues with inferring how to flatten SRAMs. I can provide some more details if needed, but a fix would be if there were some way to ensure the `ChirrtlEmitter` runs at the start of the transform pipeline, rather that getting moved to the end.

To be clearer, for a "normal" design (without our extra FIRRTL passes), the SFC invocation looks like this (with `FIRRTL_LOGLEVEL` set to `info`):

```
[info]   firrtl.stage.transforms.Compiler
[info]   ├── firrtl.transforms.BlackBoxSourceHelper
[info]   └── firrtl.ChirrtlEmitter
```

In that case, SRAM flattening works perfectly, and the Verilog emitted by `firtool` looks as I'd expect.

However, when our annotations get inserted, which trigger some FIRRTL passes which require lower forms of FIRRTL, the SFC run looks more like this:

```
[info]   firrtl.stage.transforms.Compiler
[info]   ├── firrtl.stage.transforms.CheckScalaVersion
[info]   ├── firrtl.passes.CheckChirrtl$
[info]   ├── firrtl.passes.CInferTypes$
[info]   ├── firrtl.passes.CInferMDir$
[info]   ├── firrtl.passes.RemoveCHIRRTL$
...
[info]   ├── firrtl.passes.LowerTypes$
...
[info]   ├── our.secret.transform.AnalyzeCircuit
[info]   ├── firrtl.transforms.BlackBoxSourceHelper
[info]   └── firrtl.ChirrtlEmitter
```

Our FIRRTL transforms have empty `optionalPrerequisiteOf` definitions, so there shouldn't be any reason the `ChirrtlEmitter` _can't_ run earlier, but I don't really know how we'd make that happen. Any ideas? Thanks!

common.mk
```diff
@@ -111,42 +109,100 @@ $(FIRRTL_FILE) $(ANNO_FILE) &: $(SCALA_SOURCES) $(sim_files) $(SCALA_BUILDTOOL_D
...
 
+$(TOP_MODS_FILELIST) $(MODEL_MODS_FILELIST) $(ALL_MODS_FILELIST) $(BB_MODS_FILELIST) &: $(FIRTOOL_MODEL_MOD_HRCHY_JSON) $(FIRTOOL_FILELIST) $(FIRTOOL_BB_MODS_FILELIST)
+	$(base_dir)/scripts/split-module-files.py \
+		--model-hier-json $(FIRTOOL_MODEL_HRCHY_JSON) \
+		--dut $(TOP) \
+		--out-dut-filelist $(TOP_MODS_FILELIST) \
+		--out-model-filelist $(MODEL_MODS_FILELIST) \
+		--in-all-filelist $(FIRTOOL_FILELIST) \
+		--target-dir $(OUT_DIR)
+	$(SED) -e 's;^;$(OUT_DIR)/;' $(FIRTOOL_BB_MODS_FILELIST) > $(BB_MODS_FILELIST)
```
tymcauley, line 192: If you generate a design which doesn't have any black-box modules, then `$(OUT_DIR)/firrtl_black_box_resource_files.f` won't exist, and this line throws an error. Maybe we should `touch` that path to cover that case?

abejgonzalez: SGTM. Will do.

.github/scripts/defaults.sh
seldridge: This issue is now fixed (thanks to a contribution by @tymcauley!).

abejgonzalez: I just saw it. Thanks shoutout @tymcauley!

common.mk
```diff
@@ -13,14 +11,15 @@ endif
...
-"   ENABLE_SBT_THIN_CLIENT = if set, use sbt's experimental thin client (works best when overridding SBT_BIN with the mainline sbt script)" \
-"   EXTRA_CHISEL_OPTIONS   = additional options to pass to the Chisel compiler" \
-"   EXTRA_FIRRTL_OPTIONS   = additional options to pass to the FIRRTL compiler"
+"   EXTRA_GENERATOR_REQS      = additional make requirements needed for the main generator" \
+"   EXTRA_SIM_CXXFLAGS        = additional CXXFLAGS for building simulators" \
+"   EXTRA_SIM_LDFLAGS         = additional LDFLAGS for building simulators" \
+"   EXTRA_SIM_SOURCES         = additional simulation sources needed for simulator" \
+"   EXTRA_SIM_REQS            = additional make requirements to build the simulator" \
+"   ENABLE_SBT_THIN_CLIENT    = if set, use sbt's experimental thin client (works best when overridding SBT_BIN with the mainline sbt script)" \
+"   ENABLE_CUSTOM_FIRRTL_PASS = if set, enable custom firrtl passes (SFT lowers to Low FIRRTL & CFT does the rest) \
```
abejgonzalez, line 20:
```suggestion
"   ENABLE_CUSTOM_FIRRTL_PASS = if set, enable custom firrtl passes (SFC lowers to LowFIRRTL & FIRTOOL converts to Verilog) \
```

common.mk
```diff
@@ -142,15 +141,34 @@ FIRTOOL_TARGETS = \
 	$(FIRTOOL_SMEMS_CONF) \
 	$(FIRTOOL_TOP_SMEMS_JSON) \
 	$(FIRTOOL_TOP_HRCHY_JSON) \
-	$(FIRTOOL_MODEL_MOD_HRCHY_JSON) \
+	$(FIRTOOL_MODEL_HRCHY_JSON) \
```
abejgonzalez, line 144: It might be good to have a linter go over the Makefile to make sure nothing is broken (maybe look into this pre-commit hook: https://github.com/mrtazz/checkmake#pre-commit-usage)

common.mk
```diff
@@ -142,15 +141,34 @@ FIRTOOL_TARGETS = \
...
 	$(FIRTOOL_MODEL_SMEMS_JSON) \
 	$(FIRTOOL_FILELIST) \
 	$(FIRTOOL_BB_MODS_FILELIST)
 
+ifeq (,$(ENABLE_CUSTOM_FIRRTL_PASS))
```
abejgonzalez, line 149: I would add a comment here describing the intent of the variables (aka what does the repl-seq-mem pass, etc)

common.mk
```diff
@@ -159,9 +177,14 @@ $(FIRTOOL_TARGETS) &: $(FIRRTL_FILE) $(FINAL_ANNO_FILE) $(VLOG_SOURCES)
 		--annotation-file $(FINAL_ANNO_FILE) \
 		--log-level $(FIRRTL_LOGLEVEL) \
 		--allow-unrecognized-annotations \
-		-X $(if $(shell grep "Fixed<" $(FIRRTL_FILE)),middle,none) \
+	  $(REPL_SEQ_MEM) \
```
abejgonzalez, line 180: Spacing issue here

common.mk
```diff
@@ -142,15 +141,34 @@ FIRTOOL_TARGETS = \
...
+	REPL_SEQ_MEM = none
+	TOP_TARGETS = none
+	HARNESS_TARGETS = none
+	TRANSFORMS = barstools.tapeout.transforms.GenerateTop
+else
+	REPL_SEQ_MEM = --infer-rw --repl-seq-mem -c:$(MODEL):-o:$(TOP_SMEMS_CONF)
+	TOP_TARGETS = $(TOP_SMEMS_CONF)
+	HARNESS_CONF_FLAGS = -thconf $(HARNESS_SMEMS_CONF)
+	HARNESS_TARGETS = $(HARNESS_SMEMS_CONF)
+	TRANSFORMS = barstools.tapeout.transforms.GenerateTopAndHarness
```
abejgonzalez, line 159: Technically this is the SFC name that holds the xforms. Maybe rename it to something like `SFC_MAIN`?

common.mk
```diff
@@ -142,15 +141,34 @@ FIRTOOL_TARGETS = \
...
 	$(FIRTOOL_TOP_SMEMS_JSON) \
 	$(FIRTOOL_TOP_HRCHY_JSON) \
-	$(FIRTOOL_MODEL_MOD_HRCHY_JSON) \
+	$(FIRTOOL_MODEL_HRCHY_JSON) \
 	$(FIRTOOL_MODEL_SMEMS_JSON) \
 	$(FIRTOOL_FILELIST) \
 	$(FIRTOOL_BB_MODS_FILELIST)
 
+ifeq (,$(ENABLE_CUSTOM_FIRRTL_PASS))
+	REPL_SEQ_MEM = none
```
abejgonzalez, line 150: I'm not sure I like this `none` usage. It should be clear that certain recipes generate certain files (and not have a layer of abstraction with "dummy" files called `none`). Maybe instead of having 2 steps (generate the collateral needed to generate the `TOP/HARNESS_SMEMS_CONF`, then actually generate it) instead just combine https://github.com/ucb-bar/chipyard/blob/092a4a3b894684106d7b7d95772eb590e95815c6/common.mk#L198-L204 into the https://github.com/ucb-bar/chipyard/blob/092a4a3b894684106d7b7d95772eb590e95815c6/common.mk#L150 step. Then your targets would just be generated in 1 step:

```make
FIRRTL_COMPILER_TARGETS = \
	$(FIRTOOL_TOP_HRCHY_JSON) \
	$(FIRTOOL_MODEL_MOD_HRCHY_JSON) \
	$(FIRTOOL_FILELIST) \
	$(FIRTOOL_BB_MODS_FILELIST) \
        $(TOP_SMEMS_CONF) \
        $(HARNESS_SMEMS_CONF)
```

common.mk
```diff
@@ -159,9 +177,14 @@ $(FIRTOOL_TARGETS) &: $(FIRRTL_FILE) $(FINAL_ANNO_FILE) $(VLOG_SOURCES)
...
 		--allow-unrecognized-annotations \
-		-X $(if $(shell grep "Fixed<" $(FIRRTL_FILE)),middle,none) \
+	  $(REPL_SEQ_MEM) \
+		$(HARNESS_CONF_FLAGS) \
+		-X $(SFC_LEVEL) \
 		$(EXTRA_FIRRTL_OPTIONS))
-	$(if $(shell grep "Fixed<" $(FIRRTL_FILE)),mv $(SFC_FIRRTL_BASENAME).mid.fir $(SFC_FIRRTL_FILE),)
+	-mv $(SFC_FIRRTL_BASENAME).mid.fir $(SFC_FIRRTL_FILE)
+	-mv $(SFC_FIRRTL_BASENAME).lo.fir $(SFC_FIRRTL_FILE)
+	$(if $(ENABLE_CUSTOM_FIRRTL_PASS), cat $(SFC_ANNO_FILE) | jq 'del(.[] | select(.target | test("io.cpu"))?)' > $(build_dir)/tmp.json,)
```
abejgonzalez, line 186: I thought you just needed to delete the `SRAMAnnotation`? What is this `io.cpu`? and `select...` stuff?

abejgonzalez: Also can we not do this within the same file? If not, maybe dump the tmp file to `/tmp/descriptive-name-here`.

common.mk
```diff
@@ -172,38 +195,40 @@ $(FIRTOOL_TARGETS) &: $(FIRRTL_FILE) $(FINAL_ANNO_FILE) $(VLOG_SOURCES)
...
-	sed -i 's/\.\///' $(TOP_MODS_FILELIST)
-	sed -i 's/\.\///' $(MODEL_MODS_FILELIST)
-	sed -i 's/\.\///' $(BB_MODS_FILELIST)
+	$(SED) -e 's;^;$(OUT_DIR)/;' $(FIRTOOL_BB_MODS_FILELIST) > $(BB_MODS_FILELIST)
+	$(SED) -i 's/\.\///' $(TOP_MODS_FILELIST)
+	$(SED) -i 's/\.\///' $(MODEL_MODS_FILELIST)
+	$(SED) -i 's/\.\///' $(BB_MODS_FILELIST)
 	sort -u $(TOP_MODS_FILELIST) $(MODEL_MODS_FILELIST) $(BB_MODS_FILELIST) > $(ALL_MODS_FILELIST)
 
+ifeq (,$(ENABLE_CUSTOM_FIRRTL_PASS))
```
abejgonzalez, line 223: Can't comment on the barstools change, but can you remove the extra pass I attempted to write and open a PR in that repo

common.mk
joey0320: By not in the same file, do you mean make a script that does this `jq` stuff? After adding back the `REPL_SEQ_MEM` option, all the `SRAMAnnotation` stuff went away & only the `io.cpu.dcache` related things (they were all `dontTouchTransforms` & they were there with the `SRAMAnnotation`s when transforming to LowFIRRTL).

common.mk
joey0320: I tried merging these two, but it started to break for the `MidFIRRTL` to `Verilog` case. I think this is because the `SFC` pass that strips out SRAMs and replaces them with blackboxes are performed in `MidFIRRTL` & if `REPL_SEQ_MEM` is not `none`, the SFC will perform this pass for designs with `Fixed` inside them. Hence, `FIRTOOL` does not generate the `FIRTOOL_SMEMS_CONF` file because the `SRAM` pass has been already executed by `SFC`. [](https://cs.github.com/chipsalliance/firrtl/blob/135739f0c7ef3f6404ba7115f77c7d7b913f6748/src/main/scala/firrtl/passes/memlib/ReplaceMemTransform.scala#L118) this link seems to show that the pass is performed in `MidFIRRTL`. 
Admittedly, the current version is ugly & I would also like to come up with a better method. Any suggestions?

common.mk
joey0320: Done!

common.mk
joey0320: Sure, will do.

common.mk
abejgonzalez: So if I understand correctly, this can be fixed by ignoring the `MidFIRRTL` case and instead having 2 flows (SFC NoOp -> All Firtool, SFC to LowFIRRTL -> Firtool LowFIRTOOL to Vlog)?

common.mk
abejgonzalez: Basically is there a way to have `jq` operate in place. In other words, instead of generating a `tmp.json` file, can you have a single `jq` operation that does `jq --some-args --in-place $(JSON)` (like `sed -e`)

joey0320: I had to use `tmp.json` because if I did something like `cat some_file | jq 'some operation' > some_file` (which would be the commands w/o `tmp.json`), `some_file` became a blank file. I'll update it to use `/tmp/descriptive-name`.

joey0320: `jq` supports this, but for some reason, if I do something like `jq do-sth input_file > input_file`, the `input_file` becomes blank.

common.mk
```diff
@@ -177,14 +170,16 @@ endif
...
+		$(REPL_SEQ_MEM) \
 		$(HARNESS_CONF_FLAGS) \
 		-X $(SFC_LEVEL) \
 		$(EXTRA_FIRRTL_OPTIONS))
 	-mv $(SFC_FIRRTL_BASENAME).mid.fir $(SFC_FIRRTL_FILE)
 	-mv $(SFC_FIRRTL_BASENAME).lo.fir $(SFC_FIRRTL_FILE)
-	$(if $(ENABLE_CUSTOM_FIRRTL_PASS), cat $(SFC_ANNO_FILE) | jq 'del(.[] | select(.target | test("io.cpu"))?)' > $(build_dir)/tmp.json,)
-	$(if $(ENABLE_CUSTOM_FIRRTL_PASS), cat $(build_dir)/tmp.json > $(SFC_ANNO_FILE) && rm $(build_dir)/tmp.json,)
+	-cat $(TOP_SMEMS_CONF)
+	-cat $(HARNESS_SMEMS_CONF)
```
abejgonzalez, line 180: Extra debug lines?

common.mk
```diff
@@ -111,42 +110,111 @@ $(FIRRTL_FILE) $(ANNO_FILE) &: $(SCALA_SOURCES) $(sim_files) $(SCALA_BUILDTOOL_D
...
+	$(FIRTOOL_TOP_SMEMS_JSON) \
+	$(FIRTOOL_TOP_HRCHY_JSON) \
+	$(FIRTOOL_MODEL_HRCHY_JSON) \
+	$(FIRTOOL_MODEL_SMEMS_JSON) \
+	$(FIRTOOL_FILELIST) \
+	$(FIRTOOL_BB_MODS_FILELIST) \
+	$(TOP_SMEMS_CONF) \
+	$(HARNESS_SMEMS_CONF)
+
+SFC_MAIN = barstools.tapeout.transforms.GenerateTopAndHarness
```
abejgonzalez, line 151: Does this need to be a variable if it never should change? I would inline it.

abejgonzalez (commented):
I have some comments on the other barstools PR but overall looking good. You should start looking to resolve some of the other comments when you get the chance.

common.mk
abejgonzalez: Can you add a comment on the `io.cpu` `DontTouchAnnos` causing issues and needing to be deleted from the JSON. I can see this being another issue in the future if people put `DontTouchAnnos` on SRAM items.

joey0320:
> I have some comments on the other barstools PR but overall looking good. You should start looking to resolve some of the other comments when you get the chance.

One thing I wanted to mention about this fix is that I'm not sure if this produces the exact compilation steps that we want. When using the `REPL_SEQ_MEM` option without `none`, the `SFC_MAIN` performs a bunch of FIRRTL passes even though the `-X` option is set to `none`. Also, it seems less robust compared to the previous one since a bunch of CI tests are failing & started to complain about SRAM stuff again.

jerryz123:
Can you open PRs to Ibex and Sodor associated with this change?

joey0320:
> Can you open PRs to Ibex and Sodor associated with this change?

Sure, but I don't see any changes that are need to be made to the cores. Are there any specific changes required?

jerryz123:
This PR currently bumps those submodules. @abejgonzalez did you bump those?

joey0320:
Abe bumped those. Made PRs, can you check?

jerryz123:
I merged the PRs.

jerryz123:
Can we address Schuyler's and Tynan's comments? I think they got lost in the long history of this PR.

On firtool flags:
 - https://github.com/ucb-bar/chipyard/pull/1239/files#r995243134
 - https://github.com/ucb-bar/chipyard/pull/1239/files#r995247072
 - https://github.com/ucb-bar/chipyard/pull/1239/files#r995245501
 
 On explicit python3
 - https://github.com/ucb-bar/chipyard/pull/1239/files#r996126886
 - https://github.com/ucb-bar/chipyard/pull/1239/files#r996127177

joey0320:
> Can we address Schuyler's and Tynan's comments? I think they got lost in the long history of this PR.
> 
> On firtool flags:
> 
> * https://github.com/ucb-bar/chipyard/pull/1239/files#r995243134
> * https://github.com/ucb-bar/chipyard/pull/1239/files#r995247072
> * https://github.com/ucb-bar/chipyard/pull/1239/files#r995245501
> 
> On explicit python3
> 
> * https://github.com/ucb-bar/chipyard/pull/1239/files#r996126886
> * https://github.com/ucb-bar/chipyard/pull/1239/files#r996127177

done

common.mk
```diff
@@ -146,29 +146,29 @@ FIRTOOL_TARGETS = \
...
-	HARNESS_TARGETS = $(HARNESS_SMEMS_CONF)
-	TRANSFORMS = barstools.tapeout.transforms.GenerateTopAndHarness
-endif
+SFC_REPL_SEQ_MEM = --infer-rw --repl-seq-mem -c:$(MODEL):-o:$(SFC_SMEMS_CONF)
 
 
 # DOC include start: FirrtlCompiler
-# hack: lower to middle firrtl if Fixed types are found
-$(HARNESS_TARGETS) $(TOP_TARGETS) $(FIRTOOL_TARGETS) &: $(FIRRTL_FILE) $(FINAL_ANNO_FILE) $(VLOG_SOURCES)
+# This step can take either one of two paths. The first path is when SFC
```
abejgonzalez, line 153: This is incorrect. 

1st case: The Chisel generator creates CHIRRTL from the Chisel files. Then FIRTOOL converts this CHIRRTL into Verilog
2nd case: The Chisel generator creates the CHIRRTL from the Chisel files. Then SFC converts CHIRRTL to LowFIRRTL. Then FIRTOOL converts this CHIRRTL into Verilog.

abejgonzalez: Non-blocking Nit: I'm not sure whether we should refer to things as `firtool`, `FIRTOOL`, `CIRCT`, `circt`, etc. Probably worth snooping other sources... but maybe we should call this `MFC` (MLIR FIRRTL compiler) (see https://github.com/sifive/chisel-circt) to match `SFC`

common.mk
```diff
@@ -146,29 +146,29 @@ FIRTOOL_TARGETS = \
...
+# compiles Chisel to CHIRRTL, and FIRTOOL compiles CHIRRTL to Verilog. Otherwise,
+# when custom FIRRTL transforms are included or if a Fixed type is used within
+# the dut, SFC compiles Chisel to LowFIRRTL and FIRTOOL compiles it to Verilog.
+# Users can indicate to the Makefile of custom FIRRTL transforms by setting the
+# "ENABLE_CUSTOM_FIRRTL_PASS" env variable.
```
abejgonzalez, line 158:
```suggestion
# "ENABLE_CUSTOM_FIRRTL_PASS" variable.
```

Since this can be set by the Makefile, as a Make arg, etc (not just an env. var)

common.mk
```diff
@@ -146,29 +146,29 @@ FIRTOOL_TARGETS = \
...
+#
+# hack: lower to low firrtl if Fixed types are found
+# hack: when using dontTouch, io.cpu annotations are not removed by SFC,
```
abejgonzalez, line 161: This should turn into an issue. I think there is probably a better way to deal with this but I don't want to block the PR any further.

common.mk
```diff
@@ -177,32 +177,30 @@ endif
...
 		--lowering-options=emittedLineLength=2048,noAlwaysComb,disallowLocalVariables,explicitBitcast,verifLabels,locationInfoStyle=wrapInAtSquareBracket \
 		--repl-seq-mem \
-		--repl-seq-mem-circuit=$(MODEL) \
 		--repl-seq-mem-file=$(FIRTOOL_SMEMS_CONF) \
+		--repl-seq-mem-circuit=$(MODEL) \
 		--annotation-file=$(SFC_ANNO_FILE) \
 		--split-verilog \
 		-o $(OUT_DIR) \
 		$(SFC_FIRRTL_FILE)
+	-mv $(SFC_SMEMS_CONF) $(FIRTOOL_SMEMS_CONF)
```
abejgonzalez, line 203: Is this the correct location for this `mv`? Wouldn't the SFC smems file override the FIRTOOL one (if they both generate an smems file)? Or are you implicitly depending that either FIRTOOL or SFC generate the file separately and never at the same time?

scripts/split-mems-conf.py
```diff
@@ -23,43 +24,73 @@
 }
 """
 
+sys.setrecursionlimit(100)
+
+
+
```
abejgonzalez, line 30: Nit: Spacing

scripts/split-mems-conf.py
```diff
@@ -23,43 +24,73 @@
...
-
-                    if name in dut_mods:
-                        odsc.write(l)
-                    elif name in model_mods:
-                        otsc.write(l)
-                    else:
-                        assert False, "Unable to find smem CONF module in firtool emitted JSON files."
-                else:
-                    exit(0)
+  parser = argparse.ArgumentParser(description='Use CIRCT (firtool) model-hrchy JSONs to create DUT and test harness smems confs or ')
```
abejgonzalez, line 61: Nit: Match FIRTOOL capitalization and/or naming.

abejgonzalez: Nit: Match test harness naming

abejgonzalez: This description needs to be updated/fixed.

scripts/split-mems-conf.py
```diff
@@ -23,43 +24,73 @@
...
+        # the line can't be split then stop immediately (normally an empty file)
+        if len(sl) > 2:
+          name = sl[1]
+
+          if name in dut_submodules:
+            odsc.write(l)
+          elif name in model_submodules:
+            otsc.write(l)
+          else:
+            assert False, "Unable to find smem CONF module in firtool emitted JSON files."
```
abejgonzalez, line 94: Ditto

abejgonzalez (commented):
Sorry for the delay. Final questions and nits added.

common.mk
joey0320: IIRC `SFC_SMEMS_CONF` is only generated in the SFC-LowFIRRTL case, while `FIRTOOL_SMEMS_CONF` is generated only in the SFC-CHIRRTL case. So this implicitly assumes that the smems file is generated by by either the SFC or FIRTOOL, not both.

common.mk
joey0320: Should I make a issue now or after this is merged?

abejgonzalez: Whenever works as long as you remember it

jerryz123 (approved):

jerryz123:
I think you'll need to rebase this branch against main, or merge main into this branch.

jerryz123:
Great stuff. Everything is in place now. Lets get this in.

PR #1249, opened 2022-10-09
Misc release cleanup

abejgonzalez:
Small things that weren't addressed in the release due to time. Fixes include:
- Cleaning up the `.gitignore` now that folders are removed
- Deduping conda requirements by splitting up the requirements files. This includes renaming them, moving them into a new folder, and adding a new `platforms` field (removes the need for a cmdline option in `build-setup.sh`)
- Fixed a warning that would occur when `source env.sh` is run. The check would occur before the `conda activate ...` in the `env.sh`. This moves the warning to the makefiles
- Now that we use `make >= 4.3` we can use the grouped target functionality (see https://www.gnu.org/software/make/manual/html_node/Multiple-Targets.html) which simplifies the makefile.

**Related PRs / Issues**:

**Type of change**:
- [ ] Bug fix
- [ ] New feature
- [x] Other enhancement

**Impact**:
- [ ] RTL change
- [x] Build system change
- [ ] Other

**Contributor Checklist**:


common.mk
```diff
@@ -9,6 +9,10 @@ else
 $(info Running with RISCV=$(RISCV))
 endif
 
+ifndef CONDA_DEFAULT_ENV
```
jerryz123, line 12: Can this check be merged with the existing check for RISCV in this makefile? If both RISCV and CONDA are undefined, error.

jerryz123 (changes requested):

common.mk
abejgonzalez: My impression was that some people wanted to not use conda hence the warning instead of erroring.

jerryz123: No I mean combine with this: https://github.com/ucb-bar/chipyard/blob/b00da55a870ea5429cf32968a47a3e5db297ac4a/common.mk#L6-L10

If `RISCV` is not defined, throw an error (as before), except change the error message to ask the user if they did the Conda stuff specifically.

If `RISCV` is defined, don't throw any error or warning.

abejgonzalez: Ok, so you don't want a conda check. Instead, just check RISCV and just change the error message to point to conda. Correct?

jerryz123: Yeah

jerryz123 (approved):

PR #1272, opened 2022-11-22
Revert "fix: S-interpolator for assert, assume and printf"

jerryz123:
Reverts ucb-bar/chipyard#1242

My bad for merging that, we aren't up-to-date enough to use that fix yet.


abejgonzalez (approved):
Hmmmmm

PR #1301, opened 2023-01-01
Set ENABLE_SBT_THIN_CLIENT to 1 as default for CY conda-env

joey0320:
**Related PRs / Issues**:

**Type of change**:
- [ ] Bug fix
- [ ] New feature
- [x] Other enhancement

**Impact**:
- [ ] RTL change
- [ ] Build system change
- [x] Other

**Contributor Checklist**:

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -18,7 +18,6 @@ HELP_COMPILATION_VARIABLES += \
 "   EXTRA_SIM_LDFLAGS      = additional LDFLAGS for building simulators" \
 "   EXTRA_SIM_SOURCES      = additional simulation sources needed for simulator" \
 "   EXTRA_SIM_REQS         = additional make requirements to build the simulator" \
-"   ENABLE_SBT_THIN_CLIENT = if set, use sbt's experimental thin client (works best when overridding SBT_BIN with the mainline sbt script)" \
 "   EXTRA_CHISEL_OPTIONS   = additional options to pass to the Chisel compiler" \
 "   EXTRA_FIRRTL_OPTIONS   = additional options to pass to the FIRRTL compiler"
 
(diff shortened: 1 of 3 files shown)


jerryz123:
Perhaps it's neater to make thin client the default in the makefile. And have a new flag DISABLE_THIN_CLIENT use the old behavior. 

Then we don't have to manage another environment variable.

abejgonzalez:
I personally don't like having this be the default since there is no way to auto-terminate the daemon if CY switches branches, is deleted, etc. This flag will lead a lot of folks having dangling build servers lying around that they would have to cancel themselves. I like it disabled for super users to find out about and consciously deal with (including managing when to shutdown the server or realize it's dangling).

joey0320:
> I personally don't like having this be the default since there is no way to auto-terminate the daemon if CY switches branches, is deleted, etc. This flag will lead a lot of folks having dangling build servers lying around that they would have to cancel themselves. I like it disabled for super users to find out about and consciously deal with (including managing when to shutdown the server or realize it's dangling).
> 
> For example, I had to add the following code in my `bashrc` to auto-shutdown any potential server.
> 
> https://github.com/abejgonzalez/dotfiles/blob/771757d8643f417c6e45e80f376ab1af56085a01/.bash_common#L27-L36
> 
> Maybe if we use bloop this problem would be solved (I'm considering re-looking into this since using a bloop build server would integrate well with VSCode + NVim metals).



> I personally don't like having this be the default since there is no way to auto-terminate the daemon if CY switches branches, is deleted, etc. This flag will lead a lot of folks having dangling build servers lying around that they would have to cancel themselves. I like it disabled for super users to find out about and consciously deal with (including managing when to shutdown the server or realize it's dangling).
> 
> For example, I had to add the following code in my `bashrc` to auto-shutdown any potential server.
> 
> https://github.com/abejgonzalez/dotfiles/blob/771757d8643f417c6e45e80f376ab1af56085a01/.bash_common#L27-L36
> 
> Maybe if we use bloop this problem would be solved (I'm considering re-looking into this since using a bloop build server would integrate well with VSCode + NVim metals).

I see. Do you think it would be too messy if we just added the above code to each user's bashrc when we run `build-setup.sh`?

jerryz123:
We should not append stuff to users bashrc. 

I agree with Abe then, we ought to leave this as an optional feature. Power users can add it to their bashrc if they want.

PR #1307, opened 2023-01-06
Spike-as-a-Tile

jerryz123:
This PR integrates a spike processor model as a tile into chipyard.
The spike processor model supports any RISC-V ISA configuration Spike supports, and dynamically links with spike-as-a-library.
The spike processor interfaces with chipyard through a cache-coherent Tilelink model.
Of course, this shouldn't be used for any performance-modelling.

It is useful for stress-testing MMIO devices, or accelerators.
Adding RoCC support would be a cool extension.

Also, atomics aren't guaranteed to be atomic.

TODO:
- [x] some documentation 

**Related PRs / Issues**:

**Type of change**:
- [ ] Bug fix
- [x] New feature
- [ ] Other enhancement

**Impact**:
- [ ] RTL change
- [ ] Build system change
- [x] Other

**Contributor Checklist**:

### DIFF
--- a/generators/chipyard/src/main/resources/csrc/spiketile.cc
+++ b/generators/chipyard/src/main/resources/csrc/spiketile.cc
@@ -252,12 +252,23 @@ extern "C" void spike_tile(int hartid, char* isa,
                                                    dcache_ways, dcache_sets,
                                                    cacheable, uncacheable, readonly_uncacheable, executable,
                                                    icache_sourceids, dcache_sourceids);
+    std::string* isastr = new std::string(isa);
+    cfg_t* cfg = new cfg_t(std::make_pair(0, 0),
+                           nullptr,
+                           isastr->c_str(),
+                           "MSU",
+                           "vlen:128,elen:64",
+                           false,
+                           endianness_little,
+                           false,
+                           pmpregions,
+                           std::vector<mem_cfg_t>(),
+                           std::vector<int>(),
+                           false);
     processor_t* p = new processor_t(isa_parser,
-                                     "vlen:128,elen:64",
+                                     cfg,
                                      simif,
                                      hartid,
-                                     false,
-                                     endianness_little,
                                      false,
                                      log_file->get(),
                                      sout);
@@ -901,10 +912,11 @@ void spike_thread_main(void* arg)
       host->switch_to();
     }
     while (tile->max_insns != 0) {
-      uint64_t last_bits = tile->proc->get_last_bits();
-      if (insn_should_fence(last_bits) && !tile->simif->stq_empty()) {
-        host->switch_to();
-      }
+      // TODO: Fences don't work
+      // uint64_t last_bits = tile->proc->get_last_bits();
+      // if (insn_should_fence(last_bits) && !tile->simif->stq_empty()) {
+      //   host->switch_to();
+      // }
       tile->proc->step(1);
       tile->max_insns--;
       tile->proc->get_state()->mcycle->write(tile->simif->cycle);
--- a/generators/chipyard/src/main/resources/csrc/spiketile.cc
+++ b/generators/chipyard/src/main/resources/csrc/spiketile.cc
@@ -150,6 +150,13 @@ std::map<int, tile_t*> tiles;
 std::ostream sout(nullptr);
 log_file_t* log_file;
 
+extern "C" void spike_tile_reset(int hartid)
+{
+  if (tiles.find(hartid) != tiles.end()) {
+    tiles[hartid]->proc->reset();
+  }
+}
+
 extern "C" void spike_tile(int hartid, char* isa,
                            int pmpregions,
                            int icache_sets, int icache_ways,
--- a/generators/chipyard/src/main/resources/csrc/spiketile.cc
+++ b/generators/chipyard/src/main/resources/csrc/spiketile.cc
@@ -1,0 +1,917 @@
+#include <riscv/simif.h>
+#include <riscv/processor.h>
+#include <riscv/log_file.h>
+#include <fesvr/context.h>
+#include <map>
+#include <sstream>
+#include <vpi_user.h>
+#include <svdpi.h>
+
+enum transfer_t {
+  NToB,
+  NToT,
+  BToT
+};
+
+enum cache_state_t {
+  NONE,
+  BRANCH,
+  TRUNK,
+  DIRTY
+};
+
+struct cache_line_t {
+  cache_state_t state;
+  uint64_t addr;
+  uint64_t data[8];
+};
+
+struct mem_region_t {
+  uint64_t base;
+  uint64_t size;
+};
+
+struct stq_entry_t {
+  uint64_t addr;
+  uint64_t bytes;
+  size_t len;
+};
+
+struct cache_miss_t {
+  bool valid;
+  uint64_t addr;
+  size_t way;
+  transfer_t type;
+};
+
+struct writeback_t {
+  cache_line_t line;
+  cache_state_t desired;
+  uint64_t sourceid;
+  bool voluntary;
+};
+
+
+class chipyard_simif_t : public simif_t
+{
+public:
+  char* addr_to_mem(reg_t addr) override { return NULL; };
+  bool reservable(reg_t addr) override;
+  bool mmio_fetch(reg_t addr, size_t len, uint8_t* bytes) override;
+  bool mmio_load(reg_t addr, size_t len, uint8_t* bytes) override;
+  bool mmio_store(reg_t addr, size_t len, const uint8_t* bytes) override;
+  void proc_reset(unsigned id) override { };
+  const char* get_symbol(uint64_t addr) override { return nullptr; };
+
+  bool icache_a(uint64_t *address, uint64_t *source);
+  void icache_d(uint64_t sourceid, uint64_t data[8]);
+
+  bool mmio_a(uint64_t *address, uint64_t* data, unsigned char* store, int* size);
+  void mmio_d(uint64_t data);
+
+  bool dcache_a(uint64_t *address, uint64_t* source, unsigned char* state_old, unsigned char* state_new);
+  void dcache_b(uint64_t address, uint64_t source, int param);
+  bool dcache_c(uint64_t *address, uint64_t* source, int* param, unsigned char* voluntary, unsigned char* has_data, uint64_t* data[8]);
+  void dcache_d(uint64_t sourceid, uint64_t data[8], unsigned char has_data, unsigned char grantack);
+  void drain_stq();
+  bool stq_empty() { return st_q.size() == 0; };
+  
+  ~chipyard_simif_t() { };
+  chipyard_simif_t(size_t icache_ways,
+                   size_t icache_sets,
+                   size_t dcache_ways,
+                   size_t dcache_sets,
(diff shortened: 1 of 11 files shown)


generators/chipyard/src/main/resources/csrc/spiketile.cc
```diff
@@ -0,0 +1,917 @@
...
+      std::string arg(vinfo.argv[i]);
+      if (arg == "+spike-debug") {
+        p->set_debug(true);
+      }
+      if (arg == "+spike-stq") {
+        simif->use_stq = true;
+      }
+    }
+
+    p->reset();
```
tianrui-wei, line 273: Does this comply with the chipyard reset behabior? Is there some reset tree?

generators/chipyard/src/main/resources/csrc/spiketile.cc
```diff
@@ -0,0 +1,917 @@
...
+  tile_t* tile = tiles[hartid];
+  chipyard_simif_t* simif = tile->simif;
+  processor_t* proc = tile->proc;
+
+  simif->cycle = cycle;
+  if (debug) {
+    proc->halt_request = proc->HR_REGULAR;
+  }
+
+  proc->get_state()->mip->backdoor_write_with_mask(MIP_MTIP, mtip ? MIP_MTIP : 0);
```
tianrui-wei, line 287: How about other CSRs? Like time/cycle etc

generators/chipyard/src/main/resources/vsrc/spiketile.v
```diff
@@ -0,0 +1,427 @@
...
+
+                                             input         mmio_d_valid,
+                                             input [63:0]  mmio_d_data
+ );
+
+   longint                                                 __insns_retired;
+   reg [63:0]                                              __insns_retired_reg;
+
+   wire                                                    __icache_a_ready;
+   bit                                                     __icache_a_valid;
```
tianrui-wei, line 193: why bit? bit is only 2 state

tianrui-wei (commented):
Very impressive work overall :)

generators/chipyard/src/main/resources/vsrc/spiketile.v
jerryz123: Valid is only 2 states?

generators/chipyard/src/main/resources/csrc/spiketile.cc
jerryz123: `mcycle` is written in `spike_thread_main`

jerryz123: `mtime` is accessed over the bus.

generators/chipyard/src/main/resources/csrc/spiketile.cc
jerryz123: You are right I'll modify this to behave correctly with reset.

generators/chipyard/src/main/resources/vsrc/spiketile.v
tianrui-wei: I'm a little surprised as bit is two state (0, 1) but reg, wire are four states (0, 1, X, Z). Perhaps you could use 2-state only types uniformly?

Also the ports are 4 states, as they're implicitly wire

generators/chipyard/src/main/resources/csrc/spiketile.cc
tianrui-wei: I guess the question is does mcycle sync across tiles. For example, if you have multiple spike tiles or 1 spike + 1 rocket tile

jerryz123: I believe mcycle only increments when the harts are out of reset. So its not guaranteed that they are in sync.

The current implementation does keep a consistent mcycle. The counter is tracked in the chisel and passed to the model.

generators/chipyard/src/main/resources/vsrc/spiketile.v
jerryz123: Does it matter? The DPI interface will only ever write 0 or 1 to this.

tianrui-wei: I see. It's fine in that case :)

generators/chipyard/src/main/resources/vsrc/spiketile.v
```diff
@@ -0,0 +1,429 @@
...
+         __insns_retired_reg <= 64'h0;
+
+         __icache_a_valid = 1'b0;
+         __icache_a_valid_reg <= 1'b0;
+         __icache_a_address = 64'h0;
+         __icache_a_address_reg <= 64'h0;
+         __icache_a_sourceid = 64'h0;
+         __icache_a_sourceid_reg <= 64'h0;
+
+         __mmio_a_valid = 1'b0;
```
tianrui-wei, line 275: May I ask why there's a mix of blocking and non-blocking assignment in this always block?

generators/chipyard/src/main/resources/csrc/spiketile.cc
tianrui-wei: Thanks for the explanation. That makes sense.

generators/chipyard/src/main/resources/vsrc/spiketile.v
jerryz123: Doesn't work with verilator otherwise.

tianrui-wei: Can you move the blocking assignments to a separate `always_comb` block?

sims/common-sim-flags.mk
```diff
@@ -3,6 +3,17 @@
 #----------------------------------------------------------------------------------------
 SIM_OPT_CXXFLAGS := -O3
 
+# Workaround: esp-isa-sim doesn't install libriscv,
```
abejgonzalez, line 6: What is `libriscv` used for again? What would break?

generators/chipyard/src/main/scala/config/SpikeConfigs.scala
```diff
@@ -0,0 +1,19 @@
...
+class SpikeConfig extends Config(
+  new chipyard.WithNSpikeCores(1) ++
+  new chipyard.config.AbstractConfig)
+
+// Avoids polling on the UART registers
+class SpikeFastUARTConfig extends Config(
+  new chipyard.WithNSpikeCores(1) ++
+  new chipyard.config.WithUARTFIFOEntries(128, 128) ++
+  new chipyard.config.WithMemoryBusFrequency(1) ++
+  new chipyard.config.WithPeripheryBusFrequency(1) ++
```
abejgonzalez, line 18: Why is this config nice/needed? For Linux how much speedup is obtained?

abejgonzalez: Nit: Can you order the fragments with the number of cores at the bottom (similar to how the Boom/Rocket configs are setup)

generators/chipyard/src/main/scala/SpikeTile.scala
```diff
@@ -0,0 +1,392 @@
...
+  // Private constructor ensures altered LazyModule.p is used implicitly
+  def this(params: SpikeTileParams, crossing: TileCrossingParamsLike, lookup: LookupByHartIdImpl)(implicit p: Parameters) =
+    this(params, crossing.crossingType, lookup, p)
+
+  // Required TileLink nodes
+  val intOutwardNode = IntIdentityNode()
+  val masterNode = visibilityNode
+  val slaveNode = TLIdentityNode()
+
+  override def isaDTS = "rv64gcv_Zfh"
```
abejgonzalez, line 111: Is this true with the current CY toolchain?

generators/chipyard/src/main/scala/SpikeTile.scala
```diff
@@ -0,0 +1,392 @@
...
+import freechips.rocketchip.diplomacy._
+import freechips.rocketchip.rocket._
+import freechips.rocketchip.tilelink._
+import freechips.rocketchip.interrupts._
+import freechips.rocketchip.util._
+import freechips.rocketchip.tile._
+import freechips.rocketchip.prci.ClockSinkParameters
+
+case class SpikeCoreParams(
+  val maxInsnsPerCycle: Int = 10000
```
abejgonzalez, line 19: What's this used for?

abejgonzalez (commented):
I need to look at `spiketile.*` still but here is a temp. review.

sims/common-sim-flags.mk
jerryz123: Its a dynamic library installed by spike. Nothing should break.

generators/chipyard/src/main/scala/config/SpikeConfigs.scala
jerryz123: This makes the UART run at a similar frequency to the uncore, otherwise the UART runs super slowly compare to the core, and the core gets stuck behind polling UART.

generators/chipyard/src/main/scala/SpikeTile.scala
jerryz123: This isn't passed to the toolchain.

generators/chipyard/src/main/scala/SpikeTile.scala
jerryz123: How many cycles spike can execute in a single "tick". (If spike sees a cache miss, however, the actual IPC will be lower).

sims/common-sim-flags.mk
jerryz123: What I mean in the comment is that building `SpikeTileConfig` after the user has installed `esp-isa-sim`, but not `riscv-isa-sim`, will cause a link failure.

abejgonzalez (approved):
LGTM. Skimmed the C++/V side and it seems fine for now (esp. since this is an experimental feature)

generators/chipyard/src/main/scala/SpikeTile.scala
abejgonzalez: I know it's not passed but right now we can't compile code outside `gc` right? I'm not sure it would make sense to restrict this to `rv64gc` for now.

abejgonzalez:
Now that this exists, maybe it makes sense to boot Linux in CI with this.

generators/chipyard/src/main/scala/SpikeTile.scala
jerryz123: rv64gcv_Zfh is a superset of rv64gc. The cpu is always allowed to report that it has more capability than the software needs.

jerryz123:
> Now that this exists, maybe it makes sense to boot Linux in CI with this.

This needs a faster firemarshal linux image,

generators/chipyard/src/main/resources/vsrc/spiketile.v
jerryz123: Our current verilog generation flow disallows generating always_comb. I assume there is some tool this causes problems with, so I will keep the uglier structure here.

tianrui-wei: Sounds good. I'm very excited this PR is merged!

PR #1324, opened 2023-02-03
New Hammer

harrisonliew:
This PR switches the VLSI flow over to new Hammer, which is now published on PyPI: https://pypi.org/project/hammer-vlsi/.
This main package is installed via conda, and plugins packages are installed with `pip -e` using the `init-vlsi.sh` script.

Several bugs/problems are caused by the migration to CIRCT, and should be investigated:
- [x] ~SystemVerilog packed arrays are being emitted, despite adding the `disallowPackedArrays` lowering flag. This means Yosys is unable to synthesize designs, and there is no roadmap/effort there to support multidimensional arrays. Thus, the Sky130 + OpenROAD tutorial therefore does not work anymore.~
	Fixed by llvm/circt#4626
- [x] ~Async resets are being emitted in a way that synthesis tools like Genus do not like. Specifically, we get something of the form:~
	```
	wire reset = ~_some_reset;
	always @(posedge clock or posedge reset)
		if (~_some_reset)
			...
	```
	~This is not allowed because the reset in the `if(...` should be the one in the sensitivity list.~
	~Attempts to use the `exprInEventControl` lowering flag have not solved this.~
	~The workaround is to manually change `if(~_some_reset)` to `if(reset)`. If we are sticking with this workaround, I will update the documentation.~
	Fixed by llvm/circt#4625
- [x] ~Line 64 of `vlsi/Makefile` is an ugly hack to separate out the harness blackboxes that cannot be input to synthesis tools due to the presence of DPI calls. It would be best to investigate ways to restore the top/harness blackbox separation in CIRCT.~
	Fixed by @joey0320 

Finally, these dependencies need to be included:
- [x] ucb-bar/barstools#125
- [x] ucb-bar/firtool-feedstock#3
- [x] Hammer v1.0.0 must be released (to update conda-lock) before merging.

**Related PRs / Issues**:

**Type of change**:
- [ ] Bug fix
- [ ] New feature
- [X] Other enhancement

**Impact**:
- [ ] RTL change
- [x] Build system change
- [X] Other

**Contributor Checklist**:
- [X] Did you set `main` as the base branch?
- [X] Is this PR's title suitable for inclusion in the changelog and have you added a `changelog:<topic>` label?
- [X] Did you state the type-of-change/impact?
- [X] Did you delete any extraneous prints/debugging code?
- [X] Did you mark the PR with a `changelog:` label?
- [X] (If applicable) Did you update the conda `.conda-lock.yml` file if you updated the conda requirements file?
- [X] (If applicable) Did you add documentation for the feature?


vlsi/Makefile
```diff
@@ -54,23 +55,20 @@ endif
...
 ifneq ($(CUSTOM_VLOG), )
 	VLSI_RTL = $(CUSTOM_VLOG)
-	VLSI_BB = /dev/null
+	VLSI_BB =
 else
-	VLSI_RTL = $(ALL_RTL) $(extra_v_includes)
-	VLSI_BB = $(sim_top_blackboxes)
+	VLSI_RTL = $(shell cat $(TOP_MODS_FILELIST)) $(TOP_SMEMS_FILE) 
+	# TODO: have MFC split top & harness blackboxes
+	VLSI_BB = $(build_dir)/EICG_wrapper.v $(shell rev $(BB_MODS_FILELIST) | sed -E 's/cc(.*)/c\1/g' | uniq -s 1 | sed '/c\./d' | rev)
```
abejgonzalez, line 64: - I think having this oneliner will probably result in confusion down the line. Can you add a comment on what it's doing.
- Technically, `VLSI_RTL` could contain BB's if the BB filename matches the name of the module. I'm not sure if that matters. I guess why is there a distinction between the BB's and VLSI_RTL?
- Side note: IIRC most synth. tools will do dead code elimination and thus will eliminate unused modules. So do we really need to split up the BB's between top/model? I understand from a code cleanliness perspective but the solution seems fairly heavy-handed when a downstream tool can deal with this well.

abejgonzalez (changes requested):

vlsi/Makefile
harrisonliew: 1. Sure, can clarify what it's doing.
2. So that I'm clear what you mean - blackboxes where the resource name is equivalent to the module's name get put in the `.top.f` file, and those where the names are different go into `.bb.f`? I can totally merge `VLSI_RTL` and `VLSI_BB` - the latter was just leftover from the previous Makefile.
3. Theoretically you are correct. However, the issue here is that synthesis tools don't support the DPI calls that are present in the harness' blackboxes (SimJTAG, etc.), erroring out as a result. They generally only like purely behavioral + structural Verilog.

abejgonzalez: > 2. So that I'm clear what you mean - blackboxes where the resource name is equivalent to the module's name get put in the `.top.f` file, 

This is true.

> and those where the names are different go into `.bb.f`?

The `bb.f` file holds all BB files for both the top/harness (will include the BB's put in the `top.f` file).

> I can totally merge `VLSI_RTL` and `VLSI_BB` - the latter was just leftover from the previous Makefile.

Combining `top.f` and `bb.f` (ofc removing the non-verilog files) and removing duplicate BB's files will give you all the Verilog files for the TOP and some Verilog BB files that reside in the MODEL.

> 3. Theoretically that should be true. However, the issue here is that synthesis tools don't support the DPI calls that are present in the harness' blackboxes (SimJTAG, etc.), erroring out as a result. They generally only like purely behavioral + structural Verilog.

Agreed. You at minimum need to remove the non-Verilog files.

abejgonzalez:
For future reference: To replicate the CIRCT issues you can try building a `TinyRocketConfig`

vlsi/Makefile
joey0320: currently working on 3

joey0320: @harrisonliew  @abejgonzalez  can you guys check if this is the behavior that we want for separating top.bb.f & model.bb.f?

harrisonliew: Thanks for fixing 3! Updated the Makefile to combine into `VLSI_RTL` and use the new split script.

joey0320:
For the packed array thing, we can go over this issue for now by setting `ENABLE_CUSTOM_FIRRTL_PASS=1`

harrisonliew:
Unfortunately, even with using `ENABLE_CUSTOM_FIRRTL_PASS=1`, there are still remaining SystemVerilog constructs that Yosys doesn't recognize, such as:
`some_wire <= 10'(some_wire - 10'h1);`

It would be nice to have Hammer's Yosys plugin call this: https://github.com/zachjs/sv2v, but the pre-built images require quite new glibc...

abejgonzalez:
I added `sv2v` to our conda repo: https://github.com/ucb-bar/sv2v-feedstock. I tested that running it worked with TinyRocketConfig `sv2v *.sv` for all files in the `gen-collateral` area. You should be able to install it with `conda install -c ucb-bar sv2v` (I would add it to the Hammer requirements when it's build by conda though).

uenoku:
>here are still remaining SystemVerilog constructs that Yosys doesn't recognize, such as:
> 10'(some_wire - 10'h1)

Hi, these static casts are added when `explicitBitcast` is given so you can try removing the option from https://github.com/ucb-bar/chipyard/blob/a6921d15c600517196754264ca89defe3ac04d22/common.mk#L195

abejgonzalez (changes requested):

vlsi/Makefile
```diff
@@ -54,28 +55,37 @@ endif
...
+
+ifneq ($(CUSTOM_VLOG), )
+	RTL_DEPS = $(CUSTOM_VLOG)
+else
+	RTL_DEPS = $(TOP_MODS_FILELIST) $(TOP_BB_MODS_FILELIST) $(TOP_SMEMS_FILE)
+endif
+
+$(VLSI_RTL): $(RTL_DEPS)
+ifneq ($(CUSTOM_VLOG), )
+	> $(VLSI_RTL)
```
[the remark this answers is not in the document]
harrisonliew: No. This basically writes a blank file. Equivalent to `echo "" > $(VLSI_RTL)`.

abejgonzalez (approved):

vlsi/Makefile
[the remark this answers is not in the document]
harrisonliew: Hm, turns out this breaks in csh-based shells (it should be `: > $(VLSI_RTL)`) but our scripts require bash anyway. Good to know for the future.

sagark (approved):
LGTM. this works for me as of b937a271c3341d272cc0d8ee6c7a9a1d8bf39884 (unfortunately don't have time to bump further right now).

I also pushed one small tweak to the Makefile -- I couldn't suggest change because that part of the Makefile was otherwise untouched. Feel free to revert if there's some problem with the change.

PR #1343, opened 2023-02-13
UPF Gen on new chipyard [ci skip]

sriramsridhar23:
In upf/ChipTopUPF.scala, under my default method I am having trouble figuring out what to check as the type for top. Before, there was a ChipTopLazyRawModuleImp class that worked since it had a ChipTop object as one of its fields, but in the most recent Chipyard that class has been removed. What alternative type could I check here to get the same access to a ChipTop object?

**Related PRs / Issues**:

**Type of change**:
- [ ] Bug fix
- [x] New feature
- [ ] Other enhancement

**Impact**:
- [ ] RTL change
- [ ] Build system change
- [x] Other

**Contributor Checklist**:


common.mk
```diff
@@ -108,6 +108,7 @@ $(FIRRTL_FILE) $(ANNO_FILE) &: $(SCALA_SOURCES) $(sim_files) $(SCALA_BUILDTOOL_D
 		--name $(long_name) \
 		--top-module $(MODEL_PACKAGE).$(MODEL) \
 		--legacy-configs $(CONFIG_PACKAGE):$(CONFIG) \
+		$(if chipyard.upf.ChipTopUPFAspect, --with-aspect chipyard.upf.ChipTopUPFAspect,) \
```
abejgonzalez, line 111: This `if` statement will always return as `true` resulting in the `--with-aspect` statement always being added. For now, I would just remove the `if` part of this code block.

generators/chipyard/src/main/scala/upf/ChipTopUPF.scala
```diff
@@ -0,0 +1,72 @@
...
+import chipyard.TestHarness
+import chipyard.{ChipTop, DigitalTop}
+import freechips.rocketchip.diplomacy.LazyModule
+import chisel.{UPFAspect, UPFFunction, ChiselUPFElement}
+
+import scala.collection.mutable.ListBuffer
+
+object ChipTopUPF {
+  def default: UPFFunction = {
+    case top: ChipTop => {
```
abejgonzalez, line 13: To do the proper matching, you should copy what they previously did: https://github.com/ucb-bar/chipyard/blob/a11263d1b17dbac65130aff41180404dbe712baf/generators/chipyard/src/main/scala/ChipTop.scala#L25

generators/chipyard/src/main/scala/upf/firrtl/Annotations.scala
```diff
@@ -0,0 +1,14 @@
...
+package chipyard.upf.firrtl
+
+import firrtl.annotations._
+import firrtl.stage.{RunFirrtlTransformAnnotation}
+
+trait UPFAnnotation extends Annotation {
+  val name: String
+}
+
+case class NameUPFAnnotation(target: Target, name: String) extends SingleTargetAnnotation[Target] with UPFAnnotation {
```
abejgonzalez, line 11: Since I think this Annotation is a noop, I think you can just make this a `NoTargetAnnotation`: https://github.com/chipsalliance/firrtl/blob/94d425f0f48e84bbae1be9d44d64615a37d960d8/src/main/scala/firrtl/annotations/Annotation.scala#L62

generators/chipyard/src/main/scala/upf/chisel/package.scala
```diff
@@ -0,0 +1,10 @@
+// See LICENSE for license details
+package chipyard.upf
+
+package object chisel {
```
abejgonzalez, line 4: Why is this a `package object chisel`? Why not just call this `package chipyard.upf.chisel`?

abejgonzalez: Also, I feel like this can be combined into the `UPFAspect.scala` file.

generators/chipyard/src/main/scala/upf/chisel/API.scala
```diff
@@ -0,0 +1,13 @@
+// See LICENSE for license details
+package chipyard.upf.chisel
+
+import firrtl.annotations.{Target, Annotation}
+import chipyard.upf.firrtl.NameUPFAnnotation
+
+abstract class ChiselNameUPFElement(scope: Target, name: String) {
+    def getAnnotations(): Seq[Annotation] = {
+        return Seq(NameUPFAnnotation(scope, name))
```
abejgonzalez, line 9: What happens if in the `getAnnotations` function you return an empty `Seq`? Will things still error out?

sriramsridhar23: Yup still errors out, I've instead made changes to limit the amount of info needed to create the annotations (ex. no parameters for NameUPFAnnotation).

PR #1348, opened 2023-02-16
Fix VLSI input files list emission to avoid bash "too many arguments" error

sagark:
This makes the expansion of "cat $(VLSI_RTL)" happen as a child of the shell that runs the for loop.

The existing version will sometimes produce a bash "too many arguments" error because the 

    $(shell cat $(VLSI_RTL)) 

is expanded first and then passed as a giant command to bash.

- [x] Bug fix
- [ ] New feature
- [ ] Other enhancement

**Impact**:
- [ ] RTL change
- [x] Build system change
- [ ] Other

**Contributor Checklist**:

### DIFF
--- a/vlsi/Makefile
+++ b/vlsi/Makefile
@@ -122,7 +122,7 @@ $(SYN_CONF): $(VLSI_RTL)
 	echo "synthesis.inputs:" >> $@
 	echo "  top_module: $(VLSI_TOP)" >> $@
 	echo "  input_files:" >> $@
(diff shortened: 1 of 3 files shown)


harrisonliew (changes requested):
Ah, is now a problem due to CIRCT, eh?
Can you also change this in `power.mk` and `sim.mk`?

harrisonliew (approved):

abejgonzalez (approved):

PR #1397, opened 2023-03-12
FIX (peripheral): add support for instantiating multiple peripherals of same type

T-K-233:
changelog:changed

Change how peripherals are included, so that user can instantiate multiple peripheral devices of same type in chip config.

**Related PRs / Issues**:

**Type of change**:
- [x] Bug fix
- [ ] New feature
- [ ] Other enhancement

**Impact**:
- [x] RTL change
- [ ] Build system change
- [ ] Other

**Contributor Checklist**:


generators/chipyard/src/main/scala/config/fragments/PeripheralFragments.scala
```diff
@@ -26,15 +28,20 @@ class WithBootROM extends Config((site, here, up) => {
...
+    GPIOParams(address = address, width = width, includeIOF = false))
 })
 // DOC include end: gpio config fragment
 
-class WithUART(baudrate: BigInt = 115200) extends Config((site, here, up) => {
+class WithUARTOverride(address: BigInt = 0x10020000, baudrate: BigInt = 115200) extends Config ((site, here, up) => {
   case PeripheryUARTKey => Seq(
-    UARTParams(address = 0x54000000L, nTxEntries = 256, nRxEntries = 256, initBaudRate = baudrate))
+    UARTParams(address = address, nTxEntries = 256, nRxEntries = 256, initBaudRate = baudrate))
+})
```
abejgonzalez, line 40: I think an API that makes more sense to me is: `class WithUARTOverride(no: Int, ...)` where the `no` is the index in the `PeripheryUARTKey` `Seq` to override.

abejgonzalez: Or if you want to truely override the entire key I would have two APIs:

- one called `WithUARTOverride` that takes a Seq of tuples (enforced to be size of 2 - where the 1st element in the tuple is the address and the 2nd is the baudrate) that gets unrolled into multiple UARTParams.
- another that is the additive uart (same as the current `WithUART`)

abejgonzalez (changes requested):

generators/chipyard/src/main/scala/config/fragments/PeripheralFragments.scala
jerryz123: I actually prefer the style here, as it matches how the WithNCores behaves

T-K-233: Ideally, we should get rid of the default UART (i.e. when user does not specify `WithUART`, we shouldn't include a UART in the SoC). 

I could not find where it adds the default UART though, so this is a hack to override the default parameter.

abejgonzalez: The default UART is given here: 

https://github.com/ucb-bar/chipyard/blob/44abc91cc147d77651d1d2c1b5585fcc35727ba3/generators/chipyard/src/main/scala/config/AbstractConfig.scala#L48

We have the CY config have a UART by default so that we can boot Linux easily in SW-RTL sims.

I actually think we can change the default implementation of `WithUART` here:

https://github.com/ucb-bar/chipyard/blob/44abc91cc147d77651d1d2c1b5585fcc35727ba3/generators/chipyard/src/main/scala/config/fragments/PeripheralFragments.scala#L35-L38

To essentially be the "override" config (i.e. generalize it to override all the `Seq[UARTParams]` given a `Seq` of tuples) and renaming it as such `WithUARTOverride`. Then we can keep the current implementation of `WithUART` seen here (where it's additive and like `WithN...Cores`) and just move it to the `PeripheralFragments.scala` file.

jerryz123: We want to keep the default UART. There's hardly a good reason to build a SoC without a UART....

The default UART config should be close to "universal".. that is, rarely should users need to change the address/params of that UART.

Additional WithUART flags can add more UARTs at other addresses to the system..


We should have an *additional* WithUARTOverride fragment for special cases when the default UART settings are not correct ... I don't think that fragment will be used very frequently, but it should be available at least.

T-K-233:
Revive this PR thread:
- Removed `WithUARTOverride` per requested.
- Added several other helpful peripheral config fragments, which are used in the tapeout class chip.
- Minor addition on ignoring the .vscode/ folder.

T-K-233:
Reopen the PR

PR #1472, opened 2023-05-10
Switch RTL sims to absolute clock-generators

jerryz123:
Which verilator v5, we can use absolute clock generators in all RTL simulators, which means we can ditch `DividerOnlyClockGenerator`.

- Default configs punch out all clocks to Harness
- Default RTL harness uses ClockWithFreq unsynthesizable verilog to emit clocks for each requested clock domain

**Related PRs / Issues**:

**Type of change**:
- [ ] Bug fix
- [ ] New feature
- [x] Other enhancement

**Impact**:
- [ ] RTL change
- [x] Build system change
- [ ] Other

**Contributor Checklist**:


generators/chipyard/src/main/scala/harness/HarnessClocks.scala
```diff
@@ -36,44 +37,26 @@ trait HarnessClockInstantiator {
...
-      _clockMap(sinkParams.name.get)._2.clock := divClock
-      _clockMap(sinkParams.name.get)._2.reset := divReset
-    }
-  }
+class ClockSourceAtFreqMHz(val freqMHz: Double) extends BlackBox(Map(
+  "PERIOD" -> DoubleParam(freqMHz)
+)) with HasBlackBoxInline {
+  val io = IO(new ClockSourceIO)
+
+  setInline("ClockSourceAtFreqMHz.v",
```
abejgonzalez, line 45: Nit: Can you get the module name in Chisel and set the file name and verilog module name to that? I think we should start to enforce that in inlined modules.

jerryz123: Sorry, I'm not quite sure I follow. I thought for black boxes the module name would be stuck at what the Chisel defines it as

abejgonzalez: I see that the filename on L45 and the module name at L47 are hardcoded to `ClockSourceAtFreqMHz`. Can you get the Scala class name w/ `this.getClass.getSimpleName` and put that in both the lines.

abejgonzalez: I also forget what the API in Chisel is to get the module name as well.

generators/chipyard/src/main/scala/harness/HarnessClocks.scala
```diff
@@ -36,44 +37,26 @@ trait HarnessClockInstantiator {
...
-      } else {
-        (refClock.clock, refClock.reset)
-      }
-
-      _clockMap(sinkParams.name.get)._2.clock := divClock
-      _clockMap(sinkParams.name.get)._2.reset := divReset
-    }
-  }
+class ClockSourceAtFreqMHz(val freqMHz: Double) extends BlackBox(Map(
+  "PERIOD" -> DoubleParam(freqMHz)
```
harrisonliew, line 41: The period is not the frequency in MHz...
Also, can the timescale be overriden in a way that breaks your desired tick time vs. the desired frequency?
Finally, nit: MHz instead of MHZ

harrisonliew (changes requested):

generators/chipyard/src/main/scala/harness/HarnessClocks.scala
jerryz123: Oops, fixed. Added `timeunit` as well.

I think this is already MHz, not MHZ?

harrisonliew: Thx. Are you certain that adding the timescale in the module body works for all simulators?

jerryz123: Yeah, this works on Verliator+VCS, now that verilator has `--timing` support. I haven't tested XCelium, but I would be surprised if that experienced issues with this.

harrisonliew (approved):

jerryz123:
I had to bump up the frequencies of the uncore to pass tests in time, previously the thing configured a 100MHz uncore, but ran it at 1 GHz with the TestDriver clock.

PR #1474, opened 2023-05-12
Unify supernode/harness-clocking across chipyard/firesim/fpga

jerryz123:
This PR unifies TestHarness/SuperNode/HarnessClocking across chipyard/firesim/and firechip. This saves 200 LOC, while maintaining existing capabilities. It also makes it easier to bring up a new FPGA target, you can see the reduction in LOC in the ArtyHarness.

The general principle here is that there are two categories of clockSinks. ChipTops can request clocks for any clock ports they expose, at any frequency. The user will also specify a single frequency HarnessBinderFrequency that be provided to all the HarnessBinder collateral as the implicit clock.

The Harness generates the set of requested frequencies.
 - RTL sims use ClockSourceAtFreq to generate the exact requested frequency
 - FireSim uses RationalClockBridge to approximate frequency ratios
 - FPGA prototypes currently tie all requested clocks to the referenceClock, but can be modified with a custom HarnessClockInstantiator to tie clocks to specific FPGA clock domais
 - Chips for tapeout should instantiate a PLL/clock-receivers, and request slow reference clock frequencies as IO from the harness

Changed
- `buildtopClock/Reset` -> `harnessBinderClock/Reset` : these are used to clock anything in a HarnessBinder
- `withClock(buildtopClock, buildtopReset)` no longer needed
- `WithClockGroupsCombinedByName` changed to allow overriding previously set clock groupings. This allows the `AbstractConfig` to specify a "default" set of clock groups across buses
- `HasHarnessSignalReferences` -> `HasHarnessInstantiators`, this trait now does almost all harness-level HarnessBinder/clocking construction, deduplicating stuff across chipyard/firechip/fpga
- Supernode stuff in firechip is now `MultiChip` in chipyard
- `ClockBridgeInstantiator` in firechip is refactored to fit chipyard's `HarnessClockInstantiator` API

Removed
- `DefaultClockFrequencyKey`. Now clock frequencies must be either requested explicitly, or a clock group with no specified frequency will take the clock of a group it is combined with. This gets rid of one more user clocking API that is difficult to reason about
- `TestChipMultiClockConfig`/`TestChipBusFreqs` , now `ChipLikeRocketConfig` demonstrates a tapeout-like design
- `WithFireSimSimpleClocks` is removed, it is the same as `WithPassthroughClockGenerator`
- `WithFireSimHarnessClockBinder` is removed, it is the same as `WithClockAndResetFromHarness`
- Ability to dynamically configure diplomatic lazy sifive fpga-shells based on the design. We want to move away from FPGA-shells for future boards anyways.
- This technically drops support for a synchronous-reset-pin on ChipTop, but that isn't a realistic design anyways.

**Related PRs / Issues**:

**Type of change**:
- [ ] Bug fix
- [ ] New feature
- [x] Other enhancement

**Impact**:
- [ ] RTL change
- [ ] Build system change
- [x] Other

**Contributor Checklist**:


T-K-233 (approved):
This is niiiiiiice!

generators/chipyard/src/main/scala/harness/HasHarnessInstantiators.scala
```diff
@@ -0,0 +1,96 @@
...
+import chipyard.harness.{ApplyHarnessBinders, HarnessBinders}
+import chipyard.iobinders.HasIOBinders
+import chipyard.clocking.{SimplePllConfiguration, ClockDividerN}
+import chipyard.{ChipTop}
+
+// -------------------------------
+// Chipyard Test Harness
+// -------------------------------
+
+case object MultiChipNChips extends Field[Int](0) // 0 means ignore MultiChipParams
```
abejgonzalez, line 21: Nit: Why not use an optional?

generators/chipyard/src/main/scala/harness/HasHarnessInstantiators.scala
```diff
@@ -0,0 +1,96 @@
...
+case class MultiChipParameters(chipId: Int) extends Field[Parameters]
+case object BuildTop extends Field[Parameters => LazyModule]((p: Parameters) => new ChipTop()(p))
+case object HarnessClockInstantiatorKey extends Field[() => HarnessClockInstantiator]()
+case object HarnessBinderClockFrequencyKey extends Field[Double](100.0) // MHz
+case object MultiChipIdx extends Field[Int](0)
+
+class WithMultiChip(id: Int, p: Parameters) extends Config((site, here, up) => {
+  case MultiChipParameters(`id`) => p
```
abejgonzalez, line 29: Can you give me a brief primer on how CDE keys can have parameters? Maybe I'm forgetting this exists but this seems new?

generators/chipyard/src/main/scala/harness/HasHarnessInstantiators.scala
```diff
@@ -0,0 +1,96 @@
...
+  // This shold be called last to build the ChipTops
+  def instantiateChipTops(): Seq[LazyModule] = {
+    val lazyDuts = chipParameters.zipWithIndex.map { case (q,i) =>
+      LazyModule(q(BuildTop)(q)).suggestName(s"chiptop$i")
+    }
+    val duts = lazyDuts.map(l => Module(l.module))
+
+    withClockAndReset (harnessBinderClock, harnessBinderReset) {
+      lazyDuts.zipWithIndex.foreach {
+        case (d: HasIOBinders, i: Int) => ApplyHarnessBinders(this, d.lazySystem, d.portMap)(chipParameters(i))
```
abejgonzalez, line 82: This PR doesn't include any fragments to wire up ChipTops to each other right?

generators/chipyard/src/main/scala/harness/HasHarnessInstantiators.scala
```diff
@@ -0,0 +1,96 @@
...
+  def success: Bool
+
+  // This can be accessed to get new clocks from the harness
+  val harnessClockInstantiator = p(HarnessClockInstantiatorKey)()
+
+  private val chipParameters = if (p(MultiChipNChips) == 0) {
+    Seq(p)
+  } else {
+    (0 until p(MultiChipNChips)).map { i => p(MultiChipParameters(i)).alterPartial {
+      case TargetDirKey => p(TargetDirKey) // hacky fix
```
abejgonzalez, line 68: Why is this getting dropped? I assume the `TargetDirKey` is set in the parameters are passed in / duplicated.

abejgonzalez (commented):
Overall looks super nice. Couple of questions for clarification.

abejgonzalez (commented):
Overall looks super nice. Couple of questions for clarification.

generators/chipyard/src/main/scala/clocking/ClockBinders.scala
```diff
@@ -111,6 +111,8 @@ class WithPassthroughClockGenerator extends OverrideLazyIOBinder({
       val (bundle, edge) = clockGroupAggNode.out(0)
 
       val clock_ios = (bundle.member.data zip edge.sink.members).map { case (b, m) =>
+        require(m.take.isDefined, s"""Clock ${m.name.get} has no requested frequency
+                                     |Clocks: ${edge.sink.members.map(_.name.get)}""".stripMargin)
```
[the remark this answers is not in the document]
jerryz123: This is the `stripMargin` feature of scala strings, to make multi-line strings look nicer.

generators/chipyard/src/main/scala/harness/HasHarnessInstantiators.scala
jerryz123: Next PR

fpga/src/main/scala/arty100t/Configs.scala
```diff
@@ -26,6 +26,12 @@ class WithArty100TTweaks extends Config(
   new WithArty100TUARTTSI ++
   new WithArty100TDDRTL ++
   new WithNoDesignKey ++
+  new chipyard.harness.WithHarnessBinderClockFreqMHz(50) ++
+  new chipyard.config.WithMemoryBusFrequency(50.0) ++
+  new chipyard.config.WithSystemBusFrequency(50.0) ++
+  new chipyard.config.WithPeripheryBusFrequency(50.0) ++
```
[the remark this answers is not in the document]
jerryz123: > For FPGA prototypes, can you do multi-clock designs? Or are you enforced to have a single frequency across all clock domains? If that is the case, is there some require/assert to enforce this?

The existing FPGA designs are all single clock. To support multi-clock FPGA, a `HarnessClockInstantiator` variant would have to be designed that would configure the clock generator IP for that particular board. That is future work for the next use case which wants multi-clock-target FPGA prototypes.

The broadcast-single-clock approach is the simpler thing to do, and hopefully it can make it more clear how one might bring up a new board. The `AllClocksFromHarnessClockInstantiator` will check/enforce that all requested clocks match the reference clock frequency.

> Why should the harness clock have a specific value? Can't you just choose the default fastest clock of the system and use that?

Its not clear to me that using the fastest clock in the system is the most obvious behavior. On a realistic system, where the tile runs faster than the uncore, this doesn't really make sense.

Its also not clear to me that using the slowest clock is the most intuitive behvior either.

I feel like the most intuitive, and realistic, behavior, is for harness clocks to be completely async/unrelated with chip clocks. 
A more aggressive refactor of the HarnessBinders could remove the `harnessBinderClock` entirely, and force all HarnessBinders to request their own clock at whatever frequency they need, instead of relying on some implicit clock.

generators/chipyard/src/main/scala/config/RocketConfigs.scala
```diff
@@ -90,24 +90,20 @@ class MulticlockRocketConfig extends Config(
...
   // Frequency specifications
   new chipyard.config.WithTileFrequency(1600.0) ++       // Matches the maximum frequency of U540
-  new chipyard.config.WithSystemBusFrequency(800.0) ++   // Ditto
+  new chipyard.clocking.WithClockGroupsCombinedByName(("uncore"   , Seq("sbus", "cbus", "implicit")),
+                                                      ("periphery", Seq("pbus", "fbus"))) ++
+  new chipyard.config.WithSystemBusFrequency(800.0) ++   // Matches the maximum frequency of U540
   new chipyard.config.WithMemoryBusFrequency(1000.0) ++  // 2x the U540 freq (appropriate for a 128b Mbus)
-  new chipyard.config.WithPeripheryBusFrequency(100) ++  // Retains the default pbus frequency
-  new chipyard.config.WithSystemBusFrequencyAsDefault ++ // All unspecified clock frequencies, notably the implicit clock, will use the sbus freq (800 MHz)
+  new chipyard.config.WithPeripheryBusFrequency(100) ++  // Slow periphery bus
```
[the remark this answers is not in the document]
jerryz123: Skimming that, I see "The recommended frequency of coreclk is 1.0 GHz" and "The L2 cache and peripherals such as UART, SPI, I2C, and PWM operate in a single clock domain (tlclk) running at coreclk/2 rate.". This can be updated to match that.

fpga/src/main/scala/vcu118/TestHarness.scala
```diff
@@ -80,19 +80,18 @@ class VCU118FPGATestHarness(override implicit val p: Parameters) extends VCU118S
...
   // connect 1 mem. channel to the FPGA DDR
-  val inParams = topDesign match { case td: ChipTop =>
-    td.lazySystem match { case lsys: CanHaveMasterTLMemPort =>
-      lsys.memTLNode.edges.in(0)
-    }
-  }
-  val ddrClient = TLClientNode(Seq(inParams.master))
+  val ddrClient = TLClientNode(Seq(TLMasterPortParameters.v1(Seq(TLMasterParameters.v1(
+    name = "chip_ddr",
+    sourceId = IdRange(0, 64)
```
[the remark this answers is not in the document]
jerryz123: Unfortunately, this PR drops the ability to configure the FPGA shell based on the target design.
I think the approach moving forwards should be that the "shell" or "harness" contains a superset of IO that the ChipTop might want, and requires should check that the requested design can fit under that IO.

I think I determined these params using printlns. I'll add some asserts to check that the requested design can work with this DDR.

generators/firechip/src/main/scala/TargetConfigs.scala
```diff
@@ -63,9 +69,11 @@ class WithNVDLASmall extends nvidia.blocks.dla.WithNVDLA("small")
 
 // Minimal set of FireSim-related design tweaks - notably discludes FASED, TraceIO, and the BlockDevice
 class WithMinimalFireSimDesignTweaks extends Config(
-  // Required*: Uses FireSim ClockBridge and PeekPokeBridge to drive the system with a single clock/reset
-  new WithFireSimHarnessClockBinder ++
-  new WithFireSimSimpleClocks ++
+  // Required*: Punch all clocks to FireSim's harness clock instantiator
+  new WithFireSimHarnessClockBridgeInstantiator ++
+  new chipyard.harness.WithHarnessBinderClockFreqMHz(1000.0) ++
```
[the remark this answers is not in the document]
jerryz123: The old way, which used SystemBusFreq as the "default" freq, configured the harness clock to 1 GHz.

generators/chipyard/src/main/scala/harness/HasHarnessInstantiators.scala
jerryz123: CDE have always supported parameters, but they are very rarely used.
To make this work, you have to define the key as a `case class SomeKey(someParams: ParamsType) extends Field[...]` instead of `case object SomeKey extends Field[...]`.

The only other trick is when matching on it, if you need to match on a variable defined outside the case expression, you have to put the variable in backticks... ``case SomeKey(`params`)`` instead of `case SomeKey(params)`

generators/chipyard/src/main/scala/harness/HasHarnessInstantiators.scala
jerryz123: The generate phase adds the `TargetDirKey` to the `Parameters` object before it is passed into this generator., but not in the nested `Parameters` within `MultiChipParameters`. 

This propagates `TargetDirKey` down to the nested parametesr in `MultiChipParameters`.

fpga/src/main/scala/vcu118/TestHarness.scala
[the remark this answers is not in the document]
jerryz123: Nvm I changed it so the old behavior is preserved.

generators/chipyard/src/main/scala/harness/HasHarnessInstantiators.scala
jerryz123: Fixed

generators/chipyard/src/main/resources/csrc/cospike.cc
```diff
@@ -139,7 +139,7 @@ extern "C" void cospike_cosim(long long int cycle,
         in_permissive = true;
       } else if (arg == "+permissive-off") {
         in_permissive = false;
-      } else if (arg == "+cospike_debug") {
+      } else if (arg == "+cospike_debug" || arg == "+cospike-debug") {
```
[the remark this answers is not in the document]
jerryz123: Intentional, was a random spot fix I found while using this to debug

harrisonliew:
This is nice! One question before I approve:

> RTL sims use ClockSourceAtFreq to generate the exact requested frequency
FireSim uses RationalClockBridge to approximate frequency ratios
FPGA prototypes tie all requested clocks to the referenceClock

Let's say I am making a chip with multiple clocks, some of which are intended to be high-speed analog clocks. For simulation, this makes sense. But is a FPGA harness required to also generate those otherwise analog clocks and tie them to the reference clock? Or is there a mechanism for denoting that some clocks only apply to simulation but not for FPGA harnesses?

jerryz123:
If you are generating a design with high-speed analog clocks, what would you even want when putting the design on a FPGA? The analog blocks wouldn't be synthesizable, so you'd have to cut those out anyways.

The FPGA prototype flows can be extended in the future to support multi-clock configurations under whatever constraints the FPGA PLLs and clock resources can support, this would involve only specifying a different HarnessClockInstantiator that ties chiptop clock domains to specific FPGA clock domains.

harrisonliew:
> The analog blocks wouldn't be synthesizable, so you'd have to cut those out anyways.

Not true, we should be able to use behavioral models (or eventually, mixed-signal simulation). Anyways, if we are envisioning being able to have those additional FPGA clock domains, this should be trivial to support.

harrisonliew (approved):

jerryz123:
Right, we should support multiclock prototypes with synthesizable models of analog blocks.

.github/scripts/run-tests.sh
```diff
@@ -110,7 +110,8 @@ case $1 in
         make run-binary-fast -C $LOCAL_SIM_DIR $DISABLE_SIM_PREREQ ${mapping[$1]} BINARY=$RISCV/riscv64-unknown-elf/share/riscv-tests/benchmarks/multiply.riscv
         ;;
     chipyard-ibex)
-        run_bmark ${mapping[$1]} #TODO: Find 32-bit test
+        # Ibex cannot run the riscv-tests binaries for some reason
+        # make run-binary-fast -C $LOCAL_SIM_DIR $DISABLE_SIM_PREREQ ${mapping[$1]} BINARY=$RISCV/riscv64-unknown-elf/share/riscv-tests/isa/rv32ui-p-simple
```
abejgonzalez: Do we have any easy way to compile a 32b test? If not, maybe we should just check that Ibex compiles (and not runs)

abejgonzalez (approved):
LGTM (lets hold off on merging this until mid-next week)

.github/scripts/run-tests.sh
jerryz123: Not at the moment, the default libgloss install is only rv64.

Ibex seems to fail this due to not implementing some necessary CSR :man_shrugging:

PR #1476, opened 2023-05-13
Support uni-directional TLSerdesser

jerryz123:
This change fixes support for configs in which the TLSerdesser is unidirectional; that is, cases in which the chip acts as only the manager, and off chip acts as only the client.

This simplifies the memory map for all default designs as we no longer have to add a stub off-chip memory to provide the chip-client/harness-manager link.

Why not use the existing `TLDesser/TLSerder`, which already do this? Those IMO can be deleted as their implementations are now supported by the generalized `TLSerdesser`.

Since we support configs now where the serializer only acts as a client to the FBus, a bunch of `WithSerialPBusMem` fragments can be removed, as these were used to specify where the stub tilelink device sat.

**Related PRs / Issues**:

**Type of change**:
- [ ] Bug fix
- [ ] New feature
- [x] Other enhancement

**Impact**:
- [x] RTL change
- [ ] Build system change
- [ ] Other

**Contributor Checklist**:

### DIFF
--- a/.github/scripts/run-tests.sh
+++ b/.github/scripts/run-tests.sh
@@ -110,7 +110,8 @@ case $1 in
         make run-binary-fast -C $LOCAL_SIM_DIR $DISABLE_SIM_PREREQ ${mapping[$1]} BINARY=$RISCV/riscv64-unknown-elf/share/riscv-tests/benchmarks/multiply.riscv
         ;;
     chipyard-ibex)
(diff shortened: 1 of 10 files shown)


T-K-233 (approved):
Looks good!

It would be better if we name "serialManager*" as "serialTLManager*", but I think this is also fine.

jerryz123:
> Looks good!
> 
> 
> 
> It would be better if we name "serialManager*" as "serialTLManager*", but I think this is also fine.

That a good idea. I'll make that change

abejgonzalez (approved):

PR #1486, opened 2023-05-26
Add higher level explanations of RoCC + more resources

nikhiljha:
Small PR based on feedback from EE194 (tapeout) to explain what RoCC is at a higher level, and some resources for actually using the RoCC interface.

**Type of change**:
- [X] Other enhancement (Documentation)

**Impact**:
- [X] Other (Documentation)

**Contributor Checklist**:
- [X] Did you set `main` as the base branch?
- [X] Is this PR's title suitable for inclusion in the changelog and have you added a `changelog:<topic>` label?
- [X] Did you state the type-of-change/impact?
- [X] Did you delete any extraneous prints/debugging code?
- [X] Did you mark the PR with a `changelog:` label?

### DIFF
--- a/docs/Customization/RoCC-Accelerators.rst
+++ b/docs/Customization/RoCC-Accelerators.rst
@@ -88,7 +88,8 @@ In your ``LazyRoCCModuleImp``, the signal ``io.mem`` is a ``HellaCacheIO``, whic
 
 At a high level, you must tag requests that you send across this interface using the ``io.mem.req.tag``, and the tag will be returned to you when the data is ready.
 Responses may come back out of order if you issue multiple requests, so you can use these tags to tell what data came back.
-Note that the top two bits of the tag are reserved, and MUST be set to zero, or the interface will exhibit undefined behavior.
+Note that the number of tag bits is controled by ``dcacheReqTagBits``, which is usually set to 6.
+Using more than 6 bits will cause errors or hangs.
 
 
 Adding RoCC accelerator to Config
--- a/docs/Customization/RoCC-Accelerators.rst
+++ b/docs/Customization/RoCC-Accelerators.rst
@@ -1,10 +1,15 @@
 .. _rocc-accelerators:
 
 Adding a RoCC Accelerator
-----------------------------
+-------------------------
 
-RoCC accelerators are lazy modules that extend the ``LazyRoCC`` class.
-Their implementation should extends the ``LazyRoCCModule`` class.
+A RoCC accelerator is a component that can be added into a particular Rocket or BooM tile.
+It receives instructions that match a certain opcode, talks to other parts of the core or SoC (L1, L2, PTW, FPU), and then optionally writes back a value into the register corresponding with the ``rd`` field of the instruction.
+RoCC accelerators are instantiated via modules that extend the ``LazyRoCC`` class.
+These modules lazily instantiate another module which extends the ``LazyRoCCModule`` class.
+This extra layer of indirection is used so that Diplomacy can figure out how to connect the RoCC module to the chip, without needing to instantiate the module ahead of time.
+Lazy modules are further explained in the :ref:`Chipyard-Basics/Configs-Parameters-Mixins:Cake Pattern / Mixin` section.
+Below is a minimal instantiation of a RoCC accelerator.
 
 .. code-block:: scala
 
@@ -31,7 +36,6 @@ Their implementation should extends the ``LazyRoCCModule`` class.
       ...
     }
 
-
 The ``opcodes`` parameter for ``LazyRoCC`` is the set of custom opcodes that will map to this accelerator.
 More on this in the next subsection.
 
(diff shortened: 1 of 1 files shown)


docs/Customization/RoCC-Accelerators.rst
```diff
@@ -46,6 +50,47 @@ the ``busy`` signal, which indicates when the accelerator is still handling an i
...
+
+
+Accessing Memory via L1 Cache
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A RoCC accelerator can access memory through the L1 Cache of the core it is attached to.
+This is typically a good idea, as a RoCC instruction that passes in an address likely has accessed it recently.
+
+In your ``LazyRoCCModuleImp``, the signal ``io.mem`` is a ``HellaCacheIO``, which is defined in ``generators/rocket-chip/src/main/scala/rocket/HellaCache.scala``.
+If your core is a RocketCore, the other end of the input is connected to a ``SimpleHellaCacheIF``, which has a replay buffer in case the L1 cache is overloaded and doesn't ACK your request.
```
jerryz123, line 63: Can you omit mentioning the SimpleHellaCacheIF, that is an implementation detail. You should emphasize the point that only the `req` and `resp` signals matter for the accelerator implementation, the rest can be ignored.

docs/Customization/RoCC-Accelerators.rst
```diff
@@ -46,6 +50,47 @@ the ``busy`` signal, which indicates when the accelerator is still handling an i
...
+        val ordered = Input(Bool())
+        val perf = Input(new HellaCachePerfEvents())
+
+        val keep_clock_enabled = Output(Bool()) // should D$ avoid clock-gating itself?
+        val clock_enabled = Input(Bool()) // is D$ currently being clocked?
+    }
+
+At a high level, you must tag requests that you send across this interface using the ``io.mem.req.tag``, and the tag will be returned to you when the data is ready.
+Responses may come back out of order if you issue multiple requests, so you can use these tags to tell what data came back.
+Note that the top two bits of the tag are reserved, and MUST be set to zero, or the interface will exhibit undefined behavior.
```
jerryz123, line 92: Huh. I think that might actually be a bug somewhere.

docs/Customization/RoCC-Accelerators.rst
```diff
@@ -46,6 +50,47 @@ the ``busy`` signal, which indicates when the accelerator is still handling an i
...
 
 Look at the examples in ``generators/rocket-chip/src/main/scala/tile/LazyRoCC.scala`` for detailed information on the different IOs.
+There is also more information about each of the signals in `the RoCC Documentation written by UCSD <https://docs.google.com/document/d/1CH2ep4YcL_ojsa3BVHEW-uwcKh1FlFTjH_kg5v8bxVw/edit>`_, although it is updated out of tree and may be out of date.
+
+
+Accessing Memory via L1 Cache
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A RoCC accelerator can access memory through the L1 Cache of the core it is attached to.
+This is typically a good idea, as a RoCC instruction that passes in an address likely has accessed it recently.
```
jerryz123, line 60: Can you phrase this as "This is a simpler interface for accelerator architects to implement, but will generally have lower achievable throughput than a dedicated TileLink port"

jerryz123 (changes requested):

docs/Customization/RoCC-Accelerators.rst
nikhiljha: It's been a minute but from what I recall:

What happens is SimpleHellaCacheIF puts the entire tag into its replay queue, but the actual memory IF drops the top two bits... so when the data returns the tag is missing the top two bits and SimpleHellaCacheIF won't remove it from the nack queue (since it sees that it got a message with a different tag). This leads to the nack queue filling up and simplehellacacheif permanently asserting not ready.

If you look at the two rocc accelerators in chipyard they both set the top two bits to zero presumably for this reason. That's why I mentioned SimpleHellaCacheIF-- even if it should be an implementation detail you may need to inspect it to have working code 🥲

jerryz123: What is the actual number of tag bits available to use? Is it 6?

nikhiljha: Looks like it's 

> (coreParams.dcacheReqTagBits + log2Ceil(dcacheArbPorts)

so maybe the top log2Ceil(dcacheArbPorts) bits of the tag are supposed to be the port that the request came in from and that's why they need to be zeroed(?)... I'll try to figure this out again 🤔

jerryz123: I think it might just be dcachereqtagbits

jerryz123: I updated this so we can get these docs in. Let me know if this still seems unclear.

jerryz123 (approved):

PR #1506, opened 2023-06-09
Discussion around coding style + chisel module templates

abejgonzalez:
This is to jumpstart discussion on how to make Chisel modules easier to understand / faster ramp up. Feel free to comment on any part of this PR with your thoughts/recommendations.

**Related PRs / Issues**:

**Type of change**:
- [ ] Bug fix
- [ ] New feature
- [ ] Other enhancement

**Impact**:
- [ ] RTL change
- [ ] Build system change
- [ ] Other

**Contributor Checklist**:

### DIFF
--- a/generators/chipyard-helpers/src/main/scala/chipyard-helpers/CodingStyle.scala
+++ b/generators/chipyard-helpers/src/main/scala/chipyard-helpers/CodingStyle.scala
@@ -1,0 +1,53 @@
+class ExampleModule extends Module {
+  // val chisel stuff should be in snake case
+  val temp_wire = WireDefault(...)
+  val my_cool_reg = RegInit(...)
+
+  // scala global defaults should be all caps
+  val DEFAULT_VALUE = 20
+
+  // other scala vars should be in camelcase
+  val myOtherVar = 20
+
+  // scala functions that are associated with the module should be put in an obj
+  // notice how the function is also in camel case since it gives a scala value out
+  def myFoo(intBits: Int, chisel_val: UInt): Int = {
+
+  }
+
+  // notice how the function is also in snake case since it gives a chisel value out
+  def my_bar(intBits: Int, intBits2: Int): UInt = {
+
+  }
+}
+
+class ModuleTemplate {
+  // override this function to place all ios in a known location
+  // must be scaladoc commented
+  def ios: ??? = ???
+
+  // override this function to place all chisel code in a known location
+  // must be scaladoc commented
+  def implementation: ??? = ???
+}
+
+class ChiselModuleStart extends ModuleTemplate {
+  def multiplyScalaNumbers(inputNumber0: Int, inputNumber1: Int): Int = {
+    inputNumber0 * inputNumber1
+  }
+
+  def add_two_chisel_uints(in_0: UInt, in_1: UInt): UInt = {
+    in_0 + in_1
+  }
+
+  override def ios: ??? = {
+    val io = IO(new Bundle {
+      val my_io_here = Input(UInt())
+      val my_io_again_here = Output(UInt())
+    })
+  }
+
+  override def implementation: ??? = {
+    io.my_io_again_here := add_two_chisel_uints(io.my_io_here, multiplyScalaNumbers(10, 2).U)
+  }
+}
--- a/build.sbt
+++ b/build.sbt
@@ -198,6 +198,12 @@ lazy val boom = (project in file("generators/boom"))
   .settings(libraryDependencies ++= rocketLibDeps.value)
   .settings(commonSettings)
 
+// add to whatever sbt project wants to use the helper
+lazy val chipyard_helpers = (project in file("generators/chipyard_helpers"))
+  .dependsOn(rocketchip)
+  .settings(libraryDependencies ++= rocketLibDeps.value)
+  .settings(commonSettings)
+
 lazy val cva6 = (project in file("generators/cva6"))
   .dependsOn(rocketchip)
   .settings(libraryDependencies ++= rocketLibDeps.value)


joshuaminwookang:
@jerryz123 @sagark @kevindna @joey0320

generators/chipyard-helpers/src/main/scala/chipyard-helpers/CodingStyle.scala
```diff
@@ -0,0 +1,53 @@
+class ExampleModule extends Module {
```
jerryz123, line 1: This ought to just be in the examples, and directly referenced in the docs as a style guide...I don't think there's any value in putting this in a separate package

generators/chipyard-helpers/src/main/scala/chipyard-helpers/CodingStyle.scala
```diff
@@ -0,0 +1,53 @@
...
+  val myOtherVar = 20
+
+  // scala functions that are associated with the module should be put in an obj
+  // notice how the function is also in camel case since it gives a scala value out
+  def myFoo(intBits: Int, chisel_val: UInt): Int = {
+
+  }
+
+  // notice how the function is also in snake case since it gives a chisel value out
+  def my_bar(intBits: Int, intBits2: Int): UInt = {
```
jerryz123, line 19: Functions should be camelCase always, only chisel hardware variables get snake_case

generators/chipyard-helpers/src/main/scala/chipyard-helpers/CodingStyle.scala
```diff
@@ -0,0 +1,53 @@
...
+
+  }
+}
+
+class ModuleTemplate {
+  // override this function to place all ios in a known location
```
jerryz123, line 25: Can you provide an example of a module which follows this template?

generators/chipyard-helpers/src/main/scala/chipyard-helpers/CodingStyle.scala
```diff
@@ -0,0 +1,53 @@
...
+class ChiselModuleStart extends ModuleTemplate {
+  def multiplyScalaNumbers(inputNumber0: Int, inputNumber1: Int): Int = {
+    inputNumber0 * inputNumber1
+  }
+
+  def add_two_chisel_uints(in_0: UInt, in_1: UInt): UInt = {
+    in_0 + in_1
+  }
+
+  override def ios: ??? = {
```
jerryz123, line 43: It seems like this is just the more restrictive old-chisel IO system

```
abstract class SimpleModule extends Module {
  val io: IO
}
```

generators/chipyard-helpers/src/main/scala/chipyard-helpers/CodingStyle.scala
```diff
@@ -0,0 +1,53 @@
...
+    val io = IO(new Bundle {
+      val my_io_here = Input(UInt())
+      val my_io_again_here = Output(UInt())
+    })
+  }
+
+  override def implementation: ??? = {
```
jerryz123, line 50: I feel like this will just cause more confusion then it saves. some questions I foresee:

 - When should I use Module instead of ModuleTemplate?
 - Why does ModuleTemplate force this implementation block, while in the normal modules I can just put it in the class body?
 - What's the difference between a Module, a LazyModule, a LazyModuleImp, and a ModuleTemplate?
 - How do I mix ModuleTemplates and Modules?

kevindna: Comments or an accompanying document to brief explain the uses? At least the Lazymodule stuff. Don't know anything about ModuleTemplate

generators/chipyard-helpers/src/main/scala/chipyard-helpers/CodingStyle.scala
abejgonzalez: Sure. I just put things here so it was more separated. We can choose where this lives later.

generators/chipyard-helpers/src/main/scala/chipyard-helpers/CodingStyle.scala
abejgonzalez: I originally did what you said... but I do think there is some value in having the functions returning Chisel objects be in snake_case. It makes it clear to end users what is returning a Chisel type or Scala type (strictly based on the snake_case or camelCase of the function name).

generators/chipyard-helpers/src/main/scala/chipyard-helpers/CodingStyle.scala
abejgonzalez: The point is to be more restrictive such that people can parse modules easier.

generators/chipyard-helpers/src/main/scala/chipyard-helpers/CodingStyle.scala
abejgonzalez: This clearly doesn't compile... but the point is something like the module below would 'implement' the template by defining the def's.

generators/chipyard-helpers/src/main/scala/chipyard-helpers/CodingStyle.scala
jerryz123: This contradicts existing helper functions everywhere in chisel, chisel.util, and rocketchip. 

Ex: PopCount, RegNext, OHToUInt, PriorityEncoder

generators/chipyard-helpers/src/main/scala/chipyard-helpers/CodingStyle.scala
jerryz123: I think we'd need to see an example of a compilable template, the syntax of the template would need to be intuitive and match how scoping in Scala and Chisel already behave.

generators/chipyard-helpers/src/main/scala/chipyard-helpers/CodingStyle.scala
abejgonzalez: > Comments or an accompanying document to brief explain the uses? At least the Lazymodule stuff. Don't know anything about ModuleTemplate

This type of template is not supposed to help people to understand `LazyModule` or `LazyModuleImp`'s. That we need to discuss more since that's significantly harder to explain concisely :). This work is attempting to do 2 things:

1. Have a coding style guide that others can reference when making new modules
2. Have a more restrictive form of a normal Chisel `Module` that (softly) forces users to organize code in a particular way (IOs in one place, the rest of the module in another place) so it is easier to read. I argue that if this isn't in a "template" form, users will just ignore this file and never adhere to it (unless someone wants to write a Chisel + Scala linter - which I think is the ultimate/best solution).

> I feel like this will just cause more confusion then it saves. some questions I foresee:
> 
> * When should I use Module instead of ModuleTemplate?

Yea, I think the optimal thing would be to somehow make this a trait that gets mixed in and forces an organization... but I don't know how that would work / if it could work.

> * Why does ModuleTemplate force this implementation block, while in the normal modules I can just put it in the class body?

Yea, I don't have a good response to this other than us writing good docs saying it's a very restrictive form of creating a module (for readability)

> * What's the difference between a Module, a LazyModule, a LazyModuleImp, and a ModuleTemplate?

We need to build up a story for the Lazy*'s and making them easier to understand.

> * How do I mix ModuleTemplates and Modules?

With the right documentation... you can say that they are interchangeable.

generators/chipyard-helpers/src/main/scala/chipyard-helpers/CodingStyle.scala
jerryz123: If we want this more restrictive module to be used more commonly, then adding it to Chisel would be necessary for adoption. I think this is worth bringing up to the chisel devs at least.

generators/chipyard-helpers/src/main/scala/chipyard-helpers/CodingStyle.scala
jerryz123: My point is that this would introduce yet-another abstraction to module construction that would need to be explained, so we'd need to tread carefully here.
To help understand the problem more, can you provide examples of existing modules that would benefit from a more restrictive module format? (Benefit as in, improve readability)

generators/chipyard-helpers/src/main/scala/chipyard-helpers/CodingStyle.scala
abejgonzalez: I don't want to change the world... so I'm fine not doing this.

abejgonzalez:
Lukewarm take: It should be pretty easy to write a Scala/Chisel linter for the coding style guide (camelCase/snake_case).

See the tutorial below:
https://scalacenter.github.io/scalafix/docs/developers/setup.html

FireSim already uses ScalaFix so it should already be easily extensible/useable in Chipyard

jerryz123:
I think a linter for variable naming would be very useful, and definitely worth pursuing.

joey0320:
I also agree that the coding style should be dictated by a linter rather than a module template.

joey0320:
https://scalameta.org/scalafmt/docs/installation.html
There also seems to be a scala formatter that is compatible w/ sbt

abejgonzalez:
> https://scalameta.org/scalafmt/docs/installation.html There also seems to be a scala formatter that is compatible w/ sbt

Looking at ScalaFmt vs *Fix, it seems like *Fix is extensible whereas *Fmt isn't. With that being said, they both seem to basically be the same thing (minus the extensibility)

generators/chipyard-helpers/src/main/scala/chipyard-helpers/CodingStyle.scala
abejgonzalez: > If we want this more restrictive module to be used more commonly, then adding it to Chisel would be necessary for adoption. I think this is worth bringing up to the chisel devs at least.

I think for right now if we decide to go in this direction and if this is used internally and if this is deemed useful we can upstream it to Chisel. Right now this is so early on that I think we should wait on this. (Ditto for getting feedback, we need to establish the details a bit more before reaching out)

abejgonzalez: > I also feel like this is a bit restrictive. What if the user wants to mix in IOs as a trait (although I think that isn't necessarily a good idea...)?

If you are mixing in IOs (doing other mixins) then this wouldn't be for you. I think of this as "training wheels" to write your Chisel module.

joey0320: Hmm so are we kind of letting the user choose either the template version vs the current way to define modules?
I guess it would be more intuitive for incoming users.

abejgonzalez: Yea, that's the intention behind this... whether this will actually help is still yet to be decided.

abejgonzalez:
Also https://github.com/Intensivate/learning-journey

generators/chipyard-helpers/src/main/scala/chipyard-helpers/CodingStyle.scala
abejgonzalez: Points where I think this type of construct might help:

"Where are the IOs of this module?" - 
https://github.com/chipsalliance/rocket-chip/blob/47f7b7144727f0340d511d35b9f6c7a91b2a276f/src/main/scala/rocket/RocketCore.scala#L122

"What does it mean to have multiple IOs?" -
https://github.com/OpenXiangShan/XiangShan/blob/3eeae490776c53712da27da9143d26c0c8238996/src/main/scala/top/Top.scala#L116

"Are these functions utilities specific to the module or can they be useful elsewhere?" -
https://github.com/chipsalliance/rocket-chip/blob/47f7b7144727f0340d511d35b9f6c7a91b2a276f/src/main/scala/rocket/TLB.scala#L733

I think there is more... but stuff like this seems like a starting point. I think enforcing functions makes the boundaries clear of what should be put where (instead of allowing users to potentially put IOs anywhere, have multiple IO definitions, functions to be everywhere, etc)

Issue #1046, opened 2021-11-15
Which .specs file is suitable for development of baremetal systems with JTAG interface ?


johnny-wang16:
I'm trying to set up a DUT that's suitable for development of baremetal systems. The path I'm trying is the JTAG interface according to [this part of chipyard documentation](https://chipyard.readthedocs.io/en/latest/Advanced-Concepts/Chip-Communication.html). However, the ` .specs` files under `chipyard/esp-tools-install/riscv64-unknown-elf/lib/` does not seem to be suitable for this use case (ex: `htif_nano/nano.specs` assumes an OS exist which isn't true for baremetal systems. Some specs file includes `libc` and some redundant libraries for dumping functionality which makes the code size unnecessarily larger. ) So, I'm wondering if there's a minimal `specs `file  that's compatible with JTAG but without including redundant libraries. Or, if there are some documentations on how to modify the `specs `files. Thanks.

Issue #194, opened 2019-08-08
Distributing tests folder


abejgonzalez:
Should we distribute the `tests` folder so things like `blkdev.c` and `.h` reside in the resources of the blockdev device (similar for `nic.h`).

zhemao:
The problem is that there's a bunch of template code needed for these baremetal tests. I don't think it makes sense to separate everything out. We may want to create a `chipyard-tests` or `peripheral-tests` submodule though.

PR #1484, opened 2023-05-26
Provide example of tethered-config simulation with MultiHarnessBinders

jerryz123:
This adds a `BringupHostConfig` to go alongside the `ChipLikeRocketConfig`. The `BringupHostConfig` is the FPGA design that can bring-up the `ChipLikeRocketConfig`. 

This also adds the `TetheredChipLikeRocketConfig`, which uses the `MultiChip`/`MultiHarnessBinder` APIs to build `ChipLikeRocketConfig` as `chiptop0`, `BringupHostConfig` as `chiptop1`, connect their serial-tl ports, connect SimTSI to the `BringupHostConfig`'s UART-TSI pot, and simulate the entire system. The entire system can be simulated in RTL simulation, and tested with the standard `testchip_tsi` plusarg/htif commands.

Ex: 
```
make CONFIG=TetheredChipLikeRocketConfig EXTRA_SIM_FLAGS="+cflush=0x2010200" run-binary LOADMEM=1 BINARY=hello.riscv
```
This approach to simulation will be very slow, as it will simulate both the slow TSI-over-UART and serial-tilelink interfaces at accurate frequencies.

The procedure for those using Chipyard to tape-out should be to craft an equivalent `ChipConfig` and `BringUpHostConfig` specific to their design, connect them in a `MultiChip` harness configuration using the `MultiHarnessBinder` API to model the chip-to-FPGA link, and simulate the system entirely in RTL simulation before tapeout. 

Added:
- `MultiHarnessBinder` - like HarnessBinders, but for chiptop-to-chiptop connections
- `CanHavePeripheryUARTTSI` - adds a UART port that transports TSI, which gets converted to a tilelink master. This should only ever be built as part of FPGA bringup platforms.
- `OBUS` - "OffchipBus" - devices which transport tilelink to an off-chip tilelink network (ex: FPGA bringup board) should hang off this. For a chip, the OBUS contains all the off-chip memory. For a bringup FPGA config, the OBUS contains the chip's on-chip memory.
- `BringupHostConfig` - a no-core config that just contains some buses to support the `ChipConfig`'s off-chip memory, and bridging between UART-TSI and the test chip.

Changes:
- Clean up the run-tests CI script to dedup a lot of the commands
- Reorganize AbstractConfig default IOBinders to delineate between IOCell and punchthrough IOs
- SerialTL hangs off the OBUS instead of MBUS now (OBUS can hang off the MBUS).

Fixed:
- `run-binary-debug BINARY=none` should not attempt to objdump none

[...]

PR #1534, opened 2023-06-27
Add Baremetal IDE support

T-K-233:
This PR adds an integrated environment for compiling baremetal code for both simulation and post-silicon chips.

Different from the simple test code from `/tests` folder, this environment also includes the complete initialization procedure of the memory layout, global, stack, and frame pointers, and other C run-time requirements.

With additional Microsoft Visual Studio Code plugins, users can launch interactive JTAG GDB debug sessions.

~Additionally, this PR provides an example chip config that is generated from the GUI interface, which can both run in simulation and on Arty FPGA (FPGA support added in another PR).~

**Related PRs / Issues**:
[#194](https://github.com/ucb-bar/chipyard/issues/194)
[#1046](https://github.com/ucb-bar/chipyard/issues/1046)

**Type of change**:
- [ ] Bug fix
- [x] New feature
- [x] Other enhancement

**Impact**:
- [x] RTL change
- [x] Software change (RISC-V software)
- [ ] Build system change
- [ ] Other

**Contributor Checklist**:


generators/chipyard/src/main/scala/config/ChipConfigs.scala
```diff
@@ -34,5 +34,113 @@ class ChipLikeQuadRocketConfig extends Config(
...
 
-  new chipyard.config.AbstractConfig)
+  new chipyard.config.AbstractConfig
+)
+
+/**
+  * Chip Config generated by the GUI
+  */
+class ExampleChipConfig extends Config(
+  // ==================================
```
jerryz123, line 44: Can this be merged into ChipLikeRocketConfig?

T-K-233: Where is the ChipLikeRocketConfig defined...?

jerryz123: Actualy I forgot, hasn't been merged yet, after https://github.com/ucb-bar/chipyard/pull/1484 I change ChipLikeQuadRocket to ChipLikeRocket

jerryz123:
Is the GUI going to be part of this PR?

T-K-233:
> Is the GUI going to be part of this PR?

It's on a separate branch of the submodule. There's still some work to do before it can be published.

jerryz123:
Will GUI be part of submodule? Is it possible to included it upstream?

generators/chipyard/src/main/scala/config/ChipConfigs.scala
```diff
@@ -42,7 +42,115 @@ class ChipLikeRocketConfig extends Config(
...
+  new chipyard.iobinders.WithTraceIOPunchthrough ++
+  new chipyard.iobinders.WithExtInterruptIOCells ++
+  new chipyard.iobinders.WithCustomBootPin ++
+
+  // ==================================
+  //   Set up Memory Devices
+  // ==================================
+  // External memory section
+  new testchipip.WithSerialTLClientIdBits(4) ++                     // support up to 1 << 4 simultaneous requests from serialTL port
+  new testchipip.WithSerialTLWidth(32) ++                           // fatten the serialTL interface to improve testing performance
```
jerryz123, line 94: This is wider than what we usually build (1-4 bits)

generators/chipyard/src/main/scala/config/ChipConfigs.scala
```diff
@@ -42,7 +42,115 @@ class ChipLikeRocketConfig extends Config(
...
+  new testchipip.WithCustomBootPin ++                               // add a custom-boot-pin to support pin-driven boot address
+  new testchipip.WithBootAddrReg ++                                 // add a boot-addr-reg for configurable boot address                            // use default bootrom
+
+  // ==================================
+  //   Set up tiles
+  // ==================================
+  // Debug settings
+  new chipyard.config.WithJTAGDTMKey(idcodeVersion = 2, partNum = 0x000, manufId = 0x489, debugIdleCycles = 5) ++
+  new freechips.rocketchip.subsystem.WithNBreakpoints(2) ++
+  new freechips.rocketchip.subsystem.WithNBreakpoints(2) ++
```
jerryz123, line 113: Duplicate

generators/chipyard/src/main/scala/config/fragments/PeripheralFragments.scala
```diff
@@ -29,7 +29,7 @@ import chipyard.{ExtTLMem}
   * @param hang the power-on reset vector, i.e. the program counter will be set to this value on reset
   * @param contentFileName the path to the BootROM image
   */
-class WithBootROM(address: BigInt = 0x10000, size: Int = 0x10000, hang: BigInt = 0x10040) extends Config((site, here, up) => {
+class WithBootROM(address: BigInt = 0x10000, size: Int = 0x10000, hang: BigInt = 0x10000) extends Config((site, here, up) => {
```
jerryz123, line 32: If this is changed, then we need to change the testchipip bootrom as well. Does this PR have a corresponding testchipip PR that changes that? 

Also, what is the motivation for this change?

jerryz123 (changes requested):

generators/chipyard/src/main/scala/config/ChipConfigs.scala
T-K-233: The ExampleChipConfig is generated by the GUI interface. It lists every Config fragment in the chip explicitly within this class. 

The idea is that this gives the user a clearer view of what components are included inside a chip, what Config fragment can be swapped with what else etc., and therefore can be used as a good teaching material for getting started tutorials (tutorials coming soon).

Listing Config fragments out in a single class also helps resolve the "what is overridden by what" problem. For example, for a tapeout chip, instead of overriding `WithXXXPunchthrough` with `WithXXXIOBinder`, the user will replace and switch between these two fragments for simulation vs. tapeout.

I'll try to merge this to the ChipLikeRocketConfig in future PR, but since this ExampleChipConfig is subject to user change (GUI will auto-generate the ChipConfigs.scala based on user configuration), and the scope can expand beyond using rocket cores, I suggest giving the class a more versatile name.

generators/chipyard/src/main/scala/config/ChipConfigs.scala
```diff
@@ -42,7 +42,115 @@ class ChipLikeRocketConfig extends Config(
...
+  // ==================================
+  //   Set up I/O harness
+  // ==================================
+  // The IOBinders instantiate ChipTop IOs to match desired digital IOs
+  // IOCells are generated for "Chip-like" IOs, while simulation-only IOs are directly punched through
+  new chipyard.iobinders.WithAXI4MemPunchthrough ++
+  new chipyard.iobinders.WithAXI4MMIOPunchthrough ++
+  new chipyard.iobinders.WithTLMemPunchthrough ++
+  new chipyard.iobinders.WithL2FBusAXI4Punchthrough ++
+  new chipyard.iobinders.WithBlockDeviceIOPunchthrough ++
```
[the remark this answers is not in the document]
T-K-233: The idea is to make the default example a simulation-only setting. These Config fragments are directly extracted from `chipyard.config.AbstractConfig`.

jerryz123 (changes requested):
I think this needs to be split into 3 PRs... I see three orthogonal changes here

- BootROM changes... we can change this, but we'd have to check that all the tests pass. It looks like there are some failures related to this currently, so this may take some time to debug. Moving these to a separate PR would prevent them from blocking the other changes.
- GUI-generated Configs. We ought to have a longer discussion on how these configs are generated+presented, so its clear what the differences are between the included configs, the GUI-generated configs, and the example Chip config. The GUI should be added to chipyard, not a submodule, since we would need to keep the GUI up-to-date with the SoC configuration.
- IDE support. I think there's a strong argument for inlining this into Chipyard, rather than maintaining a separate submodule. Docs and regression tests should also be included with this.

generators/chipyard/src/main/scala/config/ChipConfigs.scala
jerryz123: At one point, `AbstractConfig` did not exist, and every single example config included every single relevant config fragment, as you describe. I fought for this model for quite some time, until it became untenable to maintain so many 50-line configs.

My perspective on "tapeout configs" is that they should be as explicit as possible, and will often require careful hand-edits/bespoke-configs. With that in mind, removing `AbstractConfig` from the ChipLike configs is a move I would support.

However, I'm not sure that using GUI-configs as "tapeout configs" is the right direction, given that the GUI will only be able to support a smaller design space than what the full system is capable of. 

I'd also prefer to consolidate any "ChipLike" example config into the `ChipLikeRocketConfig`, as currently there is CI/regressions around that Config that test many features, including a regression that builds and tests the FPGA bringup harness.

A proposal based on the above points:
- Expand `ChipLikeRocketConfig` to not override `AbstractConfig`. 
- Have GUI dump configs into a separate file. Since I expect users will often want to edit `ChipConfigs.scala` by hand, this also prevents edit conflicts. One idea... user specifies a name in the GUI... GUI dumps the file into `configs/GUI_<user_config_name>.scala`. 
- Make GUI flexible enough to support generating the exact same design (and exact same config text) as the `ChipLikeRocketConfig`.

T-K-233:
> I think this needs to be split into 3 PRs... I see three orthogonal changes here
> 
> * BootROM changes... we can change this, but we'd have to check that all the tests pass. It looks like there are some failures related to this currently, so this may take some time to debug. Moving these to a separate PR would prevent them from blocking the other changes.
> * GUI-generated Configs. We ought to have a longer discussion on how these configs are generated+presented, so its clear what the differences are between the included configs, the GUI-generated configs, and the example Chip config. The GUI should be added to chipyard, not a submodule, since we would need to keep the GUI up-to-date with the SoC configuration.
> * IDE support. I think there's a strong argument for inlining this into Chipyard, rather than maintaining a separate submodule. Docs and regression tests should also be included with this.

I see. I'll make the BootROM change a separate PR.

As for the IDE support, my design idea for the IDE is that it will be a standalone environment for firmware development, with the folder `workspace` being the working directory. To create multiple projects, user would clone multiple instance of the IDE repository. I'm not sure if inlining into Chipyard main repo would be optimal, since this would create a huge overhead for repo cloning and maintenance for those who just want to work on firmware/software development.

jerryz123:
> As for the IDE support, my design idea for the IDE is that it will be a standalone environment for firmware development, with the folder `workspace` being the working directory. To create multiple projects, user would clone multiple instance of the IDE repository. I'm not sure if inlining into Chipyard main repo would be optimal, since this would create a huge overhead for repo cloning and maintenance for those who just want to work on firmware/software development.

Ok. That sounds reasonable to me. In that case, I think we should add regressions to chipyard that test that the default IDE configurations match the default chipyard configuration. Tests should check out the IDE, compile some simple hello-world that may access some devices (UART/GPIOs/SPI), and run them on the chipyard chip configuration.

T-K-233:
I think we can add regression tests at another PR. 

I would like to add the submodule to Chipyard first to get the Lab 1 flow working for now.

jerryz123 (approved):

jerryz123:
I've merged this to unblock lab development. 
Please add a documentation link from the Chipyard docs at your earliest convenience.

PR #1636, opened 2023-10-25
Support using HarnessBinders without IOBinders

jerryz123:
This makes it easier to build systems with very-custom ChipTops that don't use IOBinders for whatever reaso. Those ChipTops just need to instantiate a list of `Port` objects that the HarnessBinders attach to.

**Related PRs / Issues**:

**Type of change**:
- [ ] Bug fix
- [ ] New feature
- [ ] Other enhancement

**Impact**:
- [ ] RTL change
- [ ] Build system change
- [ ] Other

**Contributor Checklist**:


generators/chipyard/src/main/scala/ChipTop.scala
```diff
@@ -31,5 +31,5 @@ class ChipTop(implicit p: Parameters) extends LazyModule with BindingScope
   // of ChipTop (ex: ClockGroup) do not receive clock or reset.
   // However. anonymous children of ChipTop should not need an implicit Clock or Reset
   // anyways, they probably need to be explicitly clocked.
-  lazy val module: LazyModuleImpLike = new LazyRawModuleImp(this) {  }
+  lazy val module: LazyModuleImpLike = new LazyRawModuleImp(this) with DontTouch {  }
```
abejgonzalez, line 34: Why is this needed?

generators/chipyard/src/main/scala/example/FlatChipTop.scala
```diff
@@ -136,12 +149,17 @@ class FlatChipTop(implicit p: Parameters) extends LazyModule {
...
 
+    ports = ports :+ JTAGPort(() => jtag_pad)
+
     //==========================
     // UART
     //==========================
     require(system.uarts.size == 1)
     val (uart_pad, uartIOCells) = IOCell.generateIOFromSignal(system.module.uart.head, "uart_0", p(IOCellKey))
-
+    val where = PBUS // TODO fix
```
abejgonzalez, line 159: Can we fix this before merge?

abejgonzalez (approved):
Neat. LGTM

generators/chipyard/src/main/scala/example/FlatChipTop.scala
jerryz123: This is actually copied from the existing IOBinder for UART. This comment is because the underlying CanHaveUART code hardcodes the PBUS as the attachment point, instead of parameterizing it. If that were to get fixed, we would fix this too.

generators/chipyard/src/main/scala/ChipTop.scala
jerryz123: Previously, we used IOBinders to dontTouch ChipTop ports. 

Now, we use this rocketchip util. This way, the dontTouch will work correctly on designs without iobinders.

abejgonzalez (approved):
LGTM

Issue #1624, opened 2023-10-16
MultiSoC Configs should handle uniquification between harness and multiple ChipTops


abejgonzalez:
### Background Work

- [X] Yes, I searched the [mailing list](https://groups.google.com/forum/#!forum/chipyard)
- [X] Yes, I searched the [documentation](https://chipyard.readthedocs.io/)

### Feature Description

Currently, both `firtool` and `scripts/uniquify-*` support having 1 ChipTop to uniquify modules (`firtool` can only mark 1 module underneath the harness as a DUT - used to generate the corresponding json)(`scripts/uniquify-*` directly checks if the `--dut` arg matches a specific module).

This should be supported somehow.

### Motivating Example

Can't generate multi-SoC config. where modules might need to be uniquified between the harness and multiple chiptops.

jerryz123:
Shouldn't the multi-soc config tread everything that is not ChipTop0 as within the harness, from a unifuification perspectivice?

abejgonzalez:
Correct, it does. However, I think all "ChipTop{_*}"'s should be "bundled" together. Otherwise, there are modules that are uniquified incorrectly causing missing Verilog errors in VCS/Verilator.

jerryz123:
Hmmm, does the default compile for some multi-top config fail because of this? I thought the uniquification would never cause modules to be "lost" like this

abejgonzalez:
Yea I was seeing errors with my specific SoC config (fairly beefly dual-Rocket-SoC + rocc accelerators + mmio devices config). I'll try to recreate this, but this issue serves as a reminder that this is an issue.

abejgonzalez:
Also, I'm not sure this is related exactly to this issue but also comes up in a similar setup...

If you have two `addPath` (from `HasBlackBoxPath`) calls that are the same for two modules firtool errors saying "can't find <path to one of the files>" even though it exists. This can be bypassed by using `addResource` (from `HasBlackBoxResource`) which inlines the aforementioned file and then re-emits it.

jerryz123:
> If you have two addPath (from HasBlackBoxPath) calls that are the same for two modules firtool errors saying "can't find " even though it exists. This can be bypassed by using addResource (from HasBlackBoxResource) which inlines the aforementioned file and then re-emits it.

Does this happen for a single-soc build as well?

abejgonzalez:
Untested right now but this should be easy to replicate both in/out of Chipyard.

PR #1639, opened 2023-11-01
Rebuild `jar`s when `sv`/`v` files are changed + Add documentation on blackboxes

abejgonzalez:
- Moves Verilog make dependencies up to `jar` generation so that Verilog files are properly updated in the `jar`. This was bugged before since you would have to recompile Scala sources to see the Verilog sources change.
- Adds documentation on some of the corner cases of adding blackboxes
- Updates insert includes script to be recursive (if an `include` has `include` within it, rerun the replacement)
- Bump firtool to 1.58.0
- Updates the `generate-conda-lockfiles` script to delete the lockfile before regenerating (fixes issue where `conda-lock` would get stuck regenerating the lockfile if the lockfile pre-existed)
- Moves GCD to avoid 0x1000 (same address as bootrom)

It also exposes a bug mentioned here: https://github.com/ucb-bar/chipyard/issues/1624#issuecomment-1780183452. This is an issue in MFC that needs to get solved.

The error being that duplicate `addPath` calls cannot exist (not sure if having a different module name will fix the issue). Error message is the following: 

``Cannot find file /scratch/abejgonza/cy/generators/chipyard/src/main/resources/vsrc/GCDMMIOBlackBox.v``

**Related PRs / Issues**:

**Type of change**:
- [x] Bug fix
- [ ] New feature
- [ ] Other enhancement

**Impact**:
- [ ] RTL change
- [x] Build system change
- [ ] Other

**Contributor Checklist**:

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -248,7 +248,6 @@ $(SFC_MFC_TARGETS) &: $(TAPEOUT_CLASSPATH_TARGETS) $(FIRRTL_FILE) $(FINAL_ANNO_F
 		--format=fir \
 		--dedup \
 		--export-module-hierarchy \
-		--emit-metadata \
 		--verify-each=true \
 		--warn-on-unprocessed-annotations \
 		--disable-annotation-classless \
@@ -257,7 +256,6 @@ $(SFC_MFC_TARGETS) &: $(TAPEOUT_CLASSPATH_TARGETS) $(FIRRTL_FILE) $(FINAL_ANNO_F
 		--lowering-options=$(shell cat $(MFC_LOWERING_OPTIONS)) \
 		--repl-seq-mem \
 		--repl-seq-mem-file=$(MFC_SMEMS_CONF) \
-		--repl-seq-mem-circuit=$(MODEL) \
 		--annotation-file=$(SFC_ANNO_FILE) \
 		--split-verilog \
 		-o $(GEN_COLLATERAL_DIR) \
--- a/common.mk
+++ b/common.mk
@@ -246,7 +246,6 @@ $(SFC_MFC_TARGETS) &: $(TAPEOUT_CLASSPATH_TARGETS) $(FIRRTL_FILE) $(FINAL_ANNO_F
 	@if [ $(shell cat $(SFC_LEVEL)) = low ]; then cat $(TMP_DIR)/unnec-anno-deleted2.sfc.anno.json > $(SFC_ANNO_FILE) && rm $(TMP_DIR)/unnec-anno-deleted.sfc.anno.json && rm $(TMP_DIR)/unnec-anno-deleted2.sfc.anno.json; fi
 	firtool \
 		--format=fir \
-		--dedup \
 		--export-module-hierarchy \
 		--verify-each=true \
 		--warn-on-unprocessed-annotations \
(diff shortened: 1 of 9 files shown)


docs/Customization/Incorporating-Verilog-Blocks.rst
```diff
@@ -161,4 +161,31 @@ transformed or augmented by any Chipyard FIRRTL transform.
...
+
+Additionally, both mechanisms do not enforce the order of files added.
+For example:
+
+.. code-block:: scala
+
+    addResource("fileA")
+    addResource("fileB")
+
+In this case, ``fileA`` is not guaranteed to be before ``fileB`` when passed to downstream tools.
```
jerryz123, line 189: What are cases where the file needs to be passed in a specific order to the tools?

generators/chipyard/src/main/scala/example/GCD.scala
```diff
@@ -46,10 +46,11 @@ trait HasGCDIO extends BaseModule {
 }
 
 // DOC include start: GCD blackbox
-class GCDMMIOBlackBox(val w: Int) extends BlackBox(Map("WIDTH" -> IntParam(w))) with HasBlackBoxResource
+class GCDMMIOBlackBox(val w: Int) extends BlackBox(Map("WIDTH" -> IntParam(w))) with HasBlackBoxPath
```
jerryz123, line 49: Were you intending to change this example?

abejgonzalez:
Testing with `firtool-1.58` seems to fix the bug. I'll update the conda-forge feedstock.

docs/Customization/Incorporating-Verilog-Blocks.rst
abejgonzalez: For example, SV headers need to be before any SV file. If you do `addResource(a.svh)` then `addResource(b.sv)` there isn't a guarantee that the `svh` file comes before the `sv` file.

generators/chipyard/src/main/scala/example/GCD.scala
abejgonzalez: This is to expose the bug. I'll revert it once a solution is found (see the comment below).

common.mk
```diff
@@ -246,7 +246,6 @@ $(SFC_MFC_TARGETS) &: $(TAPEOUT_CLASSPATH_TARGETS) $(FIRRTL_FILE) $(FINAL_ANNO_F
 	@if [ $(shell cat $(SFC_LEVEL)) = low ]; then cat $(TMP_DIR)/unnec-anno-deleted2.sfc.anno.json > $(SFC_ANNO_FILE) && rm $(TMP_DIR)/unnec-anno-deleted.sfc.anno.json && rm $(TMP_DIR)/unnec-anno-deleted2.sfc.anno.json; fi
 	firtool \
 		--format=fir \
-		--dedup \
```
jerryz123, line 249: What happens when you remove this? We still want dedup on modules we intend to dedup.

abejgonzalez: Dedup will always happen now. This flag is removed in 1.57.0

abejgonzalez: https://github.com/llvm/circt/blob/c823db7a53131ced67668839a764bd7af84cc1b8/lib/Firtool/Firtool.cpp#L90

jerryz123: Nice

generators/chipyard/src/main/scala/example/GCD.scala
abejgonzalez: Reverted

jerryz123 (approved):
LGTM.

PR #1642, opened 2023-11-02
Add Caliptra AES256 ECB engine as RoCC accelerator

abejgonzalez:
Adds https://github.com/chipsalliance/Caliptra 's AES 256 ECB cryptography core as a RoCC accelerator. Also adds a repo to help users create RoCC memory streaming accelerators (i.e. extend the classes/traits to quickly create a streaming accel). 

**Related PRs / Issues**:

**Type of change**:
- [ ] Bug fix
- [x] New feature
- [ ] Other enhancement

**Impact**:
- [x] RTL change
- [ ] Build system change
- [ ] Other

**Contributor Checklist**:

### DIFF
--- a/.github/scripts/defaults.sh
+++ b/.github/scripts/defaults.sh
@@ -30,7 +30,7 @@ REMOTE_COURSIER_CACHE=$REMOTE_WORK_DIR/.coursier-cache
 declare -A grouping
 grouping["group-cores"]="chipyard-cva6 chipyard-ibex chipyard-rocket chipyard-hetero chipyard-boom chipyard-sodor chipyard-digitaltop chipyard-multiclock-rocket chipyard-nomem-scratchpad chipyard-spike chipyard-clone chipyard-prefetchers chipyard-shuttle"
 grouping["group-peripherals"]="chipyard-dmirocket chipyard-dmiboom chipyard-spiflashwrite chipyard-mmios chipyard-nocores chipyard-manyperipherals chipyard-chiplike chipyard-tethered"
-grouping["group-accels"]="chipyard-mempress chipyard-sha3 chipyard-hwacha chipyard-gemmini chipyard-manymmioaccels chipyard-nvdla"
+grouping["group-accels"]="chipyard-mempress chipyard-sha3 chipyard-hwacha chipyard-gemmini chipyard-manymmioaccels chipyard-nvdla chipyard-aes256ecb"
 grouping["group-constellation"]="chipyard-constellation"
 grouping["group-tracegen"]="tracegen tracegen-boom"
 grouping["group-other"]="icenet testchipip constellation rocketchip-amba rocketchip-tlsimple rocketchip-tlwidth rocketchip-tlxbar"
@@ -67,6 +67,7 @@ mapping["chipyard-shuttle"]=" CONFIG=ShuttleConfig"
 mapping["chipyard-multiclock-rocket"]=" CONFIG=MulticlockRocketConfig"
 mapping["chipyard-nomem-scratchpad"]=" CONFIG=MMIOScratchpadOnlyRocketConfig"
 mapping["chipyard-constellation"]=" CONFIG=SharedNoCConfig"
+mapping["chipyard-aes256ecb"]=" CONFIG=AES256ECBRocketConfig verilog"
 
 mapping["constellation"]=" SUB_PROJECT=constellation"
 mapping["firesim"]="SCALA_TEST=firesim.firesim.RocketNICF1Tests"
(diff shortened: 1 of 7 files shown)


build.sbt
```diff
@@ -234,6 +234,16 @@ lazy val nvdla = (project in file("generators/nvdla"))
   .settings(libraryDependencies ++= rocketLibDeps.value)
   .settings(commonSettings)
 
+lazy val caliptra_aes = (project in file("generators/caliptra-aes-acc"))
+  .dependsOn(rocketchip, rocc_acc_utils, midasTargetUtils)
+  .settings(libraryDependencies ++= rocketLibDeps.value)
+  .settings(commonSettings)
+
+llazy val rocc_acc_utils = (project in file("generators/rocc-acc-utils"))
```
jerryz123, line 242: ttypo

jerryz123: Also, is there a case for sticking this in testchipip? If rocc accelerators already need to depend on rocket-chip + testchipip, it would reduce some dependency complexity.

generators/chipyard/src/main/scala/config/RoCCAcceleratorConfigs.scala
```diff
@@ -56,3 +56,9 @@ class HwachaLargeBoomConfig extends Config(
   new boom.common.WithNLargeBooms(1) ++
   new chipyard.config.WithSystemBusWidth(128) ++
   new chipyard.config.AbstractConfig)
+
+class AES256ECBRocketConfig extends Config(
+  new aes.WithAES256ECBAccel ++                                   // use Caliptra AES 256 ECB accelerator
+  new freechips.rocketchip.subsystem.WithNBigCores(1) ++
+  new chipyard.config.WithSystemBusWidth(128) ++
```
jerryz123, line 63: Does this require 128b sbus to function correctly? We should try to isolate system-bus configuration from core/accelerator configurations.

.github/scripts/defaults.sh
```diff
@@ -67,6 +67,7 @@ mapping["chipyard-shuttle"]=" CONFIG=ShuttleConfig"
 mapping["chipyard-multiclock-rocket"]=" CONFIG=MulticlockRocketConfig"
 mapping["chipyard-nomem-scratchpad"]=" CONFIG=MMIOScratchpadOnlyRocketConfig"
 mapping["chipyard-constellation"]=" CONFIG=SharedNoCConfig"
+mapping["chipyard-aes256ecb"]=" CONFIG=AES256ECBRocketConfig verilog"
```
jerryz123, line 70: Do verilator builds of calliptra not work

jerryz123 (changes requested):

jerryz123:
The submodule should also get added to `.github/scripts/check-commit.sh`

build.sbt
abejgonzalez: It makes sense to reduce the dependency complexity... but at the same time I would like to keep `testchipip` as minimal as possible (or at least keep it closer to "test chip stuff" as possible). Right now, `testchipip` isn't enough of a bag of random components to warrant us to put other stuff in there. However, we could make the decision now to make it more of a "bag of ucb-bar stuff".

generators/chipyard/src/main/scala/config/RoCCAcceleratorConfigs.scala
abejgonzalez: I think it requires a 256b `sbus` right now since the `L2MemHelper` code is specifically made for that. But yes I agree with you.

.github/scripts/defaults.sh
abejgonzalez: It should, I just haven't tested it yet. I'll add this in later on.

generators/chipyard/src/main/scala/config/RoCCAcceleratorConfigs.scala
abejgonzalez: I abstracted out the `rocc-acc-utils` code to be SBUS agnostic but kept a require for this accelerator to use a 256b SBUS.

jerryz123: Is that a performance requirement, or a functionality requirement?

.github/scripts/defaults.sh
abejgonzalez: Ok it compiles in Verilator now (doesn't run but that can be addressed later).

generators/chipyard/src/main/scala/config/RoCCAcceleratorConfigs.scala
abejgonzalez: Functionality requirement right now. It could be abstracted with a bit more work to be a performance requirement. That's for a later day

abejgonzalez:
Requires #1639 before this PR is merged.

abejgonzalez:
@mergifyio rebase

build.sbt
jerryz123: Fair enough

jerryz123 (approved):

PR #1670, opened 2023-11-25
Install uart_tsi utility as part of build-toolchain-extra

jerryz123:
This is useful for bringup.

**Related PRs / Issues**:

**Type of change**:
- [ ] Bug fix
- [ ] New feature
- [ ] Other enhancement

**Impact**:
- [ ] RTL change
- [ ] Build system change
- [ ] Other

**Contributor Checklist**:


scripts/build-toolchain-extra.sh
```diff
@@ -118,4 +118,11 @@ cd tools/DRAMSim2
 make libdramsim.so
 cp libdramsim.so $RISCV/lib/
 
+echo '==>  Installing uart_tsi bringup utility'
+cd $RDIR
+git submodule update --init generators/testchipip
+cd generators/testchipip/uart_tsi
+make
```
rohanku, line 125: This make step will fail on Mac since there are some baudrate constants that are missing on Mac's version of termios.h. Is there a reason we need to use termios baud rate constants? If we just replace them with the actual values it will work on Mac.

There's also some Linux warnings that end up erroring on Mac, so we should probably fix that before merging this in.

If having this work on Mac is important, I can make a PR to fix the uart_tsi code to work on Mac and linux.

jerryz123: I didn't realize people were building this on Mac. Would this be the only thing that breaks the Mac build? If so, then we should fix this.

rohanku: I haven't actually built the extra toolchain on Mac, so if Mac isn't necessary to be supported we probably don't need to change uart_tsi. I only got uart_tsi to work on Mac but eventually just used a Linux VM since I only wanted to have the riscv-toolchain installed on Linux.

rohanku (approved):

PR #1683, opened 2023-12-04
FIX: Change default UART FIFO size and combine UART Config fragments

T-K-233:
Many tapeout chips are using the default `WithUART` config in the `AbstractConfig`. In `WithUART`, the TX and RX FIFO depth is set to 256, which is an unrealistically large value. Moreover, it makes UART debugging very difficult, since the FIFO is not flushed when chip is under reset. 

This PR changes the default FIFO depth to 8, which is the value used on [FE310](https://cdn.sparkfun.com/assets/7/f/0/2/7/fe310-g002-manual-v19p05.pdf). 

Additionally, since now we have control over FIFO depth in the `WithUART` config fragment, the `WithUARTFIFOEntries` is redundant. This PR changes the code that references `WithUARTFIFOEntries` to `WithUART`. 

The `WithUARTFIFOEntries` is marked deprecated, and is recommended to be removed in future releases.

**Related PRs / Issues**:

**Type of change**:
- [x] Bug fix
- [ ] New feature
- [ ] Other enhancement

**Impact**:
- [x] RTL change
- [ ] Build system change
- [ ] Other

**Contributor Checklist**:


jerryz123 (approved):

generators/chipyard/src/main/scala/config/SpikeConfigs.scala
```diff
@@ -27,7 +27,7 @@ class SpikeFastUARTConfig extends Config(
 class SpikeUltraFastConfig extends Config(
   new chipyard.WithSpikeTCM ++
   new chipyard.WithNSpikeCores(1) ++
-  new chipyard.config.WithUARTFIFOEntries(128, 128) ++
+  new chipyard.config.WithUART(txEntries=128, rxEntries=128) ++
```
abejgonzalez, line 30: This will add another UART to the system with the parameters that you care about (since the `WithUART` fragment is additive instead of overriding the original UART params). Thus the original `WithUARTFIFOEntries` is needed.

abejgonzalez (changes requested):

generators/chipyard/src/main/scala/config/SpikeConfigs.scala
T-K-233: Updated. I'm not sure if this approach works...

jerryz123: I don't think removing the UART from these designs is the right thing to do. 

If we just change the default number of entries for all designs, wouldn't this solve the problem with minimal side effects?

T-K-233: Using 128-entry FIFO is unrealistic, and would not be applicable for FPGA and ASIC tapeout designs.

T-K-233: I think sagar mentioned that this is a special case that we need a huge FIFO for UART.

jerryz123: Right yeah. For simulation we need large FIFOs for responsiveness. 

We should isolate the many-fifo-entries to only the simulation configs, mainly the FireSim configs and the Spike tile configs.

abejgonzalez: Agreed. The changed code still doesn't work now since you are removing all UARTs from the design (I assume the Spike configs need some UART)

T-K-233: Oh sorry, I messed up the order of the config fragments.

T-K-233: fixed.

jerryz123: Thanks. Now the only thing left is to fix the Firesim configs, which need 128+ FIFO entries. You can just put `WithUARTFIFOEntries` in `WithFiresimConfigTweaks`.

T-K-233: fixed

generators/firechip/src/main/scala/TargetConfigs.scala
```diff
@@ -136,6 +136,9 @@ class WithFireSimHighPerfClocking extends Config(
 
 // Tweaks that are generally applied to all firesim configs setting a single clock domain at 1000 MHz
 class WithFireSimConfigTweaks extends Config(
+  new chipyard.config.WithUART(txEntries=128, rxEntries=128) ++ // FireSim requires a larger UART FIFO buffer,
```
abejgonzalez, line 139: Keeping the same functionality needs 256 entries on both. Also, this should be moved into the `WithFireSimDesignTweaks` (this frag. is mainly used for frequency changes).

abejgonzalez (changes requested):

generators/firechip/src/main/scala/TargetConfigs.scala
T-K-233: fixed

abejgonzalez (approved):
SFTM.

PR #1716, opened 2023-12-26
Add support for symmetric serial-tilelink

jerryz123:
Updated APIs:
 - `ClockedIO[SerialIO]` and variants are now `InternalSyncSerialIO/ExternalSyncSerialIO`, for clarity
 - `SimDRAM`/`SimTSI` now can take a `chipId` parameter, allowing simulating multiple SimTSI interfaces, as well as multiple DRAM interfaces across multiple chips
 
 New:
  - `SourceSyncSerialIO`, configuration option for testchipip's serial-TL generator that generates a symmetric credited source-synchronous interface
  - `SymmetricRocketChipletConfig`, demonstrates a RocketConfig that exposes a second symmetric serial-TL port, that can be connected to an instance of itself

**Related PRs / Issues**:

**Type of change**:
- [ ] Bug fix
- [ ] New feature
- [ ] Other enhancement

**Impact**:
- [ ] RTL change
- [ ] Build system change
- [ ] Other

**Contributor Checklist**:

### DIFF
--- a/generators/chipyard/src/main/scala/config/ChipConfigs.scala
+++ b/generators/chipyard/src/main/scala/config/ChipConfigs.scala
@@ -22,8 +22,18 @@ class ChipLikeRocketConfig extends Config(
   //==================================
   // Set up I/O
   //==================================
-  new testchipip.serdes.WithSerialTLWidth(4) ++                                         // 4bit wide Serialized TL interface to minimize IO
-  new testchipip.serdes.WithSerialTLMem(size = (1 << 30) * 4L) ++                       // Configure the off-chip memory accessible over serial-tl as backing memory
+  new testchipip.serdes.WithSerialTL(Seq(testchipip.serdes.SerialTLParams(              // 1 serial tilelink port
+    manager = Some(testchipip.serdes.SerialTLManagerParams(                             // port acts as a manager of offchip memory
+      memParams = Seq(testchipip.serdes.ManagerRAMParams(                               // 4 GB of off-chip memory
+        address = BigInt("80000000", 16),
+        size    = BigInt("100000000", 16)
+      )),
+      isMemoryDevice = true
+    )),
+    client = Some(testchipip.serdes.SerialTLClientParams()),                            // Allow an external manager to probe this chip
+    phyParams = testchipip.serdes.ExternalSyncSerialParams(width=4)                     // 4-bit bidir interface, sync'd to an external clock
+  ))) ++
+
   new freechips.rocketchip.subsystem.WithNoMemPort ++                                   // Remove axi4 mem port
   new freechips.rocketchip.subsystem.WithNMemoryChannels(1) ++                          // 1 memory channel
 
@@ -60,10 +70,16 @@ class ChipBringupHostConfig extends Config(
   //=============================
   // Setup the SerialTL side on the bringup device
   //=============================
-  new testchipip.serdes.WithSerialTLWidth(4) ++                                // match width with the chip
-  new testchipip.serdes.WithSerialTLMem(base = 0x0, size = 0x80000000L,        // accessible memory of the chip that doesn't come from the tethered host
-                                        idBits = 4, isMainMemory = false) ++   // This assumes off-chip mem starts at 0x8000_0000
-  new testchipip.serdes.WithSerialTLClockDirection(provideClockFreqMHz = Some(75)) ++ // bringup board drives the clock for the serial-tl receiver on the chip, use 75MHz clock
+  new testchipip.serdes.WithSerialTL(Seq(testchipip.serdes.SerialTLParams(
+    manager = Some(testchipip.serdes.SerialTLManagerParams(
+      memParams = Seq(testchipip.serdes.ManagerRAMParams(                            // Bringup platform can access all memory from 0 to DRAM_BASE
+        address = BigInt("00000000", 16),
+        size    = BigInt("80000000", 16)
+      ))
+    )),
+    client = Some(testchipip.serdes.SerialTLClientParams()),                         // Allow chip to access this device's memory (DRAM)
+    phyParams = testchipip.serdes.InternalSyncSerialParams(width=4, freqMHz = 75)    // bringup platform provides the clock
+  ))) ++
 
   //============================
   // Setup bus topology on the bringup system
--- a/generators/chipyard/src/main/scala/config/ChipletConfigs.scala
+++ b/generators/chipyard/src/main/scala/config/ChipletConfigs.scala
@@ -36,3 +36,11 @@ class SymmetricChipletRocketConfig extends Config(
   new testchipip.soc.WithOffchipBus ++
   new freechips.rocketchip.subsystem.WithNBigCores(1) ++
   new chipyard.config.AbstractConfig)
+
+// Simulates 2X of the SymmetricChipletRocketConfig in a multi-sim config
+class MultiSimSymmetricChipletRocketConfig extends Config(
+  new chipyard.harness.WithAbsoluteFreqHarnessClockInstantiator ++
+  new chipyard.harness.WithMultiChipSerialTL(chip0=0, chip1=1, chip0portId=1, chip1portId=1) ++
+  new chipyard.harness.WithMultiChip(0, new SymmetricChipletRocketConfig) ++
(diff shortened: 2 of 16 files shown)


generators/chipyard/src/main/scala/harness/HarnessBinders.scala
```diff
@@ -205,35 +203,52 @@ class WithTiedOffDMI extends HarnessBinder({
...
+        val clock = port.io match {
+          case io: InternalSyncSerialIO => io.clock_out
+          case io: ExternalSyncSerialIO => th.harnessBinderClock
+        }
+        withClock(clock) {
+          val ram = Module(LazyModule(new SerialRAM(port.serdesser, port.params)(port.serdesser.p)).module)
+          ram.io.ser.in <> io.out
+          io.in <> ram.io.ser.out
+
+          val success = SimTSI.connect(ram.io.tsi, clock, th.harnessBinderReset, th.p(MultiChipIdx))
```
schwarz-em, line 247: Should MultiChipIdx be specific to each chiptop instead of generalized to the harness?

jerryz123: The code currently sets `MultiChipIdx` independently for each ChipTop (each time the HarnessBinders are applied). See lines 71-76 in HasHarnessInstantiators.scala

jerryz123: I admit its a bit clunky.. I considered changing the HarnessBinder interface so they are partial functions on `th: HasHarnessInstantiators, p: Port, chipId: Int`, but that pollutes this PR, and can be done separately

schwarz-em: That's true, but `th.p(MultiChipIdx)` doesn't access the individual `MultiChipIdx`s for each ChipTop, it accesses `MultiChipIdx` for the testharness through th=HasHarnessInstantiators. I checked this using scala printlns and also the verilog instantiations of SimTSI both have CHIPID of 0 in the symmetric rocket config

schwarz-em: I also couldn't think of a nice scala solution for this. The individual chip parameters are passed in through the implicit p on ApplyHarnessBinders and it seems like it might be nontrivial to pass Parameters down through the HarnessBinder class.

jerryz123: Ah, yeah, you are right. Good catch. I think the only way then is to add a new argument `chipId: Int` to the `HarnessBinderFunction` definition.

generators/chipyard/src/main/scala/config/ChipConfigs.scala
```diff
@@ -22,8 +22,18 @@ class ChipLikeRocketConfig extends Config(
   //==================================
   // Set up I/O
   //==================================
-  new testchipip.serdes.WithSerialTLWidth(4) ++                                         // 4bit wide Serialized TL interface to minimize IO
-  new testchipip.serdes.WithSerialTLMem(size = (1 << 30) * 4L) ++                       // Configure the off-chip memory accessible over serial-tl as backing memory
+  new testchipip.serdes.WithSerialTL(Seq(testchipip.serdes.SerialTLParams(              // 1 serial tilelink port
+    manager = Some(testchipip.serdes.SerialTLManagerParams(                             // port acts as a manager of offchip memory
+      memParams = Seq(testchipip.serdes.ManagerRAMParams(                               // 4 GB of off-chip memory
+        address = BigInt("80000000", 16),
```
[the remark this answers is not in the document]
jerryz123: This is describing the base address of DRAM in the example `ChipLikeRocketConfig`. The user is free to select any base+size here, as long as it does not conflict with something else, and as long as the FPGA setup is configured to match.

generators/chipyard/src/main/scala/config/ChipConfigs.scala
```diff
@@ -22,8 +22,18 @@ class ChipLikeRocketConfig extends Config(
...
+        size    = BigInt("100000000", 16)
+      )),
+      isMemoryDevice = true
+    )),
+    client = Some(testchipip.serdes.SerialTLClientParams()),                            // Allow an external manager to probe this chip
+    phyParams = testchipip.serdes.ExternalSyncSerialParams(width=4)                     // 4-bit bidir interface, sync'd to an external clock
```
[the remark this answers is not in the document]
jerryz123: Yes, wider interfaces have been built and tested in the past. I haven't changed the behavior of the serializer for the ExternalSync/InternalSync cases, so this is just changing to the user parameterizes the width.

generators/chipyard/src/main/scala/config/ChipletConfigs.scala
```diff
@@ -0,0 +1,46 @@
...
+    replicationBase = Some(1L << 32)                                // The upper 4GB goes off-chip
+  ) ++
+  new testchipip.soc.WithOffchipBus ++
+  new freechips.rocketchip.subsystem.WithNBigCores(1) ++
+  new chipyard.config.AbstractConfig)
+
+// Simulates 2X of the SymmetricChipletRocketConfig in a multi-sim config
+class MultiSimSymmetricChipletRocketConfig extends Config(
+  new chipyard.harness.WithAbsoluteFreqHarnessClockInstantiator ++
+  new chipyard.harness.WithMultiChipSerialTL(chip0=0, chip1=1, chip0portId=1, chip1portId=1) ++
```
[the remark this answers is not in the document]
jerryz123: Each chip exposes 2 serial TL ports
- Chip0
 - serTL0 => goes to FPGA bringup platform
 - serTL1 => goes to Chip1
- Chip1
 - serTL0 => goes to FPGA bringup platform
 - serTL1 => goes to Chip0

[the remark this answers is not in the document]
vikramjain236: Ok, should we also support a config with Chip0 having both serTL0 and serTL1? We would need atleast one chip to have access to the bringup platform, right?

generators/chipyard/src/main/scala/harness/HarnessBinders.scala
```diff
@@ -205,40 +203,57 @@ class WithTiedOffDMI extends HarnessBinder({
...
-    port.io.bits.in.bits := DontCare
   }
 })
 
 class WithSimTSIOverSerialTL extends HarnessBinder({
-  case (th: HasHarnessInstantiators, port: SerialTLPort) => {
-    val bits = port.io.bits
-    if (DataMirror.directionOf(port.io.clock) == Direction.Input) {
-      port.io.clock := th.harnessBinderClock
+  case (th: HasHarnessInstantiators, port: SerialTLPort, chipId: Int) if (port.portId == 0) => {
```
schwarz-em, line 228: Does the SerialTL port to external memory / bringup FPGA have to be on port 0? If so, is this documented somewhere?

jerryz123: No, its not required to be port0, but for convention, I think it should be. This HarnessBinder assumes its port0, as you've noticed. We should stick to this standard, but I'm not sure this is easy to enforce in the generator (since the generator doesn't know what a serial-TL port will connect to.).

schwarz-em: That seems reasonable. It does seem difficult to add an assert for that. How about adding a config to the TL-serdesser which allows for adding a port instead of requiring the user to configure all the ports at once? Like WithSerialTL but instead of passing a seq, pass SerialTLParams and (optionally?) an index. Since the abstract config already adds the external serial-TL port, this will help enforce that port0 is the external port and overwriting all the ports can be use-at-your-own-risk.
This could probably be added later though. I do think the chipyard documentation should be updated at some point with all the chiplet/multi-chiptop changes, and the external port0 thing would be good to note.

jerryz123: > How about adding a config to the TL-serdesser which allows for adding a port instead of requiring the user to configure all the ports at once? 

Yeah, that can be added. I think it will inevitably be confusing, however, to keep track of what is being "appended to", vs "overwritten".

vikramjain236 (approved):

jerryz123:
Merging this to fix a bug on main

PR #1745, opened 2024-01-19
Makefile: add print-FOO utility target that prints the value of a make variable

oharboe:
useful for debugging/checking value of a variable


jerryz123 (approved):
TIL you could do this... this is very useful.

PR #1749, opened 2024-01-22
Remove skip-validate option

joey0320:
The default installation option should not ask "y/n".
I think most people would want to run `build-setup.sh` and walk away 🥲 

**Related PRs / Issues**:

**Type of change**:
- [ ] Bug fix
- [ ] New feature
- [ ] Other enhancement

**Impact**:
- [ ] RTL change
- [ ] Build system change
- [ ] Other

**Contributor Checklist**:

### DIFF
--- a/.github/actions/run-tests/action.yml
+++ b/.github/actions/run-tests/action.yml
@@ -23,7 +23,7 @@ runs:
     - name: Init submodules (since only the RTL is cached)
       run: |
         conda activate ${{ env.conda-env-name-no-time }}-$(date --date "${{ env.workflow-timestamp }}" +%Y%m%d)-${{ inputs.toolchain }}
-        ./scripts/init-submodules-no-riscv-tools.sh --force
+        ./scripts/init-submodules-no-riscv-tools.sh
       shell: bash -leo pipefail {0}
 
     # Note: You shouldn't need the other inputs since it shouldn't build RTL from scratch
--- a/.github/scripts/remote-do-rtl-build.sh
+++ b/.github/scripts/remote-do-rtl-build.sh
@@ -15,7 +15,7 @@ SCRIPT_DIR="$( cd "$( dirname "$0" )" && pwd )"
 source $SCRIPT_DIR/defaults.sh
 
 cd $REMOTE_CHIPYARD_DIR
-./scripts/init-submodules-no-riscv-tools.sh --force
+./scripts/init-submodules-no-riscv-tools.sh
 
 # Constellation can run without espresso, but this improves
 # elaboration time drastically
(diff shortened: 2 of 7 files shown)


jerryz123:
I personally agree with this change

jerryz123:
It's probably a good idea to leave the flags as no-ops so people's existing flows and documentation won't all immediately break.

jerryz123:
> > It's probably a good idea to leave the flags as no-ops so people's existing flows and documentation won't all immediately break.
> 
> 
> 
> Hmmm.... but then we need to cleanup the "no-op" flags later...

Yeah, at some point. But IMO this is an instance where we should prioritize minimizing user pain.

joey0320:
There are times when radical changes just have to take place.......

jerryz123:
> There are times when radical changes just have to take place.......

Removing the flag and all relevant documentation is fine. 

Leaving behind the parsing as a no-op is a minimal QoL feature that has no impact otherwise.

joey0320:
Hmm.. I think it is even more confusing to users if there is a "flag that does nothing".

joey0320:
😢

jerryz123:
Remove it from the documentation and help text. Users should not be able to discover it and be confused. 

It would only affect existing users following old docs, or with their own scripts.

SeahK:
Oh, this change seems good. I sometimes forget to do Y and waste an hour.

SeahK (approved):

T-K-233:
Nice change. I will update the tutorials when we do a version release with this change.

T-K-233 (approved):

jerryz123:
I'd like to wait for @abejgonzalez and @sagark before merge.

jerryz123 (changes requested):
Wait for @abejgonzalez and @sagark

joey0320:
😠

joey0320:
Release has to happen...

jerryz123:
I know, but as others have asked for this feature in the past, it would be rude to unilaterally remove it without their feedback.

abejgonzalez:
This is going to get rejected by @sagark most likely. This flag being there exists strictly for new users who accidentally don't check out a release to have some sort of prompt telling them "Hey this isn't a release, this may not work".

jerryz123:
If @sagark concedes then we can merge this, after my backwards-compatibility concerns are addressed. I think we certainly have some consensus that this is the right move. 

FWIW, this isn't the Wild West of rocket-chip and RTL changes between releases anymore... at this point, most changes are strictly QoL improvements or new-features. The latest commit on main is much more likely to be usable than the latest release, and we will continue to only provide support for the latest main. 

Thus, I don't see the benefit of pushing users into using tagged releases.

sagark:
> FWIW, this isn't the Wild West of rocket-chip and RTL changes between releases anymore... at this point, most changes are strictly QoL improvements or new-features. The latest commit on main is much more likely to be usable than the latest release, and we will continue to only provide support for the latest main.
> 
> Thus, I don't see the benefit of pushing users into using tagged releases.

I'll approve, let's give it a try.

Since this is so early in the setup process, I also think it's fine to just delete the flag completely (no deprecation warning), as long as we've scrubbed references to it throughout the docs/code. It'll just error immediately for anyone using the flag and then they can fix it.

sagark (approved):

jerryz123 (approved):
Thanks for bringing this up @joey0320

PR #1788, opened 2024-02-07
Switch update-circt action to v1

seldridge:
Change the version of the update-circt action from v1.0.0 to v1.  This pulls in a change from v1.0.1 which fixes a bug where the staging branch was not updated.  (I have no idea if this is being used.)  However, this generally future-proofs the action by having it pull from v1 which is manually updated to match the latest v1 release (pulling in minor and patch changes).


jerryz123 (approved):
Thanks @seldridge !

Issue #1812, opened 2024-03-06
"Hammer Mentor Plugin" does not exist


T-K-233:
### Background Work

- [X] Yes, I searched the [mailing list](https://groups.google.com/forum/#!forum/chipyard)
- [X] Yes, I searched [prior issues](https://github.com/ucb-bar/chipyard/issues)
- [X] Yes, I searched the [documentation](https://chipyard.readthedocs.io/)

### Chipyard Version and Hash

Release: 1.10.0
Hash: [5459aa3](https://github.com/ucb-bar/chipyard/commit/5459aa337126ae8b460166bef79fa2fe1f0770e2)

### OS Setup

[here](https://gist.github.com/T-K-233/38e0aff721f2219a96f116822f9dafda)

### Other Setup

Ex: Prior steps taken / Documentation Followed / etc...

### Current Behavior

`vlsi/hammer-mentor-plugins` is pointing to https://github.com/ucb-bar/hammer-mentor-plugins/tree/eca060af456043ded569bbcc8bea49c03bff6911, which does not exist anymore

### Expected Behavior

NA

T-K-233:
Solution: when initializing or updating the repository, do not use `git submodule init` and `git submodule update`

jerryz123:
That is correct. Either use build-setup, or init-submodules scripts.

waseem-10xe:
When I run the **init-submodules-no-riscv-tools** I get following can you confrim.
+ git_submodule_exclude _skip
+ for name in toolchains/*-tools/* generators/cva6 generators/nvdla toolchains/libgloss generators/gemmini generators/rocket-chip generators/compress-acc sims/firesim software/nvdla-workload software/coremark software/firemarshal software/spec2017 tools/dsptools tools/rocket-dsp-utils tools/circt vlsi/hammer-mentor-plugins
+ _skip toolchains/riscv-tools/riscv-isa-sim
+ git config --local submodule.toolchains/riscv-tools/riscv-isa-sim.update none
+ for name in toolchains/*-tools/* generators/cva6 generators/nvdla toolchains/libgloss generators/gemmini generators/rocket-chip generators/compress-acc sims/firesim software/nvdla-workload software/coremark software/firemarshal software/spec2017 tools/dsptools tools/rocket-dsp-utils tools/circt vlsi/hammer-mentor-plugins
+ _skip toolchains/riscv-tools/riscv-openocd
[...]

PR #1814, opened 2024-03-06
FIX: remove submodule that no longer exists

T-K-233:
This PR removes the Hammer Mentor Plugin that no longer exists.

**Related PRs / Issues**:
This PR fixes Issue #1812 

**Type of change**:
- [x] Bug fix
- [ ] New feature
- [ ] Other enhancement

**Impact**:
- [ ] RTL change
- [ ] Build system change
- [x] Other

**Contributor Checklist**:

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -28,9 +28,6 @@
 [submodule "generators/sha3"]
 	path = generators/sha3
 	url = https://github.com/ucb-bar/sha3.git
-[submodule "vlsi/hammer-mentor-plugins"]
-	path = vlsi/hammer-mentor-plugins
-	url = https://github.com/ucb-bar/hammer-mentor-plugins.git
 [submodule "tools/axe"]
 	path = tools/axe
 	url = https://github.com/CTSRD-CHERI/axe.git


harrisonliew:
The Mentor plugin is only supposed to be initialized when the `init-vlsi.sh` is called. Otherwise, it should be ignored during repo setup. Did something break in this logic? https://github.com/ucb-bar/chipyard/blob/5459aa337126ae8b460166bef79fa2fe1f0770e2/scripts/init-submodules-no-riscv-tools-nolog.sh#L67-L90

jerryz123:
The `init-submodules` and `build-setup` scripts should not be trying to initialize that submodule in the first place. Can you report your git version perhaps? Does the same issue appear if you just try to run `init-submodules`?

harrisonliew:
> When running build-setup.sh and initializing the submodules, git cannot find the repository of the mentor plugin. It is pointed at https://github.com/ucb-bar/hammer-mentor-plugins, but the location does not exist on Github. Did we rename the repo or moved to somewhere else?

This plugin is supposed to be private. Only those with access to Mentor tools can request access to it. Anyways, please try to investigate why this submodule isn't being ignored during setup.

T-K-233:
Ah I see. So the issue happens because I ran `git submodule update` instead of the full `build-setup.sh` script.

Do you think it's better to by-default not include the private repos in the upstream Chipyard, and only add and initialize the submodules only when user request to (by running `init-vlsi.sh` or something). 

The current set up is a bit confusing. I've asked around a bit and turns out people have ran into this problem often, and the common solution is to remove the entry in .gitmodules repeatly.

Otherwise I can update the tutorial document to inform people not use `git submodule` commands (although this sounds evil).

harrisonliew:
Keeping the submodules included is a way to do version control without packaging or encoding specific commits/branches into scripts. You already identified the cause - that nobody should be using `git submodule` commands unless they are modifying specific submodules. The instructions to use the setup scripts are already in the Chipyard documentation; not sure what tutorials you're talking about.

PR #1822, opened 2024-03-14
Bump conda-lockfiles for latest version of conda/conda-lock

joonho3020:
Latest version of conda doesn't support `conda-lock=1.4`.

**Related PRs / Issues**:

**Type of change**:
- [ ] Bug fix
- [ ] New feature
- [ ] Other enhancement

**Impact**:
- [ ] RTL change
- [ ] Build system change
- [ ] Other

**Contributor Checklist**:


vighneshiyer:
Let's bump conda-lock to a modern version (v2+). And also update the setup docs. But I'm not sure if the new lockfiles can be read by conda-lock v1 - we should check that too; otherwise this is a breaking change

abejgonzalez:
> Let's bump conda-lock to a modern version (v2+). And also update the setup docs. But I'm not sure if the new lockfiles can be read by conda-lock v1 - we should check that too; otherwise this is a breaking change

+1 on this. We should try to bump to the latest version of `conda-lock` but still "pin" it in the setup docs and in the reqs.yaml files. We should also make sure to update FireSim as well (since the lockfile + docs there is still using 1.4). 

Also to ease the breaking change it might make sense to add a check before the conda-lock invocation to ensure that it is up-to-date (kinda how we do the same with Make).

abejgonzalez:
Unneeded since https://github.com/ucb-bar/chipyard/pull/1957 sets up conda-lock properly for all users.

PR #1852, opened 2024-04-18
Add DMI Bridge to FireSim

soohyuk-cho:
Adds a FireSim DMI Bridge for an alternative bringup interface other than TSI.

- Adds a new DMI target config for FireSim for quick testing
- Adds bridge binder for DMI
- Enables DebugModule for all FireSim designs (now that clock gating can be disabled). This matches CY's behavior of always having it.

**Related PRs / Issues**:

**Type of change**:
- [ ] Bug fix
- [ ] New feature
- [ ] Other enhancement

**Impact**:
- [ ] RTL change
- [ ] Build system change
- [ ] Other

**Contributor Checklist**:

### DIFF
--- a/generators/firechip/src/main/scala/TargetConfigs.scala
+++ b/generators/firechip/src/main/scala/TargetConfigs.scala
@@ -5,19 +5,13 @@ import java.io.File
 import chisel3._
 import chisel3.util.{log2Up}
 import org.chipsalliance.cde.config.{Parameters, Config}
-import freechips.rocketchip.groundtest.TraceGenParams
 import freechips.rocketchip.tile._
 import freechips.rocketchip.tilelink._
-import freechips.rocketchip.rocket.DCacheParams
 import freechips.rocketchip.subsystem._
-import freechips.rocketchip.devices.tilelink.{BootROMLocated, BootROMParams}
-import freechips.rocketchip.devices.debug.{DebugModuleParams, DebugModuleKey}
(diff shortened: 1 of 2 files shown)


generators/firechip/src/main/scala/TargetConfigs.scala
```diff
@@ -276,6 +270,13 @@ class FireSimSmallSystemConfig extends Config(
   new freechips.rocketchip.subsystem.WithInclusiveCache(nWays = 2, capacityKB = 64) ++
   new chipyard.RocketConfig)
 
+class FireSimDmiRocketConfig extends Config(
```
jerryz123, line 273: Is the `WithNoSerialTL` the right thing here? I think we should tie it off in the harness instead of modifying the target.

jerryz123 (changes requested):

generators/firechip/src/main/scala/TargetConfigs.scala
abejgonzalez: The problem is that we have both bridge harness binders auto-added to all designs if the IO's exist. This is a quick way to just disable one of the IOs so that the harness connects properly

jerryz123: Just add a OverrideHarnessBinder that ties it off?

jerryz123 (approved):

PR #1892, opened 2024-05-31
Added new configs and harness binders for the VCU118 to use the UART TSI connection in the same way as the Arty

BMorgan1296:
Issues: The non-TSI UART pins map to the same as the TSI pins, so I moved them to unused pins on the PMOD connectors, but this required modification of fpga-shells. Also, the LEDs do not work for the TSI connection yet.

### DIFF
--- a/fpga/src/main/scala/vcu118/Configs.scala
+++ b/fpga/src/main/scala/vcu118/Configs.scala
@@ -39,6 +39,12 @@ class WithSystemModifications extends Config((site, here, up) => {
   case SerialTLKey => Nil // remove serialized tl port
 })
 
+class WithSystemModificationsBareMetal extends Config((site, here, up) => {
+  case DTSTimebase => BigInt((1e6).toLong)
+  case ExtMem => up(ExtMem, site).map(x => x.copy(master = x.master.copy(size = site(VCU118DDRSize)))) // set extmem to DDR size
+  case SerialTLKey => Nil // remove serialized tl port
+})
+
 // DOC include start: AbstractVCU118 and Rocket
 class WithVCU118Tweaks extends Config(
   // clocking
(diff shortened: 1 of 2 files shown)


jerryz123:
> The non-TSI UART pins map to the same as the TSI pins, 

Right, you'd need a separate PMOD UART for the target UART

> Also, the LEDs do not work for the TSI connection yet.

The LEDs were a debug feature when I was setting it up on the Arty. Not require for VCU118, but still very useful.

BMorgan1296:
> Right, you'd need a separate PMOD UART for the target UART

So would we want another PR for fpga-shells which implements a different PMOD UART which uses say the male connectors on the vcu118? There is no other spare female PMOD connector.

> The LEDs were a debug feature when I was setting it up on the Arty. Not require for VCU118, but still very useful.
I will see if I can add them...

Anything else that needs to be done from what you can see?

fpga/src/main/scala/vcu118/Configs.scala
```diff
@@ -63,6 +69,37 @@ class WithVCU118Tweaks extends Config(
   new freechips.rocketchip.subsystem.WithNMemoryChannels(1)
 )
 
+class WithVCU118TweaksBareMetal extends Config(
```
jerryz123, line 72: Can we drop the "BareMetal" naming, and name it `UARTTSIVCU118` in all places. The "baremetal" terminology I feel does not accurately describe what the additional capability of these configs is.

fpga/src/main/scala/vcu118/HarnessBinders.scala
```diff
@@ -37,6 +40,32 @@ class WithDDRMem extends HarnessBinder({
...
 
+//Bare Metal Extension
+class WithVCU118UARTTSI extends HarnessBinder({
+  case (th: HasHarnessInstantiators, port: UARTTSIPort, chipId: Int) => {
+    val rawModule = th.asInstanceOf[LazyRawModuleImp].wrapper.asInstanceOf[VCU118FPGATestHarness]
+    val harnessIO = IO(new UARTPortIO(port.io.uartParams)).suggestName("uart_tsi")
+    harnessIO <> port.io.uart
+    val packagePinsWithPackageIOs = Seq(
+      ("AW25" , IOPin(harnessIO.rxd)),
+      ("BB21", IOPin(harnessIO.txd)))
```
jerryz123, line 51: Your comment indicates that this overlaps with the UART device, which uses the UARTOverlay-generated pins in the VCU118TestHarness, but this generates its own pins. Do these overlap with the UARTOverlay pins?

PR #1911, opened 2024-06-26
Bump Checkpointing Script + Docs

abejgonzalez:
Various checkpointing fixes + QoL updates:

- Fixed and more robust ckpt script
- Support changing PMPs in SpikeTile configs
- New ckpt specific configs
- Dedup + fix SpikeTile configs
- Bump FireMarshal for HTIF-only serial console Linux version
- Many docs + comments for clarification

**Related PRs / Issues**:

**Type of change**:
- [ ] Bug fix
- [ ] New feature
- [ ] Other enhancement

**Impact**:
- [ ] RTL change
- [ ] Build system change
- [ ] Other

**Contributor Checklist**:


scripts/generate-ckpt.sh
```diff
@@ -7,14 +7,18 @@ usage() {
     echo ""
     echo "Options"
     echo "  --help -h  : Display this message"
-    echo "  -n <n>     : Number of harts"
     echo "  -b <elf>   : Binary to run in spike"
     echo "  -p <pc>    : PC to take checkpoint at [default 0x80000000]"
     echo "  -i <insns> : Instructions after PC to take checkpoint at [default 0]"
+    echo "  -o <out>   : Output directory to store the checkpoint in. [default <elf>.<randomnum>.loadarch]"
+    echo "  -v         : Verbose"
+    echo ""
```
jerryz123, line 15: Why this space here?

jerryz123 (approved):

scripts/generate-ckpt.sh
abejgonzalez: It separates out sections of commands. Maybe I should just write that things are mutually exclusive within the next two sections (DTB overrides any individual parameters given).

scripts/generate-ckpt.sh
```diff
@@ -65,22 +77,32 @@ if [[ $VERBOSE -eq 1 ]] ; then
...
-    BASEMEM="$((0x80000000)):$((0x10000000))"
-else
-    BASEMEM=$MEMOVERRIDE
-fi
-SPIKEFLAGS="-p$NHARTS --pmpregions=0 --isa=$ISA -m$BASEMEM"
 BASENAME=$(basename -- $BINARY)
 
 if [ -z "$OUTPATH" ] ; then
-    OUTPATH=$BASENAME.$PC.$INSNS.loadarch
+    OUTPATH=$BASENAME.$((RANDOM)).loadarch
```
jerryz123, line 83: I don't like generating randomly named output directories. Can we keep the old default naming?

abejgonzalez: The old naming doesn't make sense when there are new options (using a DTS/DTB).  Maybe it can just be a concat of the options used (like `$BASENAME.$PC.loadarch` or `$BASENAME.dtb.loadarch`)

jerryz123: Not all the options have to be reflected in the naming ... the `PC/INSNS` is meant to let you quickly generate lots of checkpoints at different PCs, or at different cycle counts.

abejgonzalez: Ok this is updated to a deterministic name (with a bit more qualifiers).

generators/chipyard/src/main/scala/config/PeripheralDeviceConfigs.scala
```diff
@@ -61,6 +61,7 @@ class LBWIFRocketConfig extends Config(
 
 // DOC include start: DmiRocket
 class dmiRocketConfig extends Config(
+  new freechips.rocketchip.subsystem.WithExtMemSize(BigInt(2) << 30) ++
```
abejgonzalez, line 64: Will be reverted.

generators/chipyard/src/main/scala/config/SpikeConfigs.scala
```diff
@@ -17,23 +17,24 @@ class dmiSpikeConfig extends Config(
 
 // Avoids polling on the UART registers
 class SpikeFastUARTConfig extends Config(
+  new freechips.rocketchip.subsystem.WithExtMemSize(BigInt(2) << 30) ++
```
abejgonzalez, line 20: Will be reverted.

abejgonzalez:
Once #1912 is merged, I'll rebase this PR and clean up the commits. However, this currently works consistently.

docs/Advanced-Concepts/Architectural-Checkpoints.rst
```diff
@@ -29,10 +29,33 @@ Loading Checkpoints in RTL Simulation
...
+
+Checkpointing Linux Binaries
+----------------------------
+
+Checkpoints can be used to run Linux binaries with the following caveats:
+
+- The binary must only use the HTIF console and should be non-interactive (i.e no stdin available)
+- The target config must be built without a serial device (i.e. the Rocket Chip Blocks UART can't be used)
+- The binary must only use an initramfs (i.e. no block device)
+- The target config must be built without a block device (i.e. the IceBlk block device can't be used).
```
abejgonzalez, line 51: RamFS + OpenSBI must be smaller than the size of target memory (so if ramfs in FM is set to 1G, you should set it to 1G + 1M (opensbi is under 1M))

jerryz123 (changes requested):
Needs to be rebased on dev to clean the commit history

docs/Advanced-Concepts/Architectural-Checkpoints.rst
```diff
@@ -29,10 +29,34 @@ Loading Checkpoints in RTL Simulation
 --------------------------------------
 
 Checkpoints can be loaded in RTL simulations with the ``LOADARCH`` flag.
-The target config **MUST** use dmi-based bringup (as opposed to the default TSI-based bringup), and support fast ``LOADMEM``.
-The target config should also match the architectural configuration of however spike was configured when generating the checkpoint.
+The target config needs the following properties:
+
+- **MUST** use DMI-based bringup (as opposed to the default TSI-based bringup)
+- **MUST** support fast ``LOADMEM``
+- Should match the architectural configuration of however Spike was configured when generating the checkpoint (i.e. same ISA, no PMPs, etc).
```
jerryz123, line 36: Does this PMP thing really cause problems if it differs between the spike used to generate the checkpoint and the DUT?

generators/chipyard/src/main/scala/SpikeTile.scala
```diff
@@ -16,7 +16,9 @@ import freechips.rocketchip.util._
 import freechips.rocketchip.tile._
 import freechips.rocketchip.prci._
 
-case class SpikeCoreParams() extends CoreParams {
+case class SpikeCoreParams(
+  nPMPs: Int = 16
```
jerryz123, line 20: I'd prefer to just make this a hard default to 0 in the params, rather than introducing a default that is seldom useful

scripts/generate-ckpt.sh
```diff
@@ -2,30 +2,53 @@
...
+    echo "  -i <insns> : Instructions after PC to take checkpoint at [default $INSNS]"
+    echo ""
+    echo "Mutually Exclusive Option Groups (each group is mutually exclusive with one another)"
+    echo "  Group: Use Spike default DTS with modifications (can choose multiple)"
+    echo "    -n <n>     : Number of harts [default $NHARTS]"
+    echo "    -m <isa>   : ISA to pass to spike for checkpoint generation [default $ISA]"
+    echo "  Group: Use custom DT{B,S} (choose one)"
+    echo "    Important Note: a dt{b,s} only affects the devices used, pmps, mmu."
+    echo "                    the isa string and num. of harts from the dt{b,s} is inferred from this script to pass to the spike --isa and -p flags, respectively."
+    echo "    -d <dtb>   : DTB file to use. Passed to spike's '--dtb' flag. [default is to use ${DTB:-none}]"
```
jerryz123, line 47: We should fix spike to infer the isa correctly from DTS/DTB.

jerryz123: What is the reason for supporting both DTS and DTB? I'd like to avoid needless complexity

jerryz123 (changes requested):

abejgonzalez:
Only thing that seems broken is that the simulation doesn't terminate properly. Doing something like this:

image

Otherwise this PR is GTG.

docs/Advanced-Concepts/Architectural-Checkpoints.rst
abejgonzalez: For me, this fixed an issue where if an address is under `0x80200000` (where the payload is put in OpenSBI), then it would return an insn value of `0` causing an OpenSBI exception that wouldn't get resolved.

generators/chipyard/src/main/scala/SpikeTile.scala
abejgonzalez: If we are trying to match Rocket w/ the configurations maybe this should be set to 8 by default (what Rocket does). IMO it's better to match Rocket then have a specific checkpointing config that has the right arch. state set than have some implicitly set default value that works for one config (Spike) but not for the other (Rocket) when reloading checkpoints.

scripts/generate-ckpt.sh
abejgonzalez: > We should fix spike to infer the isa correctly from DTS/DTB.

Yes, but I'll leave that for a later PR (I need to do other stuff).

> What is the reason for supporting both DTS and DTB? I'd like to avoid needless complexity

This is just for ease of use, we can just support the DTB if you want to reduce flags.

jerryz123: If the DTS use case is easier (which I would think would be the case), then we don't have to support the dtb flag.

abejgonzalez: https://github.com/riscv-software-src/riscv-isa-sim/issues/1716

abejgonzalez:
> Only thing that seems broken is that the simulation doesn't terminate properly. Doing something like this:
> 
> image
> 
> Otherwise this PR is GTG.

Ok, it looks like this issue is something to do with SV39 v SV48. I'll just convert all configs to by default use SV39 again (instead of SV48) since SV39 was more thoroughly tested and it works with cospike. I'll update this PR to point to https://github.com/ucb-bar/testchipip/pull/237

jerryz123 (changes requested):
Can you force push removing the [ci skip] tag?

abejgonzalez:
> Can you force push removing the [ci skip] tag?

I want this PR reviewed and added before bumping this: https://github.com/riscv-software-src/riscv-isa-sim/pull/1721

abejgonzalez:
Bumped but still waiting for https://github.com/riscv-software-src/riscv-isa-sim/pull/1721 to fully merge this.

generators/chipyard/src/main/scala/config/PeripheralDeviceConfigs.scala
```diff
@@ -67,6 +67,14 @@ class dmiRocketConfig extends Config(
   new chipyard.config.AbstractConfig)
 // DOC include end: DmiRocket
 
+class dmiCospikeCheckpointingRocketConfig extends Config(
+  new freechips.rocketchip.rocket.WithNoCease ++                 // remove xrocket ISA extension
```
abejgonzalez, line 71: What is the right config frag. for this? I don't see this in RC?

jerryz123: I think its `WithCease(false)`

abejgonzalez:
> Bumped but still waiting for [riscv-software-src/riscv-isa-sim#1721](https://github.com/riscv-software-src/riscv-isa-sim/pull/1721) to fully merge this.

Bumped. This is ready for review.

jerryz123 (approved):

jerryz123:
Finally done.

PR #1935, opened 2024-07-31
Add hook into Chipyard build system for FireSim

abejgonzalez:
Sister PR: https://github.com/firesim/firesim/pull/1759

**Related PRs / Issues**:

**Type of change**:
- [ ] Bug fix
- [ ] New feature
- [ ] Other enhancement

**Impact**:
- [ ] RTL change
- [ ] Build system change
- [ ] Other

**Contributor Checklist**:


scripts/build-setup.sh
```diff
@@ -244,10 +242,8 @@ if run_step "6"; then
         pushd $CYDIR/sims/firesim &&
         (
             set -e # Subshells un-set "set -e" so it must be re enabled
-            echo $CYDIR
             source sourceme-manager.sh --skip-ssh-setup
             pushd sim
-            make target-classpath
```
abejgonzalez, line 250: Removed since this is a power-user target. Technically, the target classpath in FireSim should only be built for FireSim projects (for example: `midasexamples`)

abejgonzalez:
This works with CY-as-top. Since the environments are separate, if you want to run FireSim commands you need to `source sourceme-manager.sh`. This invalidates the conda environment of Chipyard (it's cleaner to have them be separate). So if you need to re-build anything related to the target (binaries, CY-RTL-sim, etc) you need to re-source Chipyards `env.sh`.

Opening this up for reivew/thoughts.

sims/firesim-wrapper/Makefile
abejgonzalez, line 1: This file is clearly minimal. I think it makes sense to have some sort of "staging" area for FireSim-related scripts + collateral s.t. in the future if Chipyard has it's own `config_*` files, etc it can put them here?

abejgonzalez: Maybe this should be some simulator agnostic folder (`sims/none`) where you can generate RTL w/o hooks into a specific simulator.

scripts/build-setup.sh
jerryz123: Is this a power-user target? I thought this was done to speed up the first firesim build.

abejgonzalez: Edit: This is *now a power user target. FireSim should now call out to the Chipyard makefile (which should already be precompiled).

jerryz123:
> This works with CY-as-top. Since the environments are separate, if you want to run FireSim commands you need to source sourceme-manager.sh. This invalidates the conda environment of Chipyard (it's cleaner to have them be separate). So if you need to re-build anything related to the target (binaries, CY-RTL-sim, etc) you need to re-source Chipyards env.sh.

This seems quite cumbersome. If this is the case, then we should leave the firesim-related packages in the CY conda environment. I don't think we can expect users to be adept at managing/switching between multiple environments.

sims/firesim-wrapper/Makefile
abejgonzalez: Yes there will be more FireSim-specific things as it's split out but IDK what they are right now. Having it be a separate staging folder is fine for now.

abejgonzalez:
> > This works with CY-as-top. Since the environments are separate, if you want to run FireSim commands you need to source sourceme-manager.sh. This invalidates the conda environment of Chipyard (it's cleaner to have them be separate). So if you need to re-build anything related to the target (binaries, CY-RTL-sim, etc) you need to re-source Chipyards env.sh.
> 
> This seems quite cumbersome. If this is the case, then we should leave the firesim-related packages in the CY conda environment. I don't think we can expect users to be adept at managing/switching between multiple environments.

If we want to have packages from both environments exist in one source of `env.sh`, I think the cleaner solution is to overlay the environments (conda supports nested environments). This allows each repo to update conda packages separately w/o having a giant single environment (at the cost of having 2 nested environments - 1 for Chipyard, 1 for FireSim).

With that being said, I think it's alright to have two separate env's that need to be sourced separately. Maybe polling Chipyard/FireSim users on whats convenient to them would be good. I suspect a lot of users/devs have separate terminals, tmux windows, etc to build things w/ FireSim v.s. Chipyard (esp. since FireSim builds/runs are typically longer lived).

abejgonzalez:
> > > This works with CY-as-top. Since the environments are separate, if you want to run FireSim commands you need to source sourceme-manager.sh. This invalidates the conda environment of Chipyard (it's cleaner to have them be separate). So if you need to re-build anything related to the target (binaries, CY-RTL-sim, etc) you need to re-source Chipyards env.sh.
> > 
> > 
> > This seems quite cumbersome. If this is the case, then we should leave the firesim-related packages in the CY conda environment. I don't think we can expect users to be adept at managing/switching between multiple environments.
> 
> If we want to have packages from both environments exist in one source of `env.sh`, I think the cleaner solution is to overlay the environments (conda supports nested environments). This allows each repo to update conda packages separately w/o having a giant single environment (at the cost of having 2 nested environments - 1 for Chipyard, 1 for FireSim).
> 
> With that being said, I think it's alright to have two separate env's that need to be sourced separately. Maybe polling Chipyard/FireSim users on whats convenient to them would be good. I suspect a lot of users/devs have separate terminals, tmux windows, etc to build things w/ FireSim v.s. Chipyard (esp. since FireSim builds/runs are typically longer lived).

Ok looks like people like the combined environment. I think the best way to do this is to have a stacked environment. I'll try it out and update this PR.

jerryz123 (commented):
The non-conda things look good. Happy to merge this once the Conda is fixed as a WIP towards independent firesim

abejgonzalez:
> > > > This works with CY-as-top. Since the environments are separate, if you want to run FireSim commands you need to source sourceme-manager.sh. This invalidates the conda environment of Chipyard (it's cleaner to have them be separate). So if you need to re-build anything related to the target (binaries, CY-RTL-sim, etc) you need to re-source Chipyards env.sh.
> > > 
> > > 
> > > This seems quite cumbersome. If this is the case, then we should leave the firesim-related packages in the CY conda environment. I don't think we can expect users to be adept at managing/switching between multiple environments.
> > 
> > 
> > If we want to have packages from both environments exist in one source of `env.sh`, I think the cleaner solution is to overlay the environments (conda supports nested environments). This allows each repo to update conda packages separately w/o having a giant single environment (at the cost of having 2 nested environments - 1 for Chipyard, 1 for FireSim).
> > With that being said, I think it's alright to have two separate env's that need to be sourced separately. Maybe polling Chipyard/FireSim users on whats convenient to them would be good. I suspect a lot of users/devs have separate terminals, tmux windows, etc to build things w/ FireSim v.s. Chipyard (esp. since FireSim builds/runs are typically longer lived).
> 
> Ok looks like people like the combined environment. I think the best way to do this is to have a stacked environment. I'll try it out and update this PR.

Update on this. A stacked environment doesn't work for us since things like Python don't support stacking (stacking is for commandline programs - i.e. appending to `PATH`).

jerryz123:
I think we ought to continue with the single environment then

jerryz123 (approved):

PR #1984, opened 2024-08-09
FireSim x Chipyard Decoupled

abejgonzalez:
Sister PR: https://github.com/firesim/firesim/pull/1791

Originally FireSim and Chipyard were coupled sharing things like (1) Scala sources (and SBT sources) (2) Conda environments (3) Bridges (...) and more. However, this PR decouples FireSim from Chipyard s.t. all "target" (ie. DUT) specific features are located in Chipyard (ie. things like FireSim UART bridges, NIC bridges, etc) and FireSim provides a hook to Chipyard so that I can run the FireSim makefile itself.

This includes the following:

- Removes existing FireSim CI tests run in `chipyard-run-tests.yml` (in the past, we directly invoked the FireSim compiler to test simple RISC-V tests) in favor of new CI run during `chipyard-full-flow.yml` (Scala-based testing moved from FireSim).
- Adds a `firesim-staging` simulator area where there is a FireSim symlink that points to FireSim when FireSim is top.
- In `build.sbt` abstracts the Scala test dependency
- Updates `common.mk` s.t. it points to the right FireSim sources.
- Separates out the firechip area into 3 SBT projects. `core` (FireSim top-level harness and target-specific bridges), `bridge-interfaces` (target bridge interfaces that are copied into FireSim), `firesim-only` (the FireSim-only side of target-specific bridges).

Chipyard documentation is minimally changed (FireSim documentation needs to be updated).

**Related PRs / Issues**:

**Type of change**:
- [ ] Bug fix
- [x] New feature
- [x] Other enhancement

**Impact**:
- [ ] RTL change
- [ ] Build system change
- [ ] Other

**Contributor Checklist**:


abejgonzalez:
This PR does not work.

build.sbt
```diff
@@ -328,32 +311,75 @@ lazy val rocketchip_inclusive_cache = (project in file("generators/rocket-chip-i
...
+
+// FireSim-side of target-specific bridges.
+// Re-uses definitions/etc provided in FireSim's midas project.
+// This is copied to FireSim's midas compiler.
+lazy val firechip_firesim_only = (project in file("generators/firechip-firesim-only"))
+  .dependsOn(firechip_bridge_interfaces)
+  .settings(
+    chiselSettings,
+    commonSettings,
+  )
```
abejgonzalez, line 372: Technically a dummy so remove.

build.sbt
```diff
@@ -328,32 +311,75 @@ lazy val rocketchip_inclusive_cache = (project in file("generators/rocket-chip-i
...
+  .settings(commonSettings)
+
+// Components of FireSim
+
+lazy val firesimAsLibrary = sys.env.get("FIRESIM_STANDALONE") == None
+lazy val firesimDir = if(firesimAsLibrary) {
+  file("sims/firesim")
+} else {
+  file("sims/firesim-staging/firesim-symlink")
+}
```
abejgonzalez, line 330: TODO: Switch to not using an env. var but always properly check the existence of a file from the Chipyard symlink.

common.mk
```diff
@@ -87,7 +87,9 @@ CHECK_SUBMODULES_COMMAND = echo "Checking all submodules in generators/ are init
 
 SCALA_EXT = scala
 VLOG_EXT = sv v
-CHIPYARD_SOURCE_DIRS = $(addprefix $(base_dir)/,generators sims/firesim/sim/src sims/firesim/sim/firesim-lib sims/firesim/sim/midas fpga/fpga-shells fpga/src tools/stage tools/stage-chisel3)
+FIRESIM_DIR = $(if $(FIRESIM_STANDALONE),sims/firesim-staging/firesim-symlink,sims/firesim)
```
abejgonzalez, line 90: TODO: Switch to not using an env. var but always properly check the existence of a file from the Chipyard symlink.

generators/firechip/core/src/main/makefrag/firesim/build.mk
```diff
@@ -0,0 +1,66 @@
...
+
+CHIPYARD_STAGING_DIR := $(chipyard_dir)/sims/firesim-staging
+
+.PHONY: copy_firesim_files
+copy_firesim_files:
+	# only copies if updated (also deletes all non-matching files)
+	rsync -pthrvza --delete \
+		$(chipyard_dir)/generators/firechip/bridge-interfaces/src/main/scala/* \
+		$(chipyard_dir)/generators/firechip/firesim-only/src/main/scala/* \
+		$(firesim_base_dir)/midas/src/main/scala/targetcopied
```
abejgonzalez, line 15: symlink plz

build.sbt
abejgonzalez: Leaving this for later.

build.sbt
```diff
@@ -328,32 +311,65 @@ lazy val rocketchip_inclusive_cache = (project in file("generators/rocket-chip-i
...
+// this should be minimally dependent on firesim
+lazy val midas_target_utils = (project in firesimDir / "sim/midas/targetutils")
+  .settings(commonSettings)
+  .settings(chiselSettings)
+
+// compiles with chisel6 (only the APIs)
+lazy val midas_standalone_target_utils = (project in file("tools/midas-targetutils"))
   .settings(commonSettings)
   .settings(chiselSettings)
+// TODO: AJG: ^ Fix
```
abejgonzalez, line 344: Mark.

abejgonzalez:
This is ready for review (passes all CI except for `commit-on-master`).

jerryz123 (commented):
I think I understand why bridges are in here  - they are target-specific.

It is not as clear to me why the `.mk` fragments are here. Aren't those core parts of firesim's build system that should remain with firesim? In what ways are they target-specific?

abejgonzalez:
> I think I understand why bridges are in here - they are target-specific.
> 
> It is not as clear to me why the `.mk` fragments are here. Aren't those core parts of firesim's build system that should remain with firesim? In what ways are they target-specific?

You can think of them similarly to `vcs.mk` or the Verilator `Makefile`. They are the hooks that FireSim provides to a downstream project (i.e. Chipyard) to configure it's make system to do the right thing. Specifically, they indicate how to regenerate the FIRRTL/annos, define extra target specific simulation flags, define default make flags, point to bridge CC files, define default driver arguments, and more. All of these things are target-specific.

jerryz123:
Ok I see, so this is a side-effect of fs's Python manager invoking makefiles that describe construction of target-specific things.

I think a longer-term goal would be to avoid having the interface to firesim be through make fragments, but this arrangement is fine fo rnow.

abejgonzalez:
> Ok I see, so this is a side-effect of fs's Python manager invoking makefiles that describe construction of target-specific things.
> 
> I think a longer-term goal would be to avoid having the interface to firesim be through make fragments, but this arrangement is fine fo rnow.

I agree. But I think this is sufficient for now (like you said)

generators/firechip/core/src/main/cc/bridges/cospike/mem_pool.cc
```diff
@@ -0,0 +1,79 @@
+#include "mem_pool.h"
+#include <assert.h>
```
jerryz123, line 2: What is this `mem_pool`/`thread_pool` stuff for? And why is it under `cospike`? If this was how it was done in fs, then its fine

jerryz123:
> Separates out the firechip area into 3 SBT projects. core (FireSim top-level harness and target-specific bridges), bridge-interfaces (target bridge interfaces that are copied into FireSim), firesim-only (the FireSim-only side of target-specific bridges).

I'm not sure I'm a fan of `firechip.core`, I found myself needing to refer to this comment to remind myself that `firechip.core` contains target-specific things, the "core" throws me off.  How about "firechip.target"? Basically everything that used to be `firechip` is now in here right?

generators/firechip/core/src/main/cc/bridges/cospike/mem_pool.cc
abejgonzalez: @joonho3020 added this. This was directly copied from FireSim.

joonho3020: IIRC, I added this as a performance optimization for the cospike bridge.

abejgonzalez:
> > Separates out the firechip area into 3 SBT projects. core (FireSim top-level harness and target-specific bridges), bridge-interfaces (target bridge interfaces that are copied into FireSim), firesim-only (the FireSim-only side of target-specific bridges).
> 
> I'm not sure I'm a fan of `firechip.core`, I found myself needing to refer to this comment to remind myself that `firechip.core` contains target-specific things, the "core" throws me off. How about "firechip.target"? Basically everything that used to be `firechip` is now in here right?

I'm avoiding calling things `target` because SBT generates the `target` directory for built Scala classes under what would originally be `core` (i.e. `firechip/target/target` for Scala classes rather than `firechip/target/src/...` for everything else). I didn't spend much time thinking of a better name than `core` to avoid this. Happy to change the package path though if you have another rec.

jerryz123:
Ah, that conflict is a very good point. 
How about: 
 - `firechip.chip` 
 - `firechip.bridges.interfaces`
 - `firechip.bridges.implementations`

generators/firechip/core/src/main/cc/bridges/cospike/mem_pool.cc
jerryz123: Ah ok, I wasn't aware of this.

joonho3020:
I find the name `firesimonly` a bit confusing.
What about

`firechip.bridges.io`: for the IO ports & case classes
`firechip.bridges.hooks`: for the annotation generator
`firechip.bridges.implementations` : for the HostBridge RTL implementation

abejgonzalez:
> I find the name `firesimonly` a bit confusing. What about
> 
> `firechip.bridges.io`: for the IO ports & case classes `firechip.bridges.hooks`: for the annotation generator `firechip.bridges.implementations` : for the HostBridge RTL implementation

IMO `interfaces` is better than `io` since IMO `io` generally denotes just Chisel IOs. I think `hooks` is ok (I think it's about as good as `stubs`). I like keeping `firesimonly` to strongly tell users "this folder is not compiled with Chipyard" directly from looking at the folder structure. Why do you think its confusing?

joonho3020:
I would like the name to convey that these are bridge RTL implementations which are elaborated when midas runs. `firesim-only` doesn't describe what these files are... What about `midas-input-impls`?

jerryz123 (commented):
Overall this looks fine. Once we resolve the package naming, I think this can be merged

abejgonzalez:
How about this instead:

* `firechip.core.firesim` -> `firechip.chip` under folder `firechip/chip/src/main/scala/*.scala`
* `firechip.core.bridges` -> `firechip.bridgeimplementations` (technically these are stubs not implementations since the implementations live on the FireSim-side - I think `stubs` would be better here) under folder `firechip/bridgeimplementations/src/main/scala/*.scala`
* `firechip.bridgeinterfaces` -> `firechip.bridgeinterfaces` under folder `firechip/bridgesinterfaces/src/main/scala/*.scala`
* `firechip.firesimonly.bridges` -> `firechip.goldengateimplementations` under folder `firechip/goldengateimplementations/src/main/scala/*.scala`

Ok @joonho3020 how about this?

joonho3020:
I don't want to nitpick too much, but now `firechip.core.bridges` -> `firechip.bridgeimplementations` is confusing. Can we just use `bridgestubs` like you mentioned?

abejgonzalez:
Nitpick now or forever hold your peace.

* `firechip.core.firesim` -> `firechip.chip` under folder `firechip/chip/src/main/scala/*.scala`
* `firechip.core.bridges` -> `firechip.bridgestubs`  under folder `firechip/bridgestubs/src/main/scala/*.scala`
* `firechip.bridgeinterfaces` -> `firechip.bridgeinterfaces` under folder `firechip/bridgeinterfaces/src/main/scala/*.scala`
* `firechip.firesimonly.bridges` -> `firechip.goldengateimplementations` under folder `firechip/goldengateimplementations/src/main/scala/*.scala`

How about now?

joonho3020:
Thanks, this I like it

jerryz123 (approved):

abejgonzalez:
Disabling CI since the latest commit(s) shouldn't affect anything.

PR #2056, opened 2024-09-18
Add example accelerator using HLS

schwarz-em:
**Related PRs / Issues**:

**Type of change**:
- [ ] Bug fix
- [x] New feature
- [ ] Other enhancement

**Impact**:
- [x] RTL change
- [ ] Build system change
- [ ] Other


.github/workflows/chipyard-run-tests.yml
abejgonzalez, line 1: Since you are just building you don't need this (you aren't running tests)

docs/Customization/Incorporating-HLS.rst
```diff
@@ -0,0 +1,69 @@
...
+
+High Level Synthesis (HLS) is a method for iterating quickly on 
+different hardware algorithms that automatically generates an RTL 
+circuit to match a specification in a high level language like C.
+
+Here, we will integrate an HLS-generated accelerator that computes 
+the Great Common Denominator (GCD) of two integers. This tutorial 
+builds on the sections :ref:`mmio-accelerators` and 
+:ref:`incorporating-verilog-blocks`. The code for this example can 
+be found in ``/generators/hls-example``
```
abejgonzalez, line 14: I would use `gh-file-ref` to have it auto-generate a link: https://github.com/ucb-bar/chipyard/blob/0ad87ff5cec66396f3b0de4d639de1b2ef7221c5/docs/Simulation/FPGA-Accelerated-Simulation.rst#L26

docs/Customization/Incorporating-HLS.rst
```diff
@@ -0,0 +1,69 @@
...
+
+Adding an HLS project
+---------------------------------------
+
+In this tutorial, we use Vitis HLS, version 2023.2.
```
abejgonzalez, line 19: The CI is using an older version of the tool FYI.

docs/Customization/Incorporating-HLS.rst
```diff
@@ -0,0 +1,69 @@
...
+
+Our project consists of 3 HLS files:
+* C program of the GCD algorithm: ``accel/HLSAccel.cpp``
+* Header file: ``accel/HLSAccel.hpp``
+* TCL script to run Vitis HLS: ``run_hls.tcl``
+
+To generate the verilog files, as well as synthesis reports, run:
+
+.. code-block:: none
+    vitis_hls run_hls.tcl
```
abejgonzalez, line 29: All code block directives need an empty space before the main text of the block.

docs/Customization/Incorporating-HLS.rst
```diff
@@ -0,0 +1,69 @@
...
+
+The files can be found in a generated folder named proj\_\<your\_project\_name>, 
+in our case, ``proj_gcd_example``.
+
+In our case, we include a ``Makefile`` to script running HLS. To generate the 
+verilog files using the Makefile, run:
```
abejgonzalez, line 35: Below you mention this also does the "move files to vsrc" step. So maybe I would add some extra saying "to script running HLS (as well as copying files around as explained later)"

generators/hls-example/accel/HLSAccel.hpp
abejgonzalez, line 1: Nitpicking a bit, can we just combine the HPP into the CPP file (instead of having two files).

generators/hls-example/accel/HLSAccel.cpp
abejgonzalez, line 1: Indentation is off in this file.

docs/Customization/Incorporating-HLS.rst
abejgonzalez: Also maybe I would point to some docs/link to Vitis HLS. Isn't there restrictions on where this can be used (and if so, should we mention something about that here)?

generators/chipyard/src/main/scala/DigitalTop.scala
abejgonzalez, line 1: Can we change the naming to something like `CanHavePeripheryHLSGCD`. Ditto for the config. fragment name.

docs/Customization/Incorporating-HLS.rst
abejgonzalez, line 1: Can you also mention the TODO of how to connect up to an AXI interface.

generators/hls-example/run_hls.tcl
```diff
@@ -0,0 +1,8 @@
+open_project -reset proj_gcd_example
+add_files accel/HLSAccel.cpp
```
abejgonzalez, line 2: HPP is missing?

generators/hls-example/run_hls.tcl
```diff
@@ -0,0 +1,8 @@
...
+set_top HLSAccelBlackBox
+open_solution -reset "solution1"
+set_part {xcvu9p-flgb2104-2-i}
```
abejgonzalez, line 5: Why do you need a part number? This is FPGA specific so maybe the docs should mention this RTL that is made is specific to an FPGA (with part number listed here) but can be run in RTL sim.

generators/hls-example/run_hls.tcl
```diff
@@ -0,0 +1,8 @@
...
+create_clock -period 10
```
abejgonzalez, line 6: I see that both the FPGA part number and the clock period should probably configurable from the makefile and piped to the Chisel.

abejgonzalez (changes requested):

abejgonzalez:
Super super cool! Thanks Ella!

abejgonzalez (commented):
Also, I think it might be good to combine this into the existing GCD example area instead of separate SBT project s.t. people can either create a GCD with 1. Chisel 2. SV 3. SV from HLS

build.sbt
```diff
@@ -263,6 +263,11 @@ lazy val rocc_acc_utils = (project in file("generators/rocc-acc-utils"))
   .settings(libraryDependencies ++= rocketLibDeps.value)
   .settings(commonSettings)
 
+lazy val hls_accel = (project in file("generators/hls-example"))
```
jerryz123, line 266: I think this should go in generators/chipyard/example, instead of its own project

docs/Customization/Incorporating-HLS.rst
schwarz-em: I looked closer on the product site and the documentation but I couldn't find any restrictions on where Vitis HLS can be used. Possibly it's in the license somewhere, but I don't really want to crawl through that... Some forum comments implied that Vitis HLS has been used in commercial tapeouts, so if there are rules they don't seem to be very strict. We should be able to just leave it up to the user to know what they're doing.

generators/chipyard/src/main/scala/example/GCD.scala
```diff
@@ -233,3 +332,8 @@ class WithGCD(useAXI4: Boolean = false, useBlackBox: Boolean = false) extends Co
   case GCDKey => Some(GCDParams(useAXI4 = useAXI4, useBlackBox = useBlackBox))
 })
 // DOC include end: GCD config fragment
+
+// useHLS cannot be used with useAXI4 and useBlackBox
+class WithHLSGCD extends Config((site, here, up) => {
+  case GCDKey => Some(GCDParams(useAXI4 = false, useBlackBox = false, useHLS = true))
+})
```
abejgonzalez, line 339: Can you just reuse the above fragment and have a check to ensure only one flag is enabled at a time.

abejgonzalez (changes requested):
Small nits otherwise LGTM

abejgonzalez (commented):
LGTM pending CI

jerryz123 (approved):

abejgonzalez (approved):

PR #2078, opened 2024-10-10
[WIP] Add scripts for sparse elf generation

soohyuk-cho:
SooHyuk: Working on sparse elf file generation to speed up the checkpoint process. 
Added some scripts I wrote for elf file scan and linker script generation.

### DIFF
--- a/sparsity-testing-scripts/create_mem_regions.py
+++ b/sparsity-testing-scripts/create_mem_regions.py
@@ -1,0 +1,218 @@
+#!/usr/bin/env python3
+
+import sys
+import struct
+import os
+
+def parse_regions(regions_file):
+    """
+    Parses the regions.txt file and returns a list of tuples containing start addresses and sizes.
+    """
+    regions = []
+    with open(regions_file, 'r') as f:
+        for line in f:
+            # Skip empty lines and comments
+            if not line.strip() or line.startswith('#'):
+                continue
+            parts = line.strip().split()
+            if len(parts) != 2:
+                print(f"Warning: Invalid line format: {line.strip()}")
+                continue
+            start_str, size_str = parts
+            try:
+                start = int(start_str, 16)
+                size = int(size_str, 10)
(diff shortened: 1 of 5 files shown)


sparsity-testing-scripts/create_mem_regions.py
```diff
@@ -0,0 +1,218 @@
...
+    cmd = ['riscv64-unknown-elf-as', '-o', obj_file, asm_file]
+    try:
+        subprocess.check_call(cmd)
+    except subprocess.CalledProcessError as e:
+        print(f"Error: Assembly failed for {asm_file}: {e}")
+        sys.exit(1)
+
+def main():
+    if len(sys.argv) != 3:
+        print("Usage: python3 extract_regions.py <mem.elf> <regions.txt>")
```
abejgonzalez, line 159: Generally, all scripts should use a shebang. See https://github.com/ucb-bar/chipyard/blob/73efe72e2a64073f47945b4a116836bd12e68c64/scripts/build-setup.sh#L1 https://github.com/ucb-bar/chipyard/blob/73efe72e2a64073f47945b4a116836bd12e68c64/scripts/insert-includes.py#L1 for examples.

sparsity-testing-scripts/create_mem_regions.py
abejgonzalez, line 1: This script creates the `.o` files needed for the linker script that's generated... why is this a separate script. I would try to combine the two files together s.t. you do:

1. Scan the memory regions to create the metadata file (C++ executable)
2. Given the metadata and elf, create all the .o's/etc + linker script
3. Pass that output to a automated way to generate the final elf file.

abejgonzalez (changes requested):
Good 1st pass. I think the main thing I want to see is a top-level script that given a ELF file generates the final sparse ELF file. Users shouldn't have to know what sequence to run all the underlying scripts.

PR #2314, opened 2025-11-15
[bug fix]: bpd sram rw the same addrparts

isaac12380:
The BTB and other BPD components are implemented using dual-port SRAMs. When the predictor performed a simultaneous write and read to the same SRAM entry, the read port would return an indeterminate value instead of either the old or new data. This violated the expected behavior and caused mispredictions.

  Environment:
  - Chipyard Version: 1.13.0
  - SRAM Model: Behavioral SRAM
  - Simulation Platform: VCS
  - Workload: Hello world

  The issue occurred when:
  1. A prediction read accessed an SRAM entry (read operation)
  2. Simultaneously, an update wrote to the same SRAM entry (write operation)
  3. The read port received an undefined value due to the read/write conflict

  Solution
  Implemented read-valid control logic to prevent read operations when a write targets the same address:

  1. Added read-valid signals for all SRAM structures:
    - data_rvalid (BIM)
    - s0_rbtb_valid, s0_rmeta_valid (BTB)
    - rtage_valid, rhius_valid, rlous_valid (TAGE)
  2. Index comparison logic to detect hazards:
    - Compare read index with write index
    - Compare read index with reset index (during initialization)
    - Disable read when addresses match
  3. Assertions to catch violations during simulation and debugging

  Modified Files
  - boom/src/main/scala/v3/ifu/bpd/bim.scala - Added data_rvalid signal and hazard detection
  - boom/src/main/scala/v3/ifu/bpd/btb.scala - Added s0_rbtb_valid and s0_rmeta_valid signals with assertions
  - boom/src/main/scala/v3/ifu/bpd/tage.scala - Added table and usefulness counter read-valid signals
  - boom/src/main/scala/v3/ifu/fetch-target-queue.scala - Added debug assertions (commented)

  Verification
  -  Simulated with VCS using behavioral SRAM models
  -  Tested with hello world workload

  Impact
  - Type: Bug fix
  - Scope: RTL change (branch prediction datapath)
  - only affects read enable logic


jimfangx:
Can you make a PR in the [BOOM repository](https://github.com/riscv-boom/riscv-boom) with your RTL changes? We'll bump the submodule when that PR is merged. Additionally, your submodule currently points to a 404: https://github.com/riscv-boom/riscv-boom/tree/4cb41fc4b78e60bfb7f611478de2d1c65b3ebd8c

Issue #2373, opened 2026-08-26
build-setup.sh pins sysroot_linux-64 to the host glibc without checking it exists


cwssemi:
### Background Work

- [x] Yes, I searched the [mailing list](https://groups.google.com/forum/#!forum/chipyard)
- [x] Yes, I searched [prior issues](https://github.com/ucb-bar/chipyard/issues)
- [x] Yes, I searched the [documentation](https://chipyard.readthedocs.io/)

### Chipyard Version and Hash

Release: main and 1.14
Hash: e602d917dcc495c58cabe906535e411707096c9c

### OS Setup

Linux ws020 6.8.0-134-generic #134~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Tue Jun 30 14:05:04 UTC  x86_64 x86_64 x86_64 GNU/Linux

Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.5 LTS
Release:	22.04
Codename:	jammy

### Other Setup

Followed documented setup from 1.13 which was OK, but 1.14 and main failed

### Current Behavior

Failed to complete build-setup.sh successfuly

### Expected Behavior

Complete build-setup.sh successfuly

### Other Information

**Severity:** blocks initial repository setup entirely on any host whose glibc is not one of the six
versions conda-forge packages a sysroot for. Compounded by a **destructive** failure mode: the
committed lockfile is removed before its replacement is known to solve, leaving the tree worse than
before the script ran.

**Verified against:** `main` `e602d917dcc495c58cabe906535e411707096c9c` — the tip as of 2026-08-26 —
and tag **1.14.0** (`0acc1e1`). Reproduced on Ubuntu 22.04, glibc **2.35**, conda-lock 2.5.7.

**Affected versions:** **1.14.0 and later**, introduced by `ucb-bar/chipyard#2271` (merged
2026-03-08, `bcb6129`). 1.13.0 has no glibc adaptation at all and pins `sysroot_linux-64=2.17`
unconditionally.

## Summary

`scripts/build-setup.sh` adapts the conda `sysroot_linux-64` pin to the machine it runs on. It reads
the host glibc and, if it differs from the pinned value, **writes the host's version string into the
pin verbatim** and regenerates the conda lockfiles from it:

```bash
# scripts/build-setup.sh:188-195 (unchanged from 1.14.0 to main e602d91)
SYS_GLIBC=$(ldd --version | awk '/ldd/{print $NF}')
[...]

Issue #2241, opened 2025-04-21
chipyard 1.13 "riscv-isa-sim” build error!


zzulb:
### Background Work

- [x] Yes, I searched the [mailing list](https://groups.google.com/forum/#!forum/chipyard)
- [x] Yes, I searched [prior issues](https://github.com/ucb-bar/chipyard/issues)
- [x] Yes, I searched the [documentation](https://chipyard.readthedocs.io/)

### Chipyard Version and Hash

Release: 1.13.0
latest

### OS Setup

ubuntu 20.0

### Current Behavior

Execute “module_all riscv-isa-sim --prefix="${RISCV}" --with-boost=no --with-boost-asio=no --with-boost-regex=no” and an error is reported. It is prompted to give the --host parameter.

### Expected Behavior

"module_all riscv-isa-sim --prefix="${RISCV}" --with-boost=no --with-boost-asio=no --with-boost-regex=no --host=x86_64-linux-gnu" This can be executed. But I don't know if it is correct.

"module_all riscv-isa-sim --prefix="${RISCV}" --with-boost=no --with-boost-asio=no --with-boost-regex=no --host=riscv${XLEN}-unknown-elf" This still reports an error, missing libpthread, I don't know how to deal with it.

zzulb:
![Image](https://github.com/user-attachments/assets/aedbe4fc-d2bb-445c-b5bf-977f28c5317d)

sdudyl:
可以交流一下？我也遇到了这个问题，大概知道原因，我的vx是13176659876

suhwanhong999:
doesn't seem to fix the issue as far as I can tell. I know the OS being old is the issue, since all of this works perfectly well on WSL (running on my laptop)

jimfangx:
Can you attempt to build https://github.com/riscv-software-src/riscv-isa-sim and https://github.com/riscv-collab/riscv-gnu-toolchain in absence of Chipyard? If that fails, might be an issue for those repositories. If those build successfully, feel free to reopen this issue and ping me and we'll take a look!

PR #2174, opened 2025-01-17
Bump sysroot to fix VCS ABI incompatibility

hansungk:
VCS shared libraries require glibc 2.33 symbols, but when using VCS within a conda environment, the conda-provided GCC attempts to link against conda sysroot that was previously too old to provide those symbols. 

The error looked like this (reported by @schwarz-em):
Screenshot 2025-01-16 112533

VCS 2024.09 looks for as new as glibc 2.33, so it works with this bump to 2.34. Ubuntu 24.04 ships with 2.39.
We might need to bump this periodically in conjunction with OS/VCS updates if we were to keep using conda GCC for VCS.

**Related PRs / Issues**:

**Type of change**:
- [x] Bug fix
- [ ] New feature
- [ ] Other enhancement

**Impact**:
- [ ] RTL change
- [ ] Software change (RISC-V software)
- [ ] Build system change
- [ ] Other

**Contributor Checklist**:
- [x] Did you set `main` as the base branch?
- [x] Is this PR's title suitable for inclusion in the changelog and have you added a `changelog:<topic>` label?
- [x] Did you state the type-of-change/impact?
- [x] Did you delete any extraneous prints/debugging code?
- [x] Did you mark the PR with a `changelog:` label?
- [x] (If applicable) Did you update the conda `.conda-lock.yml` file if you updated the conda requirements file?
- [ ] (If applicable) Did you add documentation for the feature?
- [ ] (If applicable) Did you add a test demonstrating the PR?

- [ ] (If applicable) Did you mark the PR as `Please Backport`?

**CI Help**:
Add the following labels to modify the CI for a set of features.
Generally, a label added only affect subsequent changes to the PR (i.e. new commits, force pushing, closing/reopening).
See `ci:*` for full list of labels:
- `ci:fpga-deploy` - Run FPGA-based E2E testing
- `ci:local-fpga-buildbitstream-deploy` - Build local FPGA bitstreams for platforms that are released
- `ci:disable` - Disable CI

PR #2374, opened 2026-08-26
Pin conda sysroot_linux-64 to a version that exists, and stop deleting committed lockfiles on a failed solve

cwssemi:
changelog:fixed
Please Backport

**Related PRs / Issues**:

fixes `ucb-bar/chipyard#2373`. `ucb-bar/chipyard#2271` introduced the adaptation, `ucb-bar/chipyard#2241` is the issue it was written for, and `ucb-bar/chipyard#2174` is the source of both the 2.34 pin and the VCS constraint behind it.

**Type of change**:
- [x] Bug fix
- [ ] New feature
- [ ] Other enhancement

**Impact**:
- [ ] RTL change
- [x] Build system change
- [ ] Other

**Contributor Checklist**:

**CI Help**:
- `ci:disable` - Disable CI

## Summary

`scripts/build-setup.sh` adapts the conda `sysroot_linux-64` pin to the host, and the adaptation has
three separable problems. On Ubuntu 22.04 the first one makes a fresh `./build-setup.sh` fail
outright, and the second deletes a committed lockfile on the way out.

## What is wrong

**1. The pin is set to a version that may not exist.** The script writes the host glibc into the pin
verbatim, and conda-forge publishes `sysroot_linux-64` for only six glibc versions — 2.12, 2.12.2,
2.17, 2.28, 2.34, 2.39. On a host with glibc 2.35 the solve cannot be satisfied:

```
The following packages are not available from current channels:
  - sysroot_linux-64=2.35
...
build-setup.sh: Build script failed with exit code 1 at step 1: Conda environment setup
```

That is Ubuntu 22.04, and Debian 12 (2.36) fails the same way. Note this also means
`ucb-bar/chipyard#2271` does not fix the case it was written for: `ucb-bar/chipyard#2241` came from
Ubuntu 20.04, whose glibc 2.31 is likewise unpublished, so that reporter moves from a build failure
to a solve failure.

**2. The guard never guards.** `DEFAULT_GLIBC` is read with `awk -F= '{print $2}'`, which returns
everything after the first `=` — including the trailing comment on the pinned line:

```
committed line:  - sysroot_linux-64=2.34 # need to be close to system glibc for VCS compatibility
DEFAULT_GLIBC:  [2.34 # need to be close to system glibc for VCS compatibility]
SYS_GLIBC:      [2.35]
```

`SYS_GLIBC` is a bare version, so the two never compare equal — not even on a host at exactly the
pinned 2.34. The rewrite-and-relock branch is taken on **every fresh clone**, which means the
committed lockfiles are never what a fresh clone installs.

**3. A failed solve deletes a committed file.** `generate-conda-lockfiles.sh` does `rm -rf $LOCKFILE`
before the `conda-lock` that replaces it, so problem 1 leaves the tree missing a committed lockfile
and the user cannot simply retry — the fallback was destroyed by the attempt to improve on it. Both
lockfiles are regenerated in one loop, full first, so `--use-lean-conda` does not avoid it: the lean
setup loses the non-lean lockfile it was never going to install.

## What this changes

**1/3 — read the pin without its trailing comment.** Strip `#…` before the compare, *and* move the
comment onto its own line with a note that the value is read by grep. Either alone fixes today; the
parser fix is what stops a re-added comment silently re-arming it.

**2/3 — generate lockfiles to a temp path, move on success.** With a `trap` to remove the temporary
file on any exit. A failed solve now leaves the committed lockfiles exactly as they were.

**3/3 — clamp instead of assign.** The sysroot is a build-target floor: it must be **≤** the host
glibc, and **≥ 2.33** for the VCS shared libraries that motivated the 2.34 pin
(`ucb-bar/chipyard#2174`) — it never had to *equal* the host glibc, and the comment on the pinned
line says as much. Select the greatest published sysroot ≤ host glibc, and warn instead of proceeding
silently when that lands below the VCS floor:

| host glibc | distro | before | after |
|---|---|---|---|
| 2.31 | Ubuntu 20.04, Debian 11 | 2.31 → solve fails | 2.28 + VCS-floor warning |
| 2.34 | RHEL/Rocky 9 | rewrite + relock | 2.34, no-op |
| 2.35 | Ubuntu 22.04 | 2.35 → solve fails | 2.34, no-op |
| 2.36 | Debian 12 | 2.36 → solve fails | 2.34 |
| 2.39 | Ubuntu 24.04, Fedora 40 | rewrite + relock | 2.39 |

A host that cannot be served at all — glibc older than 2.12, or no `ldd` — now keeps the committed
pin and warns, rather than writing an unusable value into the requirements file.

Version comparison is a small `version_le` in `scripts/utils.sh`, numeric per component so 2.9 sorts
below 2.10 and 2.12 below 2.12.2.

## How it was tested

* **Parsing**: the extraction returns `2.34` on the committed file, on the fixed file, and on a file
  where a trailing comment has been re-added.
* **Lockfile generation**, with a stubbed `conda-lock`: on a forced solve failure both committed
  lockfiles are untouched and no temporary file is left behind. The current script, run as a control
  in the same harness, deletes the full lockfile.
* **The clamp**, against the real code path with `ldd` stubbed: every row of the table above, plus
  glibc 2.11 and an absent `ldd`. Unstubbed on a glibc 2.35 host the whole block is now a **no-op** —
  no rewrite, no `.bak`, no relock — so the committed lockfile installs as authored.
* **Lockfiles unaffected**: `content_hash_for_platform("linux-64")` computed with conda-lock 2.5.7
  over the four requirement files is identical before and after this PR, so no lockfile regeneration
  is needed. (The same computation reproduces the hash committed in 1.14.0's lockfile from 1.14.0's
  requirement files, as a check on the method.)

Not tested: a real conda solve on a 2.31 or 2.39 host — I have neither. The "before" column for
those rows follows from the published sysroot list rather than from a run.

**Reproducing problem 1 — one trap worth knowing**: it does not reproduce by re-running in the same
tree, because the first run's `sed` rewrote the pinned line *without* its comment, after which the
comparison behaves as intended. Start from `git checkout -- conda-reqs/`.


jimfangx:
Thanks for the PR, we are taking a look at this (we're seeing some issues on our side as well that may point to this being the fix). Once we verify, we'll get this merged, thanks!

cwssemi:
No probs, it is a confounder for those first spinnig up Chipyard. Once you get your head around how things work it is easy enough to work around but is friction initially.

Removing the "# need to be close to system glibc for VCS compatibility" comment from trailing to its own line does deal with immediate failure. The rest is further hardening, perhaps I should have made it 2 PRs, I can do that if that helps.

cwssemi:
Wrong tab, sorry

